Issue #10, opened 2015-12-15
Add support for Mac OS X


gavv:
- [x] Adapt SConstruct for MacOS
- [x] Adapt 3rdparty building for MacOS (#88)
- [x] Don't use version script on MacOS or maybe don't use it at all (#89)
- [x] Disable _POSIX_C_SOURCE on MacOS (doesn't work as expected)
- [x] Remove core::SpinMutex (we don't really need it) (#91)
- [x] Add target_posixtime and target_darwin to SConstruct (#99)
- [x] Port time functions to MacOS (#99)
- [x] Add Travis build for MacOS
- [x] Update wiki

gavv:
### Portability issues

- [ ] Remove core::SpinMutex.
- [ ] Remove _POSIX_C_SOURCE.
- [ ] Add target_posixrt and target_darwin to SConstruct.
- [ ] Remove core::sleep_until_ms.
- [ ] Remove audio::TimedWriter.
- [ ] Replace core::Timer with core::Ticker with sleep_until_next_tick() method. We need two implementations: target_posixrt and tartget_darwin.
- [ ] Use core::Ticker instead of audio::TimedWriter in roc_pipeline.

gavv:
OSX builds started failing:

https://travis-ci.org/roc-project/roc/builds/285192784

gavv:
Fixed with 7a440febf3c3a41035a1cf34c96ca4a431dea992.

PR #99, opened 2017-08-02
roc-core: make time operation portable

dshil:
Due to #10 we got a problem, that mac os doesn't provide `clock_gettime` (It was added in mac os 10.12 mac os sierra) and `clock_nanosleep`.


dshil:
All tests successfully passed for all platforms, but for a OS X platform we still have a strange behavior.

When you run `roc-recv` and `roc-send` utils all work as expected, but if you switch to another app a roc will sleep forever until you'll toggle a volume button, fallback solution (with gettimeofday + nanosleep) works correctly, that's why the problem is somewhere in my OS X specific implementation.

I'll figure out what the hell going on with os x and after that I think we'll merge.

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -49,6 +52,117 @@ void sleep_for_ms(uint64_t ms) {
...
     }
 }
+#elif defined(__APPLE__) && defined(__MACH__)
+
+#include <mach/clock.h>
+#include <mach/clock_types.h>
+#include <mach/mach.h>
+#include <mach/mach_time.h>
+
+#define CLOCK_MONOTONIC SYSTEM_CLOCK
```
gavv: This define looks strange, why do we need it?

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -49,6 +52,117 @@ void sleep_for_ms(uint64_t ms) {
...
+
+uint64_t timestamp_ms() {
+    clock_serv_t cclock;
+    mach_timespec_t mts;
+
+    // Receive a permission to send a messages to a specified clock service.
+    kern_return_t ret =
+        host_get_clock_service(mach_host_self(), CLOCK_MONOTONIC, &cclock);
```
gavv: I think we should avoid calling host_get_clock_service and mach_port_deallocate every time. Are they expensive?

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -49,6 +52,117 @@ void sleep_for_ms(uint64_t ms) {
...
+        //
+        // https://opensource.apple.com/source/xnu/xnu-2422.1.72/osfmk/kern/clock_oldops.c
+        // We are interested in a @clock_sleep_trap, because it's used by
+        // @clock_sleep under the hood.
+        ret = clock_sleep(MACH_PORT_NULL, TIME_ABSOLUTE, ts, NULL);
+        if (ret == KERN_SUCCESS) {
+            break;
+        }
+
+        if (ret != EINTR) {
```
gavv: Does mach functions that return kern_return_t return POSIX errno values, like EINTR?

gavv:
Great.

> if you switch to another app a roc will sleep forever until you'll toggle a volume button

Probably some CoreAudio usage issue? You can try to debug sndio::Writer, which uses SoX, which uses CoreAudio. You can also check [this snippet](https://github.com/gavv/snippets/blob/master/decode_play/sox_play.cpp) from [this blogpost](https://gavv.github.io/blog/decode-play/), it's may be easier to debug it.

> Also, I think we need to fix a clock_nanosleep usage.

Right, this is a bug.

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -49,6 +52,117 @@ void sleep_for_ms(uint64_t ms) {
...
+    if (ret != KERN_SUCCESS) {
+        roc_panic("clock_get_time(CLOCK_MONOTONIC): %d", ret);
+    }
+
+    ret = mach_port_deallocate(mach_task_self(), cclock);
+    if (ret != KERN_SUCCESS) {
+        roc_panic("mach_port_deallocate: %d", ret);
+    }
+
+    return uint64_t(mts.tv_sec * 1000 + ((long)mts.tv_nsec) / 1000000);
```
gavv: I think you want `uint64_t` instead of `long` here.

gavv: I'd write it like this:

```
return uint64_t(mts.tv_sec) * 1000 + uint64_t(mts.tv_nsec) / 1000000;
```

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -49,6 +52,117 @@ void sleep_for_ms(uint64_t ms) {
...
+        // https://opensource.apple.com/source/xnu/xnu-2422.1.72/osfmk/kern/clock_oldops.c
+        // We are interested in a @clock_sleep_trap, because it's used by
+        // @clock_sleep under the hood.
+        ret = clock_sleep(MACH_PORT_NULL, TIME_ABSOLUTE, ts, NULL);
+        if (ret == KERN_SUCCESS) {
+            break;
+        }
+
+        if (ret != EINTR) {
+            roc_panic("clock_sleep(TIME_ABSOLUTE): %d", ret);
```
gavv: It would be handful to included the error string to panic message.

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -49,6 +52,117 @@ void sleep_for_ms(uint64_t ms) {
         }
     }
 }
+#elif defined(__APPLE__) && defined(__MACH__)
```
gavv: We need a separate target directory for macos-specific code, because it has nothing to do with posix. IIRC we already have target_darwin.

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -49,6 +52,117 @@ void sleep_for_ms(uint64_t ms) {
...
+}
+
+void sleep_for_ms(uint64_t ms) {
+    mach_timespec_t ts;
+    ts.tv_sec = (unsigned int)ms / 1000;
+    ts.tv_nsec = ms % 1000 * 1000000;
+
+    kern_return_t ret = KERN_SUCCESS;
+    for (;;) {
+        ret = clock_sleep(MACH_PORT_NULL, TIME_RELATIVE, ts, NULL);
```
gavv: It seems that this call has the same bug that you've found in current clock_nanosleep usage in case of EINTR.

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -49,6 +52,117 @@ void sleep_for_ms(uint64_t ms) {
...
+    }
+}
+#else
+uint64_t timestamp_ms() {
+    struct timeval tv;
+    if (gettimeofday(&tv, NULL) == -1) {
+        roc_panic("gettimeofday: %s", errno_to_str().c_str());
+    }
+
+    return uint64_t(tv.tv_sec * 1000 + tv.tv_usec / 1000);
```
gavv: Same note about casts.

src/modules/roc_core/target_posix/roc_core/time.cpp
```diff
@@ -17,6 +19,7 @@
 namespace roc {
 namespace core {
 
+#if defined(CLOCK_MONOTONIC) && defined(TIMER_ABSTIME)
```
gavv: I think we need separate #ifdefs for CLOCK_MONOTONIC and TIMER_ABSTIME. E.g. it's possible that CLOCK_MONOTONIC is available, and TIMER_ABSTIME is not.

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: We can remove it, it helps to remember that a xnu use SYSTEM_CLOCK for a monotonic time.

dshil: b5a17852bb5a62c71f13909df10c557f06ae930e

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: b5a17852bb5a62c71f13909df10c557f06ae930e

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: b5a17852bb5a62c71f13909df10c557f06ae930e

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: Not, clock_sleep doesn't return EINTR (4), it returns KERN_ABORTED (14), this mistake was introduced because previously I received EINTR from os x ipc subsystem during receiving rights to mach port (no need to go deeper in this concept).

I've fixed it.

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: Fixed in: 5a7a9b09bcd9954e181e805f1f0e88ef69b74090

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: I think so, because they are system calls, I get rid of them in: 2ae9b9ce57ef6afde2bc29c6521441dfa23b2430

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: Yep, I added `mach_error_string` that forms a human readable string for mach errors (I hope so).

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: See: 87c7443b3c636212ae3c02c7d2ac51edf2c6ff02

dshil:
>>Also, I think we need to fix a clock_nanosleep usage.

> Right, this is a bug.

ping @gavv,

I'll fix it in another PR or have you fixed it in packets3 brunch? (I know, you've fixed the whole project o_O)

src/modules/roc_core/target_posix/roc_core/time.cpp
dshil: Agree, I'll fix it, but mach provides a bit different API, the last param means: `The actual (absolute) time at which the sleep terminated.`

gavv:
> I'll fix it in another PR or have you fixed it in packets3 brunch?

I've switched from ms to us in that branch, but I didn't fix this bug.

src/modules/roc_core/target_darwin/roc_core/time.cpp
```diff
@@ -0,0 +1,105 @@
...
+    /* mach_absolute_time() returns a Mach Time unit - clock ticks. The
+     * length of a tick is a CPU dependent. On most Intel CPUs it probably
+     * will be 1 nanoseconds per tick, but let's not rely on this. Mach
+     * provides a conversation factor that can be used to convert abstract
+     * mach time units to nanoseconds.
+     */
+    static double steady_factor = 0;
+    static uint64_t tm_start = 0;
+
+    if (!tm_start) {
```
gavv: This is not threadsafe. You could either use a mutex with a double-check (so only the first calls will be expensive), or a singleton.

src/modules/roc_core/target_darwin/roc_core/time.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+void sleep_for_ms(uint64_t ms) {
+    mach_timespec_t ts;
+    ts.tv_sec = (unsigned int)ms / 1000;
+    ts.tv_nsec = ms % 1000 * 1000000;
+
+    kern_return_t ret = KERN_SUCCESS;
+    for (;;) {
+        ret = clock_sleep(MACH_PORT_NULL, TIME_RELATIVE, ts, NULL);
```
gavv: The last argument is still ignored, I guess it's a bug?

src/modules/roc_core/target_darwin/roc_core/time.cpp
```diff
@@ -0,0 +1,105 @@
...
+        if (ret != KERN_ABORTED) {
+            roc_panic("clock_sleep(TIME_ABSOLUTE): %s", mach_error_string(ret));
+        }
+    }
+}
+
+void sleep_for_ms(uint64_t ms) {
+    mach_timespec_t ts;
+    ts.tv_sec = (unsigned int)ms / 1000;
+    ts.tv_nsec = ms % 1000 * 1000000;
```
gavv: What is the type of nsec? It feels you need a cast.

src/modules/roc_core/target_posix_time/roc_core/time.cpp
```diff
@@ -25,7 +28,18 @@ uint64_t timestamp_ms() {
 
     return uint64_t(ts.tv_sec * 1000 + ts.tv_nsec / 1000000);
 }
+#else
```
gavv: The following code style is preferred:

```
#ifdef CLOCK_MONOTONIC
...
#else // !CLOCK_MONOTONIC
...
#endif // CLOCK_MONOTONIC
```

SConstruct
```diff
@@ -371,12 +371,19 @@ if GetOption('with_targets'):
...
+        targets = [
             'target_stdio',
             'target_gnu',
             'target_uv',
-        ])
+            'target_posix',
+        ]
+
+        if platform == 'linux':
+            targets.append('target_posix_time')
```
gavv: I'd prefer target_posixtime.

src/modules/roc_core/target_darwin/roc_core/time.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "roc_core/panic.h"
+#include "roc_core/time.h"
+
+namespace roc {
+namespace core {
+
+#include <mach/clock.h>
+#include <mach/clock_types.h>
+#include <mach/mach.h>
+#include <mach/mach_time.h>
```
gavv: #includes should always be outside of namespaces.

dshil: Oops, I missed it when splitting ifdef by targets.

src/modules/roc_core/target_darwin/roc_core/time.cpp
dshil: int

src/modules/roc_core/target_posix_time/roc_core/time.cpp
```diff
@@ -49,6 +63,26 @@ void sleep_for_ms(uint64_t ms) {
...
+void sleep_until_ms(uint64_t ms) {
+    uint64_t now = timestamp_ms();
+    if (ms > now) {
+        sleep_for_ms(ms - now);
+    }
+}
+
+void sleep_for_ms(uint64_t ms) {
```
gavv: I think we need separate #ifdefs for TIMER_ABSTIME and CLOCK_MONOTONIC, because they are actually separate features.

src/modules/roc_core/target_darwin/roc_core/time.cpp
dshil: Agree, I've fixed it.

gavv:
This snippet:

```
 void sleep_for_ms(uint64_t ms) {
     timespec ts;
     ts.tv_sec = ms / 1000;
     ts.tv_nsec = ms % 1000 * 1000000;
 
     while (clock_nanosleep(CLOCK_MONOTONIC, 0, &ts, NULL) == -1) {
         if (errno != EINTR) {
             roc_panic("clock_nanosleep(CLOCK_MONOTONIC): %s", errno_to_str().c_str());
          }
      }
  }
```

..should be under CLOCK_MONOTONOC #ifdef, not TIMER_ABSTIME one.

And this one:

```
void sleep_until_ms(uint64_t ms) {
    timespec ts;
    ts.tv_sec = ms / 1000;
    ts.tv_nsec = ms % 1000 * 1000000;

    while (clock_nanosleep(CLOCK_MONOTONIC, TIMER_ABSTIME, &ts, NULL) == -1) {
        if (errno != EINTR) {
            roc_panic("clock_nanosleep(CLOCK_MONOTONIC): %s", errno_to_str().c_str());
        }
    }
}
```

..obviously should be under both #ifdefs.

dshil:
ping @gavv, 

> This snippet ..should be under CLOCK_MONOTONOC #ifdef, not TIMER_ABSTIME one.

But what we'll do in case when CLOCK_MONOTONIC was defined but TIMER_ABSTIME wasn't ?

I'd prefer to have following:

```
# ifdef CLOCK_MONOTONIC
// use clock_gettime
# else // !CLOCK_MONOTONIC
// fallback to gettimeofday
# endif

# if defined(CLOCK_MONOTONIC) && defined(TIMER_ABSTIME)
// use clock_nanosleep or define own TIMER_ABSTIME and
// use clock_nanosleep if CLOCK_MONOTONIC was defined.
# else // !CLOCK_MONOTONIC && !TIMER_ABSTIME
// fallback to nanosleep
# endif

```

gavv:
Thanks!

PR #104, opened 2017-08-26
roc-core: use pthread_once in random related operations

dshil:


dshil:
Due to #99 we need to initialize a resource one time in a thread-safe manner. To avoid using non-reliable DCLP further it's been decided to use reliable and portable pthread_once.

dshil:
ping @gavv

gavv:
LGTM.

PR #110, opened 2017-09-20
travis: osx support

dshil:
#10


dshil:
ping @gavv

scripts/travis/osx/build.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+set -xe
+
+brew tap Homebrew/bundle
+brew bundle --file=scripts/travis/osx/Brewfile
+
+# build and test
+scons -Q --enable-werror --with-3rdparty=all test
```
gavv: --with-3rdparty=openfec,cpputest test

gavv: https://github.com/roc-project/roc/wiki/Building-%28native%29#macos

scripts/travis/run_test.sh
```diff
@@ -0,0 +1,17 @@
+#! /bin/bash
+if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
+    scripts/travis/osx/build.sh
+    scripts/travis/print_status.sh rocproject/ci-osx $?
```
gavv: Do we really need print_status for os x builds? I think it makes sense only for linux builds, where we test many docker containers in a single travis build.

scripts/travis/osx/build.sh
gavv: Also, I think it makes sense to build both debug and release variants. And probably enable sanitizers in debug variant.

gavv:
I also suggest to either move all linux-specific scripts from `scripts/travis/` to `scripts/travis/linux/`, or move files from `scripts/travis/osx/` to `scripts/travis/`.

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -149,6 +149,7 @@ TEST(resampler, upscaling_twice_single) {
     }
 }
 
+#ifndef ROC_TARGET_DARWIN
 // Check upsampling quality and the cut-off band with white noise.
 TEST(resampler, upscaling_twice_awgn) {
```
gavv: I think we want to ignore the test using IGNORE_TEST instead of completely hiding in.

scripts/travis/osx/build.sh
dshil: Fixed

scripts/travis/run_test.sh
dshil: You are right, I thought that we would have various osx versions, but each of them will be a single travis build.

src/tests/roc_audio/test_resampler.cpp
dshil: Fixed

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -149,6 +149,38 @@ TEST(resampler, upscaling_twice_single) {
     }
 }
 
+#ifdef ROC_TARGET_DARWIN
+// Check upsampling quality and the cut-off band with white noise.
+IGNORE_TEST(resampler, upscaling_twice_awgn) {
```
gavv: Why to copy the whole test? It's enough to copy only the header:

```
// Check upsampling quality and the cut-off band with white noise.
#ifdef ROC_TARGET_DARWIN
IGNORE_TEST(resampler, upscaling_twice_awgn) {
#else
TEST(resampler, upscaling_twice_awgn) {
#endif
    ...
}
```

scripts/travis/run_test.sh
```diff
@@ -0,0 +1,16 @@
...
+    scripts/travis/linux/docker.sh rocproject/ci-ubuntu:16.04 scripts/travis/linux/ubuntu-16.04.sh
+    scripts/travis/linux/docker.sh rocproject/ci-ubuntu:14.04 scripts/travis/linux/ubuntu-14.04.sh
+    scripts/travis/linux/docker.sh rocproject/ci-fedora scripts/travis/linux/fedora.sh
+    scripts/travis/linux/docker.sh rocproject/ci-debian scripts/travis/linux/debian.sh
+    scripts/travis/linux/docker.sh rocproject/ci-centos scripts/travis/linux/centos.sh
+    scripts/travis/linux/docker.sh rocproject/ci-empty scripts/travis/linux/empty.sh
+    scripts/travis/linux/docker.sh rocproject/cross-raspberry scripts/travis/linux/raspberry.sh
+    scripts/travis/linux/docker.sh rocproject/cross-linaro scripts/travis/linux/linaro.sh
+    cat build.status
+fi
```
gavv: If one of the linux scripts would fail, run_test.sh will still exit successfully and the build won't fail.

scripts/travis/run_test.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
+if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
```
gavv: BTW, is it possible to do this if in travis instead of a bash script? E.g.:

```
- os: osx
  script:
    - scripts/travis/osx/build.sh
- os: linux
  script:
    - scripts/travis/linux/docker.sh rocproject/ci-ubuntu:17.04 
    ...
```

src/tests/roc_audio/test_resampler.cpp
dshil: fixed

scripts/travis/run_test.sh
dshil: Fixed

scripts/travis/run_test.sh
dshil: Fixed

Issue #85, opened 2017-07-05
Sample rate conversion in roc_recv


baranovmv:
Due to #84, the Roc will be able to convert a sample rate of received signal on its own. For that purpose some changes in the other part of the code should be made:

* The Resampler's scaling factor should be properly adjusted according to the information from the RTP-packet headers and the sample rate of the renderer.
* The renderer (such as SoX) should provide the output sample rate to the other part of the code.

gavv:
So, the plan:

- [ ] sndio::Writer should report the output sample rate
- [ ] pipeline::Session should obtain the output sample rate
- [ ] some component in the session pipeline should obtain the input sample rate from the first received RTP packet
- [ ] the Resampler scaling factor should be adjusted according to the output sample rate and input sample rate

gavv:
We likely also want two new command line options:

- [ ] output sample rate in roc-recv (the user may want to set it when writing to a file or when the sound card sample rate was detected incorrectly)

- [ ] network sample rate in roc-send (the user may want values other than 44100, e.g. to avoid unnecessary resampling; currently, we use SoX to unconditionally resample input samples to 44100)

gavv:
This issue will fix #61.

gavv:
Finally done.

PR #113, opened 2017-10-15
roc_audio: adjust resampler coef 

dshil:
#85

### DIFF
--- a/src/lib/roc/config.h
+++ b/src/lib/roc/config.h
@@ -105,6 +105,9 @@ typedef struct roc_receiver_config {
 
     //! A bitmask of ROC_FLAG_* constants.
     unsigned int flags;
+
+    //! Number of samples per second per channel.
+    unsigned int sample_rate;
 } roc_receiver_config;
 
 #ifdef __cplusplus
--- a/src/modules/roc_audio/resampler_updater.cpp
+++ b/src/modules/roc_audio/resampler_updater.cpp
@@ -22,7 +22,8 @@ const core::nanoseconds_t LogRate = 5000000000;
 } // namespace
 
 ResamplerUpdater::ResamplerUpdater(packet::timestamp_t update_interval,
-                                   packet::timestamp_t aim_queue_size)
+                                   packet::timestamp_t aim_queue_size,
+                                   const float sample_rate_coef)
     : writer_(NULL)
     , reader_(NULL)
     , resampler_(NULL)
@@ -34,7 +35,8 @@ ResamplerUpdater::ResamplerUpdater(packet::timestamp_t update_interval,
     , head_(0)
     , has_tail_(false)
     , tail_(0)
-    , started_(false) {
+    , started_(false)
+    , sample_rate_coef_(sample_rate_coef) {
 }
 
 void ResamplerUpdater::set_writer(packet::IWriter& writer) {
@@ -96,13 +98,16 @@ bool ResamplerUpdater::update(packet::timestamp_t time) {
         update_time_ += update_interval_;
     }
 
+    const float adjusted_coef = sample_rate_coef_ * fe_.freq_coeff();
+
     if (rate_limiter_.allow()) {
-        roc_log(LogDebug, "resampler updater: queue_size=%lu fe=%.5f",
-                (unsigned long)queue_size, (double)fe_.freq_coeff());
+        roc_log(LogDebug, "resampler updater: queue_size=%lu fe=%.5f, adjust_fe=%.5f",
+                (unsigned long)queue_size, (double)fe_.freq_coeff(),
+                (double)adjusted_coef);
     }
 
     roc_panic_if(!resampler_);
-    return resampler_->set_scaling(fe_.freq_coeff());
+    return resampler_->set_scaling(adjusted_coef);
 }
 
 } // namespace audio
--- a/src/modules/roc_audio/resampler_updater.h
+++ b/src/modules/roc_audio/resampler_updater.h
@@ -34,8 +34,11 @@ public:
     //! @b Parameters
     //!  - @p update_interval defines how often to call FreqEstimator, in samples
     //!  - @p aim_queue_size defines FreqEstimator target queue size, in samples
+    //!  - @p sample_rate_coef represents a factor for converting session sample rate to
+    //!       output sample rate of a receiver.
     ResamplerUpdater(packet::timestamp_t update_interval,
-                     packet::timestamp_t aim_queue_size);
+                     packet::timestamp_t aim_queue_size,
+                     const float sample_rate_coef);
 
     //! Set output writer.
     void set_writer(packet::IWriter&);
@@ -76,6 +79,8 @@ private:
     packet::timestamp_t tail_;
 
     bool started_;
+
+    const float sample_rate_coef_;
 };
 
 } // namespace audio
--- a/src/modules/roc_pipeline/config.h
+++ b/src/modules/roc_pipeline/config.h
@@ -125,7 +125,7 @@ struct ReceiverConfig {
     //! Default parameters for session.
     SessionConfig default_session;
 
-    //! Sample rate, number of samples for all channels per second.
+    //! Number of samples per second per channel.
     size_t sample_rate;
 
     //! Channel mask.
@@ -149,7 +149,7 @@ struct SenderConfig {
     //! Parameters for the port from which repair packets are sent.
     PortConfig repair_port;
 
-    //! Sample rate, number of samples for all channels per second.
+    //! Number of samples per second per channel.
     size_t sample_rate;
 
     //! Channel mask.
--- a/src/modules/roc_pipeline/receiver_session.cpp
+++ b/src/modules/roc_pipeline/receiver_session.cpp
@@ -19,6 +19,7 @@ namespace roc {
 namespace pipeline {
 
 ReceiverSession::ReceiverSession(const SessionConfig& config,
+                                 const size_t out_sample_rate,
                                  const packet::Address& src_address,
                                  const rtp::FormatMap& format_map,
                                  packet::PacketPool& packet_pool,
@@ -28,6 +29,8 @@ ReceiverSession::ReceiverSession(const SessionConfig& config,
     : src_address_(src_address)
     , allocator_(allocator)
     , audio_reader_(NULL) {
+    roc_panic_if(out_sample_rate == 0);
+
     const rtp::Format* format = format_map.format(config.payload_type);
     if (!format) {
         return;
@@ -35,7 +38,8 @@ ReceiverSession::ReceiverSession(const SessionConfig& config,
 
     if (config.resampling) {
         resampler_updater_.reset(new (allocator_) audio::ResamplerUpdater(
-                                     config.fe_update_interval, config.latency),
+                                     config.fe_update_interval, config.latency,
+                                     (float)format->sample_rate / out_sample_rate),
                                  allocator_);
         if (!resampler_updater_) {
             return;
--- a/src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
+++ b/src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
@@ -16,14 +16,12 @@
 namespace roc {
 namespace sndio {
 
-Player::Player(pipeline::IReceiver& input,
-               core::BufferPool<audio::sample_t>& buffer_pool,
+Player::Player(core::BufferPool<audio::sample_t>& buffer_pool,
                core::IAllocator& allocator,
                bool oneshot,
                packet::channel_mask_t channels,
                size_t sample_rate)
     : output_(NULL)
-    , input_(input)
     , buffer_pool_(buffer_pool)
     , allocator_(allocator)
     , clips_(0)
@@ -32,10 +30,6 @@ Player::Player(pipeline::IReceiver& input,
     size_t n_channels = packet::num_channels(channels);
     if (n_channels == 0) {
         roc_panic("player: # of channels is zero");
-    }
-
-    if (sample_rate == 0) {
-        roc_panic("player: sample rate is zero");
     }
 
     memset(&out_signal_, 0, sizeof(out_signal_));
@@ -71,6 +65,23 @@ bool Player::open(const char* name, const char* type) {
         return false;
     }
 
+    unsigned long in_rate = (unsigned long)out_signal_.rate;
+    unsigned long out_rate = (unsigned long)output_->signal.rate;
+
+    if (in_rate != 0 && in_rate != out_rate) {
+        roc_log(LogError,
+                "can't open output file or device with the required sample rate: "
+                "required=%lu suggested=%lu",
+                out_rate, in_rate);
+        return false;
+    }
+
+    roc_log(LogInfo,
+            "player:"
+            " bits=%lu out_rate=%lu in_rate=%lu ch=%lu",
+            (unsigned long)output_->encoding.bits_per_sample, out_rate, in_rate,
+            (unsigned long)output_->signal.channels);
+
     return true;
 }
 
@@ -78,11 +89,27 @@ void Player::stop() {
     stop_ = 1;
 }
 
+void Player::start(pipeline::IReceiver& input) {
+    input_ = &input;
+    core::Thread::start();
+}
+
+size_t Player::get_sample_rate() const {
+    if (!output_) {
+        roc_panic("player: can't get sample rate for non-open output file or device");
+    }
+    return size_t(output_->signal.rate);
+}
+
 void Player::run() {
     roc_log(LogDebug, "player: starting thread");
 
+    if (!input_) {
+        roc_panic("player: thread is started not from the start() call");
+    }
+
     if (!output_) {
-        roc_panic("player: thread is started before open() returnes success");
+        roc_panic("player: thread is started before open() returned success");
     }
 
     loop_();
@@ -115,7 +142,7 @@ void Player::loop_() {
     SOX_SAMPLE_LOCALS;
 
     while (!stop_) {
-        pipeline::IReceiver::Status status = input_.read(frame);
+        pipeline::IReceiver::Status status = input_->read(frame);
 
         if (status == pipeline::IReceiver::Inactive) {
             if (oneshot_ && n_bufs_ != 0) {
--- a/src/modules/roc_sndio/target_sox/roc_sndio/player.h
+++ b/src/modules/roc_sndio/target_sox/roc_sndio/player.h
@@ -35,11 +35,9 @@ public:
     //! Initialize.
     //!
     //! @b Parameters
-    //!  - @p input is used to read samples
     //!  - @p channels defines bitmask of enabled channels in input buffers
     //!  - @p sample_rate defines sample rate of input buffers
-    Player(pipeline::IReceiver& input,
-           core::BufferPool<audio::sample_t>& buffer_pool,
+    Player(core::BufferPool<audio::sample_t>& buffer_pool,
            core::IAllocator& allocator,
            bool oneshot,
            packet::channel_mask_t channels,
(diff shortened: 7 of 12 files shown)


src/lib/roc/config.h
```diff
@@ -105,6 +105,9 @@ typedef struct roc_receiver_config {
 
     //! A bitmask of ROC_FLAG_* constants.
     unsigned int flags;
+
+    //! Number of samples for all channels per second.
+    unsigned int sample_rate;
```
gavv: This comment looks ambiguous. I suggest "Number of samples per second per channel". We should also fix corresponding comments in ReceiverConfig and SenderConfig.

src/modules/roc_audio/resampler_updater.cpp
```diff
@@ -96,13 +98,15 @@ bool ResamplerUpdater::update(packet::timestamp_t time) {
         update_time_ += update_interval_;
     }
 
+    const float adjusted_coef = float(sample_rate_coef_ * fe_.freq_coeff());
+
     if (rate_limiter_.allow()) {
         roc_log(LogDebug, "resampler updater: queue_size=%lu fe=%.5f",
-                (unsigned long)queue_size, (double)fe_.freq_coeff());
+                (unsigned long)queue_size, (double)adjusted_coef);
```
gavv: It would be reasonable to print both freq_coeff and adjusted_coef here.

src/modules/roc_audio/resampler_updater.h
```diff
@@ -35,7 +35,8 @@ class ResamplerUpdater : public packet::IWriter,
     //!  - @p update_interval defines how often to call FreqEstimator, in samples
     //!  - @p aim_queue_size defines FreqEstimator target queue size, in samples
     ResamplerUpdater(packet::timestamp_t update_interval,
-                     packet::timestamp_t aim_queue_size);
+                     packet::timestamp_t aim_queue_size,
+                     const float sample_rate_coef);
```
gavv: sample_rate_coef parameter should be documented.

src/modules/roc_pipeline/config.h
```diff
@@ -135,7 +135,7 @@ struct ReceiverConfig {
     bool timing;
 
     ReceiverConfig()
-        : sample_rate(DefaultSampleRate)
+        : sample_rate(0)
```
gavv: Why this change? The idea of default constructors of these *Config structures is that if a field was not overwritten explicitly, it has a reasonable value. This feature will be used in roc_lib, which will overwrite only those fields that were set to non-zero by user.

src/modules/roc_pipeline/receiver_session.cpp
```diff
@@ -35,7 +36,8 @@ ReceiverSession::ReceiverSession(const SessionConfig& config,
 
     if (config.resampling) {
         resampler_updater_.reset(new (allocator_) audio::ResamplerUpdater(
-                                     config.fe_update_interval, config.latency),
+                                     config.fe_update_interval, config.latency,
+                                     (float)format->sample_rate / out_sample_rate),
```
gavv: We should panic if out_sample_rate is zero. A panic is better than a segfault.

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
```diff
@@ -71,13 +65,32 @@ bool Player::open(const char* name, const char* type) {
...
+            (unsigned long)output_->signal.channels);
+
     return true;
 }
 
 void Player::stop() {
     stop_ = 1;
 }
 
+void Player::start(pipeline::IReceiver* input) {
```
gavv: Why a pointer? It's not allowed to be null, so a reference is preferred.

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
```diff
@@ -71,13 +65,32 @@ bool Player::open(const char* name, const char* type) {
...
+    input_ = input;
+    core::Thread::start();
+}
+
+size_t Player::get_sample_rate() const {
+    if (output_) {
+        return size_t(output_->signal.rate);
+    }
+    return 0;
```
gavv: If output_ is null, it means that open() was not called or failed, and the user called get_sample_rate() anyway. It looks like a bug in the user code, so I think it's better to panic here instead of silently returning zero.

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
```diff
@@ -115,7 +128,7 @@ void Player::loop_() {
     SOX_SAMPLE_LOCALS;
 
     while (!stop_) {
-        pipeline::IReceiver::Status status = input_.read(frame);
+        pipeline::IReceiver::Status status = input_->read(frame);
```
gavv: We should panic if run() was called and input_ is null. A panic is better than a segfault.

src/modules/roc_sndio/target_sox/roc_sndio/player.h
```diff
@@ -60,11 +58,23 @@ class Player : public core::Thread {
...
+    //!  - @p input is used to read samples.
+    void start(pipeline::IReceiver* input);
+
     //! Stop thread.
     //! @remarks
     //!  Can be called from any thread.
     void stop();
 
+    //! Get sample rate
+    //!
```
gavv: It's worth to mention how this sample rate is obtained.

gavv: Also, we need a test for this feature.

src/tools/roc_recv/main.cpp
```diff
@@ -148,6 +148,17 @@ int main(int argc, char** argv) {
...
+    sndio::Player player(sample_buffer_pool, allocator, args.oneshot_flag,
+                         config.channels, config.sample_rate);
+
+    if (!player.open(args.output_arg, args.type_arg)) {
+        roc_log(LogError, "can't open output file or device: %s %s", args.output_arg,
+                args.type_arg);
+        return 1;
+    }
+
+    config.sample_rate = player.get_sample_rate();
```
gavv: I think we should check if sample_rate is zero and exit with error in this case.

src/lib/roc/config.h
dshil: Fixed

src/modules/roc_audio/resampler_updater.h
dshil: Fxd

src/modules/roc_audio/resampler_updater.cpp
dshil: Fixed

src/modules/roc_pipeline/receiver_session.cpp
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.h
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
dshil: Fixed

src/modules/roc_pipeline/receiver_session.cpp
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
dshil: Fixed

src/modules/roc_pipeline/receiver_session.cpp
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.h
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
dshil: Fixed

src/tools/roc_recv/main.cpp
dshil: Fixed

src/modules/roc_pipeline/config.h
dshil: Providing zero sample rate for sox means, that it'll determine sample rate of the output file or device by its own.

UPD: Later I've checked, that if we provide unsupported sample rate to sox (e.g. 44K, while device supports 48K), sox will also detect the correct sample rate by its own.

I'll rollback this change.

dshil: > This feature will be used in roc_lib, which will overwrite only those fields that were set to non-zero by user.

Where do we check that the user provides non-zero values?

```cpp
bool make_receiver_config(pipeline::ReceiverConfig& out, const roc_receiver_config* in) {
    out.default_session.latency = in->latency;
    out.default_session.timeout = in->timeout;
    out.default_session.samples_per_packet = in->samples_per_packet;
    out.sample_rate = in->sample_rate;
   
    //...
}
```

We'll panic only during a session creation.

dshil:
ping @gavv

src/modules/roc_pipeline/config.h
gavv: > Later I've checked, that if we provide unsupported sample rate to sox (e.g. 44K, while device supports 48K), sox will also detect the correct sample rate by its own.

I think we should not rely on this, it's too implicit and can be changed when switching from SoX to something else.

I suggest the following:

* Use a reasonable default in ReceiverConfig::sample_rate (44100). This default will be ignored in roc_recv, but it will be used in roc_lib.

* Use a separate variable in roc_recv, initialize it with `--rate` option value or zero if it wan't given, pass it to SoX, then get the actual sample rate from SoX, and write it to ReceiverConfig.

> Where do we check that the user provides non-zero values?

There are no checks yet. However, I plan to add them in 0.1. It will look like:

```
if (in->sample_rate) {
    out.sample_rate = in->sample_rate;
}
```

src/modules/roc_audio/resampler_updater.cpp
```diff
@@ -96,13 +98,16 @@ bool ResamplerUpdater::update(packet::timestamp_t time) {
         update_time_ += update_interval_;
     }
 
+    const float adjusted_coef = float(sample_rate_coef_ * fe_.freq_coeff());
```
gavv: Why do we need a cast?

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
```diff
@@ -71,18 +69,41 @@ bool Player::open(const char* name, const char* type) {
...
+        roc_panic("player: can't get sample rate for non-open output file or device");
+    }
+    return size_t(output_->signal.rate);
+}
+
 void Player::run() {
     roc_log(LogDebug, "player: starting thread");
 
+    if (!input_) {
+        roc_panic("player: thread is started before device or output file was opened");
```
gavv: The panic message is a bit misleading, because input_ is set in start(), not in open().

src/modules/roc_sndio/target_sox/roc_sndio/player.h
```diff
@@ -60,11 +58,23 @@ class Player : public core::Thread {
     //!  Should be called once before calling start().
     bool open(const char* name, const char* type = NULL);
 
+    //! Start reading samples.
```
gavv: It's worth to mention that it starts a thread.

src/modules/roc_sndio/target_sox/roc_sndio/player.h
```diff
@@ -60,11 +58,23 @@ class Player : public core::Thread {
...
+
     //! Stop thread.
     //! @remarks
     //!  Can be called from any thread.
     void stop();
 
+    //! Get a user defined or auto-detected sample rate of an output file or a device.
+    //!
+    //! @remarks
+    //! Output file or device should be opened.
```
gavv: Two little notices:

* We can change `@remarks` to `@pre` (preconditions) here, it would fit a bit better.
* We usually indent `@remarks`, `@pre`, and similar sections, e.g.:

    ```
    //! @pre
    //!  Output file or device should be opened.
    ```

src/tests/roc_lib/test_sender_receiver.cpp
```diff
@@ -223,6 +224,7 @@ TEST_GROUP(sender_receiver) {
         receiver_conf.n_repair_packets = n_repair_packets;
         receiver_conf.latency = packet_len * 20;
         receiver_conf.timeout = packet_len * 300;
+        receiver_conf.sample_rate = pipeline::DefaultSampleRate;
```
gavv: Since we have a default back, this line can be removed.

src/tests/roc_sndio/target_sox/test_player_recorder.cpp
```diff
@@ -115,41 +115,47 @@ TEST_GROUP(player_recorder) {
...
     core::TempFile file("test.wav");
     CHECK(player.open(file.path()));
 
     player.stop();
-    player.start();
+    player.start(receiver);
     player.join();
 }
 
+TEST(player_recorder, player_get_sample_rate) {
```
gavv: I think we should test three cases:

* The user provided zero sample rate, SoX selected non-zero sample rate, Player was successfully opened.

* The user provided non-zero sample rate, SoX selected the same sample rate, Player was successfully opened.

* The user provided non-zero sample rate, but SoX selected some other sample rate. I think Player::open should fail in this case, because it's generally not a good idea to ignore what the user have asked explicitly.

I'm not sure if it's possible to add a test for the latter case because our tests operate with files instead of devices. However, we need to fix the behaviour anyway.

src/tools/roc_recv/main.cpp
```diff
@@ -148,6 +148,22 @@ int main(int argc, char** argv) {
     core::BufferPool<audio::sample_t> sample_buffer_pool(allocator, MaxFrameSize, 1);
     packet::PacketPool packet_pool(allocator, 1);
 
+    sndio::Player player(sample_buffer_pool, allocator, args.oneshot_flag,
+                         config.channels, config.sample_rate);
```
gavv: See https://github.com/roc-project/roc/pull/113#discussion_r145893832

src/tests/roc_lib/test_sender_receiver.cpp
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.h
dshil: Fixed

src/modules/roc_sndio/target_sox/roc_sndio/player.h
dshil: Fixed

src/modules/roc_audio/resampler_updater.cpp
dshil: o my gosh, at the beginning I had a crazy idea, that we can overflow here, but after your comment, I've reread the code. Both values are near about 1, we should throw this cast.

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
dshil: fixed

src/modules/roc_pipeline/config.h
dshil: Fixed

src/tests/roc_sndio/target_sox/test_player_recorder.cpp
dshil: Fixed

src/modules/roc_audio/resampler_updater.cpp
gavv: I don't get it. If both arguments are floats, this cast simply does nothing.

src/modules/roc_audio/resampler_updater.h
```diff
@@ -34,8 +34,11 @@ class ResamplerUpdater : public packet::IWriter,
     //! @b Parameters
     //!  - @p update_interval defines how often to call FreqEstimator, in samples
     //!  - @p aim_queue_size defines FreqEstimator target queue size, in samples
+    //!  - @p sample_rate_coef represents an adjustment coefficient for a resampler
+    //!       scaling factor, calculating from sample rates of input and output devices.
```
gavv: > from sample rates of input and output devices

Well, strictly speaking it's not rates of the input and output devices. It's a faсtor to convert session rate to receiver output rate.

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
```diff
@@ -71,18 +65,49 @@ bool Player::open(const char* name, const char* type) {
         return false;
     }
 
+    size_t in_rate = size_t(out_signal_.rate);
+    size_t out_rate = size_t(output_->signal.rate);
+
+    if (in_rate != 0 && in_rate != out_rate) {
+        roc_log(LogError, "player: sample rates mismatch, auto-detected=%zu, input=%zu",
```
gavv: I think the error message should be something like:

"can't open output file or device with the required sample rate: required=%lu suggested=%lu".

Also some notes:

* we don't use %z, it's not portable
* we don't use commas when logging multiple values, they don't increase readability, but they make grepping a bit harder

src/tests/roc_lib/test_sender_receiver.cpp
```diff
@@ -22,6 +22,7 @@
 #include "roc_packet/address_to_str.h"
 #include "roc_packet/packet_pool.h"
 #include "roc_packet/parse_address.h"
+#include "roc_pipeline/config.h"
```
gavv: This line can be removed I guess.

src/tests/roc_sndio/target_sox/test_player_recorder.cpp
```diff
@@ -115,41 +115,60 @@ TEST_GROUP(player_recorder) {
...
+    Player player(buffer_pool, allocator, true, ChMask, SampleRate);
+
+    core::TempFile file("test.wav");
+    CHECK(player.open(file.path()));
+    CHECK(player.get_sample_rate() == SampleRate);
+}
+
+TEST(player_recorder, player_open_sample_rates_mistmatch) {
+    Player player(buffer_pool, allocator, true, ChMask, SampleRate + 1);
+    CHECK(!player.open(NULL, NULL));
```
gavv: There are three problems with this test:

* Opening an audio device may have side-effects. For example, if the user is using ALSA without dmix, only one process can open an audio device. So this test will either fail to open it (if it's already opened by someone else), or prevent other processes from opening it for a short time.

* If this test will fail to open device for some other reason (e.g. an ALSA device is already opened by someone else), it will pass anyway, so we really don't known if it actually tests something.

* I'm not sure that any SoX backend is guaranteed to fail with rate 44100+1.

We theoretically may want a separate set of tests that work with a real audio device. Such tests should not be run automatically, but instead only when the user explicitly asked that. However I'm not sure it's worth it.

Thus, if we can't cover this case with a test that uses a temporary file instead of real audio device, I suggest to remove it for now.

dshil: I think that you've reviewed an old version of tests, I've rebased recently.

gavv: The first two points are still relevant after rebase.

src/modules/roc_sndio/target_sox/roc_sndio/player.cpp
dshil: Fixed

src/tests/roc_sndio/target_sox/test_player_recorder.cpp
dshil: Fixed

src/tests/roc_lib/test_sender_receiver.cpp
dshil: Fixed

src/modules/roc_audio/resampler_updater.h
dshil: Fixed

PR #118, opened 2017-10-26
roc_pipelie: sender should obtain sample rate from payload type

dshil:
Relates: #85


gavv:
LGTM. See also dbe3c0cef79615ba7cc89e213c1e9e66d9989675

PR #120, opened 2017-11-04
roc_pipeline: drop repair packets before session creation

dshil:
#85

### DIFF
--- a/src/modules/roc_pipeline/config.h
+++ b/src/modules/roc_pipeline/config.h
@@ -122,9 +122,6 @@ struct SessionConfig {
 
 //! Receiver parameters.
 struct ReceiverConfig {
-    //! Parameters for the port from which repair packets are sent.
-    PortConfig repair_port;
-
     //! Default parameters for session.
     SessionConfig default_session;
 
--- a/src/modules/roc_pipeline/config.h
+++ b/src/modules/roc_pipeline/config.h
@@ -122,6 +122,9 @@ struct SessionConfig {
 
 //! Receiver parameters.
 struct ReceiverConfig {
+    //! Parameters for the port from which repair packets are sent.
+    PortConfig repair_port;
+
     //! Default parameters for session.
     SessionConfig default_session;
 
--- a/src/modules/roc_pipeline/receiver.cpp
+++ b/src/modules/roc_pipeline/receiver.cpp
@@ -152,6 +152,12 @@ bool Receiver::create_session_(const packet::PacketPtr& packet) {
         roc_log(LogError, "receiver: can't create session, unexpected non-udp packet");
         return false;
     }
+
+    if (packet->udp()->dst_addr == config_.repair_port.address) {
+        roc_log(LogInfo, "receiver: can't create session, unexpected repair packet");
+        return false;
+    }
+
     const packet::Address src_address = packet->udp()->src_addr;
 
     core::SharedPtr<ReceiverSession> sess = new (allocator_) ReceiverSession(
--- a/src/modules/roc_pipeline/receiver.cpp
+++ b/src/modules/roc_pipeline/receiver.cpp
@@ -158,6 +158,11 @@ bool Receiver::create_session_(const packet::PacketPtr& packet) {
         return false;
     }
 
+    if (!packet->rtp()) {
+        roc_log(LogError, "receiver: can't create session, unexpected non-rtp packet");
+        return false;
+    }
+
     const packet::Address src_address = packet->udp()->src_addr;
 
     core::SharedPtr<ReceiverSession> sess = new (allocator_)
--- a/src/modules/roc_pipeline/receiver.cpp
+++ b/src/modules/roc_pipeline/receiver.cpp
@@ -160,9 +160,10 @@ bool Receiver::create_session_(const packet::PacketPtr& packet) {
 
     const packet::Address src_address = packet->udp()->src_addr;
 
-    core::SharedPtr<ReceiverSession> sess = new (allocator_) ReceiverSession(
-        config_.default_session, config_.sample_rate, src_address, format_map_,
-        packet_pool_, byte_buffer_pool_, sample_buffer_pool_, allocator_);
+    core::SharedPtr<ReceiverSession> sess = new (allocator_)
+        ReceiverSession(config_.default_session, config_.sample_rate, src_address,
+                        format_map_, packet->rtp()->payload_type, packet_pool_,
+                        byte_buffer_pool_, sample_buffer_pool_, allocator_);
 
     if (!sess || !sess->valid()) {
         roc_log(LogError, "receiver: can't create session, initialization failed");
--- a/src/modules/roc_pipeline/receiver.cpp
+++ b/src/modules/roc_pipeline/receiver.cpp
@@ -153,8 +153,8 @@ bool Receiver::create_session_(const packet::PacketPtr& packet) {
         return false;
     }
 
-    if (packet->udp()->dst_addr == config_.repair_port.address) {
-        roc_log(LogInfo, "receiver: can't create session, unexpected repair packet");
+    if (packet->flags() & packet::Packet::FlagRepair) {
+        roc_log(LogDebug, "receiver: dropping repair packet for non-existent session");
         return false;
     }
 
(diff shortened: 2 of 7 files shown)


src/modules/roc_pipeline/config.h
```diff
@@ -122,6 +122,9 @@ struct SessionConfig {
 
 //! Receiver parameters.
 struct ReceiverConfig {
+    //! Parameters for the port from which repair packets are sent.
+    PortConfig repair_port;
```
gavv: Receiver can have an arbitrary number of source and repair ports. That's why we don't have PortConfig is ReceiverConfig. Instead, ports are added dynamically using add_port().

src/modules/roc_pipeline/receiver.cpp
```diff
@@ -146,11 +146,18 @@ bool Receiver::create_session_(const packet::PacketPtr& packet) {
         roc_log(LogError, "receiver: can't create session, unexpected non-udp packet");
         return false;
     }
+
+    if (packet->udp()->dst_addr == config_.repair_port.address) {
```
gavv: This is not a correct way to check if the packet is repair (see comment above). The correct way would be the following:

```
if (packet->flags() & packet::Packet::FlagRepair) {
    ...
}
```

src/modules/roc_pipeline/receiver.cpp
```diff
@@ -146,11 +146,18 @@ bool Receiver::create_session_(const packet::PacketPtr& packet) {
...
+        roc_log(LogInfo, "receiver: can't create session, unexpected repair packet");
```
gavv: We usually don't use LogInfo for per-packet messages because they could be quite frequent. I'd change it to LogDebug here.

Also, the log message is a bit unclear. The packet is not unexpected. Actually, it's quite expected, we just don't want to handle it. I'd change it to something like:

```
receiver: dropping repair packet for non-existent session
```

src/modules/roc_pipeline/receiver.cpp
```diff
@@ -146,11 +146,18 @@ bool Receiver::create_session_(const packet::PacketPtr& packet) {
...
+    }
+
     const packet::Address src_address = packet->udp()->src_addr;
 
-    core::SharedPtr<ReceiverSession> sess = new (allocator_) ReceiverSession(
-        config_.default_session, config_.sample_rate, src_address, format_map_,
-        packet_pool_, byte_buffer_pool_, sample_buffer_pool_, allocator_);
+    core::SharedPtr<ReceiverSession> sess = new (allocator_)
+        ReceiverSession(config_.default_session, config_.sample_rate, src_address,
+                        format_map_, packet->rtp()->payload_type, packet_pool_,
```
gavv: We assume here that this is an RTP packet. Since currently all supported source packets are RTP it's ok, but we need to add a check to create_session_():

```
    if (!packet->rtp()) {
        roc_log(LogError, "receiver: can't create session, unexpected non-rtp packet");
        return false;
    }
```

src/modules/roc_pipeline/receiver_session.h
```diff
@@ -52,6 +52,7 @@ class ReceiverSession : public core::RefCnt<ReceiverSession>, public core::ListN
                     const size_t out_sample_rate,
                     const packet::Address& src_address,
                     const rtp::FormatMap& format_map,
+                    const unsigned int payload_type,
```
gavv: Nitpick: I'd move this field near `ReceiverConfig config_`, which contains other related session parameters, like sample rate and channel mask.

src/tests/roc_pipeline/test_receiver.cpp
```diff
@@ -129,6 +129,65 @@ TEST(receiver, no_ports) {
     }
 }
 
+TEST(receiver, no_session_repair_port) {
```
gavv: It seems that the first test is complete covered by the second test. I suggest to remove the first one.

src/tests/roc_pipeline/test_receiver.cpp
```diff
@@ -129,6 +129,65 @@ TEST(receiver, no_ports) {
...
+    FrameReader frame_reader(receiver, sample_buffer_pool);
+
+    for (size_t nf = 0; nf < ManyPackets * FramesPerPacket; nf++) {
+        frame_reader.skip_zeros(SamplesPerFrame * NumCh);
+
+        UNSIGNED_LONGS_EQUAL(0, receiver.num_sessions());
+    }
+}
+
+TEST(receiver, one_session_source_after_repair) {
```
gavv: I suggest to rename it to one_session_drop_leading_repair_packets.

src/tests/roc_pipeline/test_receiver.cpp
```diff
@@ -129,6 +129,65 @@ TEST(receiver, no_ports) {
...
+    config.repair_port = port1;
+
+    Receiver receiver(config, format_map, packet_pool, byte_buffer_pool,
+                      sample_buffer_pool, allocator);
+
+    CHECK(receiver.valid());
+    CHECK(receiver.add_port(port1));
+
+    PacketWriter repair_packet_writer(receiver, rtp_composer, pcm_encoder, packet_pool,
+                                      byte_buffer_pool, PayloadType, src1, port1.address);
```
gavv: Well, this writer won't produce repair packets, it'll produce source packets instead. The test works because you match packets by their address instead of their type.

To produce repair packets, you need an appropriate composer. See list of the supported composers here: https://github.com/roc-project/roc/blob/develop/src/modules/roc_pipeline/sender_port.cpp

src/tests/roc_pipeline/test_receiver.cpp
```diff
@@ -129,6 +129,65 @@ TEST(receiver, no_ports) {
...
+
+    FrameReader frame_reader(receiver, sample_buffer_pool);
+
+    for (size_t nf = 0; nf < ManyPackets * FramesPerPacket; nf++) {
+        frame_reader.skip_zeros(SamplesPerFrame * NumCh);
+
+        UNSIGNED_LONGS_EQUAL(0, receiver.num_sessions());
+    }
+
+    CHECK(receiver.add_port(port2));
```
gavv: I think we can move this add_port() call to the top near the first one.

src/modules/roc_pipeline/config.h
dshil: 455a825301accba5d3f79c7d45ef4315171d0a3f

src/modules/roc_pipeline/receiver.cpp
dshil: 455a825301accba5d3f79c7d45ef4315171d0a3f

src/modules/roc_pipeline/receiver.cpp
dshil: 455a825301accba5d3f79c7d45ef4315171d0a3f

src/modules/roc_pipeline/receiver.cpp
dshil: 78b04f7aa7c4b841a68a7095f30c0f62123cc258

src/modules/roc_pipeline/receiver_session.h
dshil: faa175e0022dcfee68511bbe6a36e018b3428b86

src/tests/roc_pipeline/test_receiver.cpp
dshil: 455a825

src/tests/roc_pipeline/test_receiver.cpp
dshil: 455a825

src/tests/roc_pipeline/test_receiver.cpp
dshil: 455a825

src/tests/roc_pipeline/test_receiver.cpp
dshil: 455a825

dshil:
@gavv, I suppose, that `PayloadType` should be removed from a session config, because we'll obtain it from a packet, but it's still used in tests during packet construction.

gavv:
Thanks. Merged manually: 28cdcc40f0b7c3bb98ac8422f071d755ead6a4c2

I've also simplified tests and removed SessionConfig::PayloadType.

Issue #63, opened 2016-05-15
Enable CPU timer when and only when working with files


baranovmv:
There is a doubt if it is correct to feed sample queue between threads by absolute timer

gavv:
PulseAudio and ALSA clock clients in the sound card domain. So we should disable our own timing when writing sound to a local device.

We still need our own timing when sending samples over RTP/UDP.

CUrrently, timing is enabled or disabled manually by user (via command line option). We should do it automatically instead.

baranovmv:
- This automation should be implemented in **roc/tools** only.
-  CAPI is to be provided by `self_timning` argument of its `*_open` functions.

gavv:
See also #100.

gavv:
### TODO

- [ ] roc-recv
- [ ] roc-send

PR #121, opened 2017-11-05
roc-recv: enable timing only when output is file

dshil:
#63


gavv:
Thanks. See 6ce5fcf418b22eebc4bad272a3363f0a7f7aee57

We also need the same for roc-send, see https://github.com/roc-project/roc/issues/63#issuecomment-347113386

dshil:
@gavv, notice that I that working on adding the same option for `roc-send`.

Issue #115, opened 2017-10-22
Rework Watchdog


gavv:
- [x] Add flags support to audio::Frame, just like we have flags in packet::Packet.

  We need three new flags:

  * `Empty` - set if no packets were extracted to the frame when the frame was built

  * `Full` - set if the frame is fully filled with samples from packets

  * `PacketDrops` - set if some queued packets were regarded as outdated and dropped when the frame was built

- [x] Set these flags in audio::Depacketizer

- [x] Replace packet::Watchdog (that works with packets) with audio::Watchdog (that works with frames). The new watchdog should detect two situations:

  - [x] When during some period all frames has Empty flag, the session should be shutted down. This will work the same way as the current watchdog.

  - [x] When during some period every batch of N frames includes at least one frame with PacketDrops flag and without Full flag, the session should be shutted down. This will fix #109.

- [x] Add command line options to configure new watchdog and test it.

gavv:
Done. Testing is described in #92 and #109.

PR #122, opened 2017-11-09
roc_audio: rework frame

dshil:
Relates: #115

### DIFF
--- a/src/modules/roc_audio/frame.h
+++ b/src/modules/roc_audio/frame.h
@@ -20,9 +20,43 @@ namespace roc {
 namespace audio {
 
 //! Audio frame.
-struct Frame {
-    //! Frame samples.
-    core::Slice<sample_t> samples;
+class Frame {
+public:
+    //! Construct empty frame.
+    Frame();
+
+    //! Construct frame from samples.
+    Frame(const core::Slice<sample_t>& samples);
+
+    //! Frame flags.
+    enum {
+        //! Set if no packets were extracted to the frame when the frame was built.
+        FlagEmpty = (1 << 0),
+
+        //! Set if some queued packets were regarded as outdated and dropped when the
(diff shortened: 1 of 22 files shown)


dshil:
@gavv, I saw a72b0d8d, where mixer was reworked. Why did you choose to log instead of panic when we can't allocate a temp buffer?

gavv:
> Why did you choose to log instead of panic when we can't allocate a temp buffer?

We can't panic there because the allocation may fail if we will use a memory allocator with a limited size.

> I'm confused because roc-recv will simply log info about it infinitely and nothing else.

Agree, this behaviour is quite strange, though better than a panic.

We can improve it by moving the allocation to the Mixer constructor and adding valid() method that will report whether the object is constructed successfully (just like we do for other objects in roc_pipeline and roc_sndio). Then we should check it in the session constructor and refuse to create session if we failed to construct the mixer.

src/modules/roc_audio/frame.h
```diff
@@ -20,9 +20,43 @@ namespace roc {
...
+        //! Set if some queued packets were regarded as outdated and dropped when the
+        //! frame was built.
+        FlagSkip = (1 << 1)
+    };
+
+    //! Add flags.
+    void add_flags(unsigned flags);
+
+    //! Returns true if frame has no packets.
+    bool is_empty() const;
```
gavv: I'd rather add a single flags() method instead of adding a getter method for every flag. The advantage of the first approach are:

* it's easier to grep all occurrences of a flag in the source code
* it's clear that we just check a flag that was set previously with add_flags() and no other magic happens

gavv: Fixed  in e5282d5358d4a150f3b393acd3eded5ade763ef5

PR #123, opened 2017-11-11
roc_audio: set frame flags

dshil:
@gavv, this PR is based on #122.

### DIFF
--- a/src/modules/roc_audio/depacketizer.cpp
+++ b/src/modules/roc_audio/depacketizer.cpp
@@ -47,7 +47,8 @@ Depacketizer::Depacketizer(packet::IReader& reader,
     , packet_samples_(0)
     , rate_limiter_(LogRate)
     , first_packet_(true)
-    , beep_(beep) {
+    , beep_(beep)
+    , dropped_packets_(0) {
 }
 
 void Depacketizer::read(Frame& frame) {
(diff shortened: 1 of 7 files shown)


dshil:
ping @gavv

src/modules/roc_audio/depacketizer.cpp
```diff
@@ -208,5 +216,19 @@ packet::PacketPtr Depacketizer::read_packet_() {
...
 
+void Depacketizer::set_frame_flags_(Frame& frame,
+                                    const size_t dropped_packets,
+                                    const packet::timestamp_t missing_samples) {
+    if (dropped_packets != dropped_packets_) {
+        frame.add_flags(Frame::FlagSkip);
+    }
+
+    const size_t diff = (size_t)ROC_UNSIGNED_SUB(packet::signed_timestamp_t,
+                                                 missing_samples_, missing_samples);
```
gavv: We don't need ROC_UNSIGNED_SUB here, because we always cast the result to the unsigned type. It was also unnecessary in some other places, I've removed them: 71d93382df1e916e06dd2b1aa503c3d6ae8cae63

src/tests/roc_audio/test_depacketizer.cpp
```diff
@@ -189,6 +189,73 @@ TEST(depacketizer, drop_late_packets) {
...
+    queue.write(new_packet(ts1, 0.11f));
+    queue.write(new_packet(ts2, 0.22f));
+
+    Frame frame = new_frame(SamplesPerPacket * 2);
+    dp.read(frame);
+
+    CHECK(!(frame.flags() & Frame::FlagSkip));
+}
+
+TEST(depacketizer, empty_frame) {
```
gavv: Could you also add a test for the case when Frame::FlagEmpty is *not* set?

dshil: I'll expand the current test

src/modules/roc_audio/depacketizer.cpp
dshil: Fixed

PR #125, opened 2017-12-04
roc_audio: improve watchdog

### LINKED ISSUE #109: Shutdown session when depacketizer drops all packets
Currently it's possible that the packet stream is delayed a bit and depacketizer timestamp goes ahead it, and depacketizer starts dropping all packets that it gets from fec reader forever.

We should either fix it or at least detect it and shutdown the session, so it could be automatically restarted.

dshil:
> When during some period every batch of N frames includes at least one frame with Skip flag, the session should be shutted down. This will fix #109.

I made some tests but I'm not sure which `N` should be used.

I suppose that the corresponding option should be added to `lib/roc/config.h` for `roc_receiver_config`.

#115


dshil:
I've decided to split this PR into several.

PR #128, opened 2017-12-09
depacketizer: rework setting frame flags when there is no first packet

dshil:
Relates: #115


dshil:
It's implemented in #130

PR #129, opened 2017-12-09
roc_audio: watchdog works with frames

dshil:
Relates: #115


dshil:
ping @gavv, I've rebased it.

src/tests/roc_audio/test_watchdog.cpp
```diff
@@ -0,0 +1,99 @@
...
+    Watchdog watchdog(test_reader, Timeout);
+
+    CHECK(watchdog.update(0));
+    check_read(0, watchdog, true);
+
+    CHECK(!watchdog.update(Timeout + 1));
+    check_read(0, watchdog, false);
+}
+
+TEST(watchdog, read_empty_frame) {
```
gavv: I think we should have two tests:

* `empty_frames_timeout` - pass `Timeout` empty frames, the session is terminated
* `empty_frames_no_timeout` - pass `Timeout - 1` empty frames, then one non-empty frame, the session is not terminated; repeat this several times

src/tests/roc_audio/test_watchdog.cpp
```diff
@@ -0,0 +1,99 @@
...
+TEST(watchdog, read) {
+    Watchdog watchdog(test_reader, Timeout);
+
+    check_read(0, watchdog, true);
+
+    CHECK(watchdog.update(Timeout / 4));
+    check_read(0, watchdog, true);
+}
+
+TEST(watchdog, read_timeout) {
```
gavv: I suggest to enhance `read_timeout` and `read` tests a bit:

* `read_timeout1` - call `update(n++)` `Timeout` times, never call read; the session is terminated
* `read_timeout2` - call `update(Timeout+1)` once, never call `read()`; the session is terminated (this is the current `read_timeout` test)
* `read_no_timeout` - call `update(n++)` `Timeout - 1` times, then call `read()` once, then call `update(n++)` once more; the session is not terminated; repeat this several times

src/tests/roc_audio/test_watchdog.cpp
```diff
@@ -0,0 +1,99 @@
...
+    void read(Frame& frame) {
+        memset(frame.samples().data(), 0, frame.samples().size() * sizeof(sample_t));
+    }
+};
+
+TestFrameReader test_reader;
+
+} // namespace
+
+TEST_GROUP(watchdog) {
```
gavv: We should also cover the case when the first timestamp passed to `update()` is larger than `Timeout` (covered by this [bit](https://github.com/dshil/roc/blob/715debb788f1413fbe4503c6c353f23354527fff/src/modules/roc_audio/watchdog.cpp#L45))

src/tests/roc_audio/test_watchdog.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    void check_read(unsigned frame_flags, IReader& reader, bool is_read) {
+        Frame frame = new_frame(SamplesPerPacket);
+        frame.add_flags(frame_flags);
+
+        reader.read(frame);
+
+        if (is_read) {
+            for (size_t n = 0; n < frame.samples().size(); n++) {
+                DOUBLES_EQUAL(0.0f, frame.samples().data()[n], 0);
```
gavv: This doesn't guarantee that Watchdog actually performed a read. Better fill the frame with non-zero values in TestFrameReader and then check them here. Also it's make sense to check for zero values when `is_read` is false.

src/tests/roc_audio/test_watchdog.cpp
```diff
@@ -0,0 +1,99 @@
...
+core::BufferPool<sample_t> sample_buffer_pool(allocator, MaxBufSize, 1);
+
+class TestFrameReader : public IReader, public core::NonCopyable<> {
+public:
+    void read(Frame& frame) {
+        memset(frame.samples().data(), 0, frame.samples().size() * sizeof(sample_t));
+    }
+};
+
+TestFrameReader test_reader;
```
gavv: We can move this object into `TEST_GROUP(watchdog)`.

src/tests/roc_audio/test_watchdog.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+namespace {
+
+enum {
+    MaxBufSize = 5000,
+
+    SamplesPerFrame = 200,
+    NumCh = 2,
+
+    Timeout = 20
```
gavv:
```
    SamplesPerFrame = 200,
    Timeout = 20
```

I don't think it makes any sense to test the case when frame is larger than timeout.

src/tests/roc_audio/test_watchdog.cpp
```diff
@@ -0,0 +1,155 @@
...
+    Watchdog watchdog(test_reader, Timeout);
+
+    for (packet::timestamp_t n = 0; n < Timeout; n++) {
+        CHECK(watchdog.update(n));
+    }
+
+    CHECK(!watchdog.update(Timeout));
+}
+
+TEST(watchdog, update_no_timeout) {
```
gavv: This test is identical to empty_frames_no_timeout..

PR #142, opened 2017-12-19
ensure to enable resampling for different in and out sample rates

dshil:
Relates: #113

### DIFF
--- a/src/modules/roc_audio/latency_monitor.cpp
+++ b/src/modules/roc_audio/latency_monitor.cpp
@@ -48,6 +48,14 @@ LatencyMonitor::LatencyMonitor(const packet::SortedQueue& queue,
         if (!init_resampler_(input_sample_rate, output_sample_rate)) {
             return;
         }
+    } else {
+        if (input_sample_rate != output_sample_rate) {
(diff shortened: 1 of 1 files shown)


src/modules/roc_pipeline/receiver_session.cpp
```diff
@@ -126,6 +126,14 @@ ReceiverSession::ReceiverSession(const SessionConfig& config,
     }
 #endif // ROC_TARGET_OPENFEC
 
+    if (format->sample_rate != out_sample_rate && !config.resampling) {
+        roc_log(LogError,
+                "receiver_session: input and output sample rates must be equal"
```
gavv: This is quite a low level check. ReceiverSession is a high level component that just brings other components together, so I think we should move this check somewhere else.

Actually, we already have a place where we check input and output rates when resampler is enabled: https://github.com/roc-project/roc/blob/develop/src/modules/roc_audio/latency_monitor.cpp#L48, so i think we should move this check to LatencyMonitor as well.

dshil: Done.

PR #157, opened 2019-01-05
Implement frame poisoning

gavv:
This PR adds the --poisoning flag. When enabled, every component that creates new frames begins to "poison" them (just fill with ones) before passing further to the pipeline. This simple technique helps to detect components that leaves some frame samples uninitialized, because now these samples become more audible.

@dshil could you please review this?

### DIFF
--- a/src/modules/roc_pipeline/config.h
+++ b/src/modules/roc_pipeline/config.h
@@ -141,10 +141,14 @@ struct ReceiverConfig {
     //! Constrain receiver speed using a CPU timer according to the sample rate.
     bool timing;
 
+    //! Fill uninitialized data with large values to make them more noticeable.
+    bool poisoning;
+
     ReceiverConfig()
         : sample_rate(DefaultSampleRate)
         , channels(DefaultChannelMask)
-        , timing(false) {
+        , timing(false)
+        , poisoning(false) {
     }
 };
 
@@ -183,6 +187,9 @@ struct SenderConfig {
     //! Constrain receiver speed using a CPU timer according to the sample rate.
     bool timing;
 
+    //! Fill unitialized data with large values to make them more noticable.
+    bool poisoning;
+
     SenderConfig()
         : sample_rate(DefaultSampleRate)
         , channels(DefaultChannelMask)
@@ -190,7 +197,8 @@ struct SenderConfig {
         , samples_per_packet(DefaultPacketSize)
         , resampling(false)
         , interleaving(false)
-        , timing(false) {
+        , timing(false)
+        , poisoning(false) {
     }
 };
 
--- a/src/modules/roc_pipeline/config.h
+++ b/src/modules/roc_pipeline/config.h
@@ -79,10 +79,10 @@ struct PortConfig {
     }
 };
 
-//! Session parameters.
+//! Receiver session parameters.
 //! @remarks
 //!  Defines per-session parameters on the receiver side.
-struct SessionConfig {
+struct ReceiverSessionConfig {
     //! Channel mask.
     packet::channel_mask_t channels;
 
@@ -107,19 +107,11 @@ struct SessionConfig {
     //! Resampler parameters.
     audio::ResamplerConfig resampler;
 
-    //! Perform resampling to to compensate sender and receiver frequency difference.
-    bool resampling;
-
-    //! Insert weird beeps instead of silence on packet loss.
-    bool beeping;
-
-    SessionConfig()
+    ReceiverSessionConfig()
         : channels(DefaultChannelMask)
         , samples_per_packet(DefaultPacketSize)
         , latency(DefaultPacketSize * 27)
-        , watchdog(DefaultSampleRate)
-        , resampling(false)
-        , beeping(false) {
+        , watchdog(DefaultSampleRate) {
         latency_monitor.min_latency =
             (packet::signed_timestamp_t)latency * DefaultMinLatency;
         latency_monitor.max_latency =
@@ -127,16 +119,18 @@ struct SessionConfig {
     }
 };
 
-//! Receiver parameters.
-struct ReceiverConfig {
-    //! Default parameters for session.
-    SessionConfig default_session;
-
+//! Receiver output parameters.
+//! @remarks
+//!  Defines common output parameters on the receiver side.
+struct ReceiverOutputConfig {
     //! Number of samples per second per channel.
     size_t sample_rate;
 
     //! Channel mask.
     packet::channel_mask_t channels;
+
+    //! Perform resampling to compensate sender and receiver frequency difference.
+    bool resampling;
 
     //! Constrain receiver speed using a CPU timer according to the sample rate.
     bool timing;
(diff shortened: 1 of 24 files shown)


src/modules/roc_pipeline/sender.cpp
```diff
@@ -130,7 +130,7 @@ Sender::Sender(const SenderConfig& config,
     if (config.resampling && config.sample_rate != format->sample_rate) {
         resampler_.reset(new (allocator) audio::ResamplerWriter(
                              *awriter, sample_buffer_pool, allocator, config.resampler,
-                             config.channels),
+                             config.channels, config.poisoning),
```
dshil: Don't we want to introduce some stupid PoisonReader, PoisonWriter? As it's a
helper mechanism I want the internal components to know about it as less as
possible.

src/modules/roc_audio/resampler_writer.cpp
```diff
@@ -73,8 +75,13 @@ void ResamplerWriter::write(Frame& input) {
             resampler_.renew_buffers(window_[0], window_[1], window_[2]);
 
             Frame out_frame(output_.data(), output_.size());
+
             while (resampler_.resample_buff(out_frame)) {
                 writer_.write(out_frame);
+
+                if (poison_) {
```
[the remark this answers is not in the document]
gavv: We should do poisoning when the new frame begins, before writing any samples to it. This `if (poison_)` point is exactly that moment: we have done with the previous output frame and are starting a new one using the same buffer (maybe immediately or maybe on next write() call). The very first frame will not be poisoned thought, but I thinks it's not a big problem.

Alternatively, we could detect the new frame start before calling resample_buff(), but this would require adding new conditions to write() method. I wanted to avoid adding clever logic specific just for poisoning. I think the implementation of poisoning should be as stupid as possible because most bugs in such a feature would be completely undetectable.

src/modules/roc_pipeline/config.h
```diff
@@ -183,14 +187,18 @@ struct SenderConfig {
     //! Constrain receiver speed using a CPU timer according to the sample rate.
     bool timing;
 
+    //! Fill unitialized data with large values to make them more noticable.
```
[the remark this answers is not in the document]
gavv: Yes, it's the maximum. I'll change implementation to use audio::SampleMax.

src/tools/roc_recv/cmdline.ggo
```diff
@@ -56,7 +56,10 @@ section "Options"
     option "oneshot" 1 "Exit when last connected client disconnects"
         flag off
 
-    option "beep" - "Enable beep on packet loss" flag off
+    option "beeping" - "Enable beeping on packet loss" flag off
```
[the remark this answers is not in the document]
gavv: I agree that it should be a separate commit, but not sure it should be a new PR :)

src/modules/roc_pipeline/sender.cpp
gavv: I like this idea. Not sure about PoisonWriter, but PoisonReader might work. I'll try both.

src/modules/roc_audio/resampler_writer.cpp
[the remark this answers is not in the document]
dshil: Agree, if we don't poison the first frame it isn't a big deal but it seems odd.
We introduce some kind of contract that `EVERY` frame created by various
components will be poisoned. Now you want to make some components poison
every new frame while another components may skip the first frame and poison
all the rest.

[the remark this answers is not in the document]
gavv: > We introduce some kind of contract that EVERY frame created by various components will be poisoned.

Well, we didn't state that anywhere :) What we did state is that this option poisons something somewhere which may help to detect uninitialized samples.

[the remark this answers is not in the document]
dshil: Ok, I got it. Without `PoisonWriter` it wasn't clear for me, don't know why :)

src/modules/roc_pipeline/sender.cpp
dshil: Could we pass `ReceiverConfig` directly to the `ReceiverSession`. It allows to reduce the number of parameters that `ReceiverSession` constructor takes.

gavv: I've extracted global receiver output parameters into ReceiverOutputConfig and now we pass it to the ReceiverSession. Passing the whole ReceiverConfig would be a bit confusing because it contains default session config. This default config is currently used for all sessions unconditionally, but in this will change when we implement some kind of session negotiation like SDP.

src/modules/roc_pipeline/config.h
```diff
@@ -107,55 +107,61 @@ struct SessionConfig {
...
     packet::channel_mask_t channels;
 
+    //! Perform resampling to to compensate sender and receiver frequency difference.
+    bool resampling;
+
     //! Constrain receiver speed using a CPU timer according to the sample rate.
     bool timing;
 
-    ReceiverConfig()
+    //! Fill unitialized data with large values to make them more noticable.
```
dshil: s/unitialized/uninitialized
s/noticable/noticeable

gavv: Fixed.

src/modules/roc_pipeline/config.h
```diff
@@ -107,55 +107,61 @@ struct SessionConfig {
...
+//! @remarks
+//!  Defines common output parameters on the receiver side.
+struct ReceiverOutputConfig {
     //! Number of samples per second per channel.
     size_t sample_rate;
 
     //! Channel mask.
     packet::channel_mask_t channels;
 
+    //! Perform resampling to to compensate sender and receiver frequency difference.
```
dshil: s/to to/to

src/tools/roc_conv/cmdline.ggo
```diff
@@ -25,3 +25,6 @@ section "Options"
 
     option "resampler-frame" - "Number of samples per resampler frame"
         int optional
+
+    option "poisoning" - "Enable poisoning to detect unitialized samples"
```
dshil: s/unitialized/uninitialized

src/tools/roc_recv/cmdline.ggo
```diff
@@ -56,7 +56,10 @@ section "Options"
...
+
+    option "poisoning" - "Enable poisoning to detect unitialized samples"
```
dshil: s/unitialized/uninitialized

src/tools/roc_send/cmdline.ggo
```diff
@@ -38,6 +38,9 @@ section "Options"
 
     option "interleaving" - "Enable packet interleaving" flag off
 
+    option "poisoning" - "Enable poisoning to detect unitialized samples"
```
dshil: s/unitialized/uninitialized

dshil:
LGTM, merged!

Issue #90, opened 2017-07-13
Memory leak when roc-send exits


gavv:
When roc-send finishes sending file and exits, LeakSanitizer aborts the program.

```
==8693==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 31 byte(s) in 1 object(s) allocated from:
    #0 0x4e4558 in __interceptor_malloc (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x4e4558)
    #1 0x54bb54 in gengetopt_strdup (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x54bb54)
    #2 0x54b5ea in update_arg (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x54b5ea)
    #3 0x545a42 in cmdline_parser_internal (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x545a42)
    #4 0x54491f in cmdline_parser2 (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x54491f)
    #5 0x54447f in cmdline_parser (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x54447f)
    #6 0x520f97 in main /home/victor/dev/roc-project/roc.develop/src/tools/roc_send/main.cpp:47:22
    #7 0x7fdd28e466bf in __libc_start_main (/lib64/libc.so.6+0x206bf)

Direct leak of 31 byte(s) in 1 object(s) allocated from:
    #0 0x4e4558 in __interceptor_malloc (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x4e4558)
    #1 0x54bb54 in gengetopt_strdup (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x54bb54)
    #2 0x54b9e1 in update_arg (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x54b9e1)
    #3 0x545a42 in cmdline_parser_internal (/home/victor/dev/roc-project/roc.develop/bin/x86_64-pc-linux-gnu/roc-send+0x545a42)
[...]

PR #159, opened 2019-01-27
tools: avoid memory leaks

dshil:
According to [1] we should release resources allocated during the
parsing process.

1. https://www.gnu.org/software/gengetopt/gengetopt.html

Relates: #90

### DIFF
--- a/src/modules/roc_core/target_posix/roc_core/scoped_destructor.h
+++ b/src/modules/roc_core/target_posix/roc_core/scoped_destructor.h
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_core/target_posix/roc_core/scoped_destructor.h
+//! @brief Scoped destructor.
+
+#ifndef ROC_CORE_SCOPED_DESTRUCTOR_H_
+#define ROC_CORE_SCOPED_DESTRUCTOR_H_
+
+#include "roc_core/stddefs.h"
+#include "roc_core/noncopyable.h"
+
+namespace roc {
+namespace core {
+
+//! Destroys the object via custom deleter.
+template <class T, void (*Func)(T)>
+class ScopedDestructor : public core::NonCopyable<> {
+public:
+    //! Initialize.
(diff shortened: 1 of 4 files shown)


src/modules/roc_core/target_posix/roc_core/cmdline_parser.h
```diff
@@ -0,0 +1,79 @@
...
+//! @tparam ParseFunc represents a function to parse command line arguments.
+//! @tparam DestroyFunc represents a function to release any resources
+//! allocated during the parsing.
+template
+    <
+      class T
+    , class ParseFunc = int(*)(int, char**, T*)
+    , class DestroyFunc = void(*)(T*)
+    >
+class CmdlineParser : public core::NonCopyable<> {
```
gavv: The class looks odd. This "parser" does not parse anything. It actually implements RAII on top of the two generic functions that are not specific to parsing.

I suggest to add a simpler helper (to roc_core) which solves the problem directly:

```
template <class T, void (*Func)(T)> class ScopedDestructor : public NonCopyable<> {
public:
    ScopedDestructor(T obj)
        : obj_(obj) {
    }

    ~ScopedDestructor() {
        Func(obj_);
    }

private:
    T obj_;
};
```

And then use it like this:

```
    gengetopt_args_info args;

    const int code = cmdline_parser(argc, argv, &args);
    if (code != 0) {
        return code;
    }

    core::ScopedDestructor<gengetopt_args_info*, cmdline_parser_free> args_destructor(
        &args);
```

You could also reuse core::UniquePtr by providing it a custom Destroyer class, but this would require too much boilerplate code, I think.

dshil: Indeed, it is. My first solution was similar to your and I also tried to
provide the Destoyer to the core::UniquePtr. I still don't like both approaches
because you separate the parse operation and wrapping object into RAII, which
ultimetely does not stop from just forgetting to wrap the parsed object.

Well, as your solution is more simple, we can use it.

gavv: > you separate the parse operation and wrapping object into RAII

This is not good, but IMHO it's not critical at all; we have just a couple of command line tools, not dozens of them :)

dshil: Agree, just wanted to clarify why this "parser" appeared.

gavv:
Merged, but moved the new file from roc_core/target_posix to roc_core.

8015e13c455e274b530688493b574ee079d2f4dd

Issue #48, opened 2016-01-06
Replace ad-hoc FEC packets with something standard-compliant


gavv:
See also #6 and #13.

I've found two alternative groups of RFCs:
1. **FEC Framework (FECFRAME)**
   - [RFC 6363: FEC Framework](https://tools.ietf.org/html/rfc6363)
     
     Supports RTP and externally defined FEC schemes.
   - [RFC 6816: LDPC FEC Scheme for FEC Framework](https://tools.ietf.org/html/rfc6816)
     
     LDPC-Staircase from OpenFEC [implements](http://openfec.org/publication_activities.html) this scheme.
   - [RFC 5170: LDPC FEC](https://tools.ietf.org/html/rfc5170)
     
     RFC 6816 inherits [Section 6: FEC Scheme Specification](https://tools.ietf.org/html/rfc5170#section-6) and [Section 5.7: PRNG](https://tools.ietf.org/html/rfc6816#section-5.7) from this RFC.
   - [RFC 6364: SDP support for FEC Framework](https://tools.ietf.org/html/rfc6364)
   
   There is also Raptor(Q) support for FEC Framework:
   - [RFC 6681: Raptor FEC Scheme for FEC Framework](https://tools.ietf.org/html/rfc6681)
   - [RFC 6682: RTP payload format for Raptor FEC Scheme for FEC Framework](https://tools.ietf.org/html/rfc6682)
2. **Generic FEC**
   - [RFC 5109: Generic Forward Error Correction for RTP](https://tools.ietf.org/html/rfc5109)
   - [RFC 2198: Redundant Audio Data](https://tools.ietf.org/html/rfc2198) (one of backends for Generic FEC)
   - [RFC 6354: Forward-Shifted Redundancy](https://tools.ietf.org/html/rfc6354) (optional extension)

Finally, these RFCs are mentioned by both FEC Framework and Generic FEC:
- [RFC 5725: Post-Repair Loss Reports](https://tools.ietf.org/html/rfc5725)
- [RFC 4585: RTP/AVPF, extended profile for RTCP Feedback](https://tools.ietf.org/html/rfc4585)
- [RFC 4588: Retransmission](https://tools.ietf.org/html/rfc4588) (uses AVPF)

gavv:
[FECFRAME working group](http://www.tech-invite.com/xtsv/tinv-ietf-wg-fecframe.html)

gavv:
I think we should start with FECFRAME support.

[...]

PR #160, opened 2019-02-16
roc_fec: handle source block number properly

dshil:
#48

Doubts:

* I still need to use seqnum in the fec reader to ensure that packets via bad
  seqnum wan't be recovered.
* I do not want to use seqnum_t as a type for the source block number. It will
  be more reasonable to define a new type for it. I am not sure where to put
  this stuff. I only found roc_packet/units.h. Should we also define types for
  other fec-specific packet fields: source_block_length?

### DIFF
--- a/src/modules/roc_fec/composer.h
+++ b/src/modules/roc_fec/composer.h
@@ -159,13 +159,13 @@ public:
         PayloadID& payload_id = *(PayloadID*)fec.payload_id.data();
 
         payload_id.clear();
-        payload_id.set_sbn(fec.blknum);
+        payload_id.set_sbn(fec.source_block_number);
 
         roc_panic_if(fec.source_block_length >= (uint16_t)-1);
         payload_id.set_k((uint16_t)fec.source_block_length);
 
-        roc_panic_if(fec.repair_symbol_id >= (uint16_t)-1);
-        payload_id.set_esi((uint16_t)fec.repair_symbol_id);
+        roc_panic_if(fec.encoding_symbol_id >= (uint16_t)-1);
+        payload_id.set_esi((uint16_t)fec.encoding_symbol_id);
 
         if (inner_composer_) {
             return inner_composer_->compose(packet);
--- a/src/modules/roc_fec/reader.cpp
+++ b/src/modules/roc_fec/reader.cpp
@@ -34,7 +34,7 @@ Reader::Reader(const Config& config,
     , started_(false)
     , can_repair_(false)
     , next_packet_(0)
-    , cur_block_sn_(0)
+    , cur_sbn_(0)
     , has_source_(false)
     , source_(0)
     , n_packets_(0) {
@@ -82,18 +82,19 @@ packet::PacketPtr Reader::read_() {
                 source_ = pp->rtp()->source;
                 has_source_ = true;
             }
-            cur_block_sn_ = pp->rtp()->seqnum;
-            skip_repair_packets_();
+            cur_sbn_ = pp->fec()->source_block_number;
+            drop_repair_packets_from_prev_blocks_();
         }
 
-        if (!pp || pp->fec()->repair_symbol_id > 0) {
+        if (!pp || pp->fec()->encoding_symbol_id > 0) {
             return source_queue_.read();
         }
 
-        roc_log(LogDebug,
-                "fec reader: got first packet in a block, start decoding:"
-                " n_packets_before=%u blk_sn=%lu",
-                n_packets_, (unsigned long)cur_block_sn_);
+        roc_log(LogDebug, "fec reader: got first packet in a block, start decoding:"
+                          " n_packets_before=%u sn=%lu sbn=%lu",
+                          n_packets_,
+                          (unsigned long)pp->rtp()->seqnum,
+                          (unsigned long)cur_sbn_);
 
         started_ = true;
     }
@@ -139,7 +140,7 @@ packet::PacketPtr Reader::get_next_packet_() {
 }
 
 void Reader::next_block_() {
-    roc_log(LogTrace, "fec reader: next block: sn=%lu", (unsigned long)cur_block_sn_);
+    roc_log(LogTrace, "fec reader: next block: sbn=%lu", (unsigned long)cur_sbn_);
 
     for (size_t n = 0; n < source_block_.size(); n++) {
         source_block_[n] = NULL;
@@ -149,7 +150,7 @@ void Reader::next_block_() {
         repair_block_[n] = NULL;
     }
 
-    cur_block_sn_ += source_block_.size();
+    cur_sbn_++;
     next_packet_ = 0;
 
     can_repair_ = false;
@@ -198,7 +199,7 @@ void Reader::try_repair_() {
 
         pp->set_data(buffer);
 
-        if (!check_packet_(pp, n)) {
+        if (!check_packet_(pp)) {
             roc_log(LogDebug, "fec reader: dropping unexpected repaired packet");
             continue;
         }
@@ -210,11 +211,11 @@ void Reader::try_repair_() {
     can_repair_ = false;
 }
 
-bool Reader::check_packet_(const packet::PacketPtr& pp, size_t pos) {
+bool Reader::check_packet_(const packet::PacketPtr& pp) {
     roc_panic_if_not(has_source_);
 
     if (!pp->rtp()) {
-        roc_log(LogDebug, "fec reader: repaired unexcpeted non-rtp packet");
+        roc_log(LogDebug, "fec reader: repaired unexpected non-rtp packet");
         return false;
     }
 
@@ -224,14 +225,6 @@ bool Reader::check_packet_(const packet::PacketPtr& pp, size_t pos) {
                 " got=%lu expected=%lu",
                 (unsigned long)pp->rtp()->source, (unsigned long)source_);
         return (alive_ = false);
-    }
-
-    if (pp->rtp()->seqnum != packet::seqnum_t(cur_block_sn_ + pos)) {
-        roc_log(LogDebug,
-                "fec reader: repaired packet has bad seqnum: got=%lu expected=%lu",
-                (unsigned long)pp->rtp()->seqnum,
-                (unsigned long)packet::seqnum_t(cur_block_sn_ + pos));
-        return false;
     }
 
     return true;
@@ -284,28 +277,25 @@ void Reader::update_source_packets_() {
             roc_panic("fec reader: unexpected non-rtp source packet");
         }
 
-        if (!packet::seqnum_lt(fec->blknum,
-                               packet::seqnum_t(cur_block_sn_ + source_block_.size()))) {
+        if (!packet::blknum_le(fec->source_block_number, cur_sbn_)) {
             break;
         }
 
         source_queue_.read();
         n_fetched++;
 
-        if (packet::seqnum_lt(fec->blknum, cur_block_sn_)) {
-            roc_log(LogTrace,
-                    "fec reader: dropping source packet from previous block:"
-                    " blk_sn=%lu pkt_sn=%lu",
-                    (unsigned long)cur_block_sn_, (unsigned long)rtp->seqnum);
+        if (packet::blknum_lt(fec->source_block_number, cur_sbn_)) {
+            roc_log(LogTrace, "fec reader: dropping source packet from previous block:"
+                              " cur_sbn=%lu pkt_sbn=%lu pkt_sn=%lu",
+                              (unsigned long)cur_sbn_,
+                              (unsigned long)fec->source_block_number,
+                              (unsigned long)rtp->seqnum);
             n_dropped++;
             continue;
         }
 
-        roc_panic_if(fec->blknum != cur_block_sn_);
-        const size_t p_num = fec->repair_symbol_id;
-
-        roc_panic_if((packet::seqnum_diff_t)p_num
-                     != packet::seqnum_diff(rtp->seqnum, cur_block_sn_));
+        roc_panic_if(fec->source_block_number != cur_sbn_);
+        const size_t p_num = fec->encoding_symbol_id;
 
         if (!source_block_[p_num]) {
             can_repair_ = true;
@@ -334,25 +324,24 @@ void Reader::update_repair_packets_() {
             roc_panic("fec reader: unexpected non-fec repair packet");
         }
 
-        if (!packet::seqnum_lt(fec->blknum,
-                               packet::seqnum_t(cur_block_sn_ + source_block_.size()))) {
+        if (!packet::blknum_le(fec->source_block_number, cur_sbn_)) {
             break;
         }
 
         repair_queue_.read();
         n_fetched++;
 
-        if (packet::seqnum_lt(fec->blknum, cur_block_sn_)) {
-            roc_log(LogTrace,
-                    "fec reader: dropping repair packet from previous block:"
-                    " blk_sn=%lu pkt_data_blk=%lu",
-                    (unsigned long)cur_block_sn_, (unsigned long)fec->blknum);
+        if (packet::blknum_lt(fec->source_block_number, cur_sbn_)) {
+            roc_log(LogTrace, "fec reader: dropping repair packet from previous block:"
+                              " cur_sbn=%lu pkt_sbn=%lu",
+                              (unsigned long)cur_sbn_,
+                              (unsigned long)fec->source_block_number);
             n_dropped++;
             continue;
         }
 
-        roc_panic_if(fec->repair_symbol_id < fec->source_block_length);
-        const size_t p_num = fec->repair_symbol_id - fec->source_block_length;
+        roc_panic_if(fec->encoding_symbol_id < fec->source_block_length);
+        const size_t p_num = fec->encoding_symbol_id - fec->source_block_length;
         roc_panic_if(p_num >= repair_block_.size());
 
         if (!repair_block_[p_num]) {
@@ -368,7 +357,7 @@ void Reader::update_repair_packets_() {
     }
 }
 
-void Reader::skip_repair_packets_() {
+void Reader::drop_repair_packets_from_prev_blocks_() {
     unsigned n_dropped = 0;
 
     for (;;) {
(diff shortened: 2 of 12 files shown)


src/modules/roc_packet/fec.h
```diff
@@ -21,8 +21,8 @@ namespace packet {
 
 //! FECFRAME packet.
 struct FEC {
-    //! Seqnum of first source packet in block.
-    seqnum_t blknum;
+    //! The source block number.
+    seqnum_t source_block_number;
```
gavv: > I do not want to use seqnum_t as a type for the source block number. It will
  be more reasonable to define a new type for it. I am not sure where to put
  this stuff. I only found roc_packet/units.h. Should we also define types for
  other fec-specific packet fields: source_block_length?

We need seqnum_t to handle the overlap correctly. Its size should be the same as in the RTP packet.

Block number may overlap too and we need to handle its overlap as well. So I guess we need blknum_t. I think we need the following:

**units.h**

```
typedef uint32_t blknum_t;
typedef int32_t signed_blknum_t;
```

**fec.h**

```
struct FEC {
    //! The source block number.
    blknum_t source_block_number;

    //! The number of source packet per block.
    blknum_t source_block_length;

    ...
```

src/modules/roc_fec/composer.h
```diff
@@ -160,7 +160,7 @@ class Composer : public packet::IComposer, public core::NonCopyable<> {
         PayloadID& payload_id = *(PayloadID*)fec.payload_id.data();
 
         payload_id.clear();
-        payload_id.set_sbn(fec.blknum);
+        payload_id.set_sbn(fec.source_block_number);
```
gavv: I would add a check:

```
        roc_panic_if(fec.source_block_number >= (uint32_t)-1);
        payload_id.set_sbn((uint32_t)fec.source_block_number);
```

Or maybe:

```
        roc_panic_if(sizeof(blknum_t) != sizeof(uint32_t));
        payload_id.set_sbn(fec.source_block_number);
```

To ensure that we don't lose bits here.

src/modules/roc_fec/reader.cpp
```diff
@@ -95,17 +100,20 @@ packet::PacketPtr Reader::read_() {
...
         }
 
         if (!pp || pp->fec()->repair_symbol_id > 0) {
             return source_queue_.read();
         }
 
         roc_log(LogDebug, "fec reader: got first packet in a block, start decoding:"
-                          " n_packets_before=%u blk_sn=%lu",
-                n_packets_, (unsigned long)cur_block_sn_);
+                          " n_packets_before=%u block_n=%lu sn=%lu",
```
gavv: I suggest to use "sbn" abbreviation when we refer to source block number in logs.

src/modules/roc_packet/fec.h
gavv: To handle overlaps correctly, the length of the blknum_t should be smaller or equal to the length of the SBN in Payload ID header.

Reed-solomon uses 32-bit SBNs, but LDPC uses 16-bit SBNs, so we need 16-bit blknum.

src/modules/roc_fec/reader.cpp
```diff
@@ -226,7 +236,7 @@ bool Reader::check_packet_(const packet::PacketPtr& pp, size_t pos) {
     roc_panic_if_not(has_source_);
 
     if (!pp->rtp()) {
-        roc_log(LogDebug, "fec reader: repaired unexcpeted non-rtp packet");
+        roc_log(LogDebug, "fec reader: repaired unexpected non-rtp packet");
         return false;
     }
```
gavv: > I still need to use seqnum in the fec reader to ensure that packets via bad
  seqnum wan't be recovered.

```
    if (pp->rtp()->seqnum != packet::seqnum_t(cur_block_sn_ + pos)) {
        roc_log(LogDebug,
                "fec reader: repaired packet has bad seqnum: got=%lu expected=%lu",
                (unsigned long)pp->rtp()->seqnum,
                (unsigned long)packet::seqnum_t(cur_block_sn_ + pos));
        return false;
    }
```

This check is redundant. We can just remove it to avoid a dependency on how the sender fills seqnums.

src/modules/roc_fec/writer.cpp
```diff
@@ -14,6 +14,14 @@
...
 
+namespace {
+
+packet::seqnum_t get_random_seqnum() {
+    return (packet::seqnum_t)core::random(packet::seqnum_t(-1));
+}
+
+} // namespace
+
 Writer::Writer(const Config& config,
```
gavv: Could you add a test that checks how the Writer fills SNs and SBNs on produced packets?

src/modules/roc_fec/reader.cpp
dshil: Done

src/modules/roc_fec/reader.cpp
dshil: Done

src/modules/roc_packet/fec.h
dshil: Done

src/modules/roc_fec/writer.cpp
dshil: Done

src/modules/roc_fec/reader.cpp
```diff
@@ -284,29 +275,25 @@ void Reader::update_source_packets_() {
             roc_panic("fec reader: unexpected non-rtp source packet");
         }
 
-        if (!packet::seqnum_lt(fec->blknum,
-                               packet::seqnum_t(cur_block_sn_ + source_block_.size()))) {
+        if (!packet::seqnum_le(fec->source_block_number, cur_sbn_)) {
```
gavv: I think we need a separate set of helpers (blknum_diff_t, blknum_diff, blknum_le, blknum_lt) for the blknum_t. We should not rely on the fact that blknum_t and seqnum_t are the same type currently.

src/modules/roc_fec/writer.cpp
```diff
@@ -147,24 +148,28 @@ packet::PacketPtr Writer::make_repair_packet_(packet::seqnum_t n) {
...
-void Writer::fill_packet_fec_id_(const packet::PacketPtr& packet,
-                                 packet::seqnum_t pack_n) {
+void Writer::fill_packet_fec_fields_(const packet::PacketPtr& packet,
+                                     packet::seqnum_t pack_n) {
     packet::FEC& fec = *packet->fec();
 
-    fec.blknum = cur_block_source_sn_;
-    fec.source_block_length = n_source_packets_;
+    fec.source_block_number = cur_sbn_;
+    fec.source_block_length = (packet::blknum_t)n_source_packets_;
```
gavv: Hm, sorry, I misled you. I think the source_block_length should have size_t type, not blknum_t. It encodes the number of packets per block. This number does not wrap so it does not need a separate type to specify its size. The default type we use for various "number of objects" variables is size_t.

src/modules/roc_packet/fec.h
```diff
@@ -21,11 +21,11 @@ namespace packet {
 
 //! FECFRAME packet.
 struct FEC {
-    //! Seqnum of first source packet in block.
-    seqnum_t blknum;
+    //! The source block number.
+    blknum_t source_block_number;
```
gavv: It's worth to emphasize in the comment that:

* this is the number of a block, not a packet
* blocks are numbered sequentially starting from a random number
* block number can wrap

src/modules/roc_packet/fec.h
```diff
@@ -21,11 +21,11 @@ namespace packet {
...
 
-    //! The number of source packet per block.
-    size_t source_block_length;
+    //! The number of source packets per block.
+    blknum_t source_block_length;
```
gavv: I'd clarify in the comment that it is the number of source packets in the block to which this packets belongs to. It's important since different blocks can have different sizes.

src/modules/roc_packet/target_stdio/roc_packet/print.cpp
```diff
@@ -35,7 +35,8 @@ void print(const Packet& p, int flags) {
     }
 
     if (p.fec()) {
-        fprintf(stderr, "fec: blk=%lu payload=%lu\n", (unsigned long)p.fec()->blknum,
+        fprintf(stderr, "fec: sbn=%lu payload=%lu\n",
+                (unsigned long)p.fec()->source_block_number,
```
gavv: Could you please also add source block length?

src/modules/roc_packet/units.h
```diff
@@ -87,6 +87,9 @@ static inline size_t num_channels(channel_mask_t ch_mask) {
     return n_ch;
 }
 
+//! Packet block number in packet stream.
+typedef uint16_t blknum_t;
```
gavv: It's worth to mention here that we're talking about FEC blocks.

src/modules/roc_packet/fec.h
dshil: Done

src/modules/roc_packet/fec.h
dshil: Done

src/modules/roc_packet/target_stdio/roc_packet/print.cpp
dshil: Done

src/modules/roc_packet/units.h
dshil: Done

src/modules/roc_fec/writer.cpp
dshil: Done

src/modules/roc_fec/reader.cpp
dshil: Done

src/modules/roc_packet/target_stdio/roc_packet/print.cpp
gavv: "sbn_len" -> "sb_len", or maybe "sbl" or "sblen". I suggest "sblen".

src/modules/roc_fec/reader.cpp
```diff
@@ -284,29 +275,26 @@ void Reader::update_source_packets_() {
...
         n_fetched++;
 
-        if (packet::seqnum_lt(fec->blknum, cur_block_sn_)) {
-            roc_log(LogTrace,
-                    "fec reader: dropping source packet from previous block:"
-                    " blk_sn=%lu pkt_sn=%lu",
-                    (unsigned long)cur_block_sn_, (unsigned long)rtp->seqnum);
+        if (packet::blknum_lt(fec->source_block_number, cur_sbn_)) {
+            roc_log(LogTrace, "fec reader: dropping source packet from previous block:"
+                              " cur_sbn=%lu recv_sbn=%lu pkt_sn=%lu",
```
gavv: "recv_sbn" -> "pkt_sbn" for consistency

src/modules/roc_fec/reader.cpp
```diff
@@ -334,19 +322,18 @@ void Reader::update_repair_packets_() {
...
         n_fetched++;
 
-        if (packet::seqnum_lt(fec->blknum, cur_block_sn_)) {
-            roc_log(LogTrace,
-                    "fec reader: dropping repair packet from previous block:"
-                    " blk_sn=%lu pkt_data_blk=%lu",
-                    (unsigned long)cur_block_sn_, (unsigned long)fec->blknum);
+        if (packet::blknum_lt(fec->source_block_number, cur_sbn_)) {
+            roc_log(LogTrace, "fec reader: dropping repair packet from previous block:"
+                              " cur_sbn=%lu recv_sbn=%lu",
```
gavv: "recv_sbn" -> "pkt_sbn" for consistency

src/modules/roc_fec/reader.cpp
```diff
@@ -82,18 +82,17 @@ packet::PacketPtr Reader::read_() {
                 source_ = pp->rtp()->source;
                 has_source_ = true;
             }
-            cur_block_sn_ = pp->rtp()->seqnum;
-            skip_repair_packets_();
+            cur_sbn_ = pp->fec()->source_block_number;
+            drop_outdated_repair_packets_();
```
gavv: Renaming skip_repair_packets_ to drop_outdated_repair_packets_ is not quite correct. 

Here we handle the (typical) situation when we are starting in the middle of a block. In this case we wait for the first source packet and then drop all repair packet from previous blocks. So originally these packets were not outdated, but we make them outdated by waiting for the source packet from the next block.

I would name this kind of drop somehow else to avoid confusion with really outdated packets. Maybe "dropping preceding repair packets".

src/modules/roc_fec/reader.cpp
```diff
@@ -226,14 +225,6 @@ bool Reader::check_packet_(const packet::PacketPtr& pp, size_t pos) {
         return (alive_ = false);
     }
 
-    if (pp->rtp()->seqnum != packet::seqnum_t(cur_block_sn_ + pos)) {
```
gavv: We can compensate removal of this check by adding a new check to rtp::Validator. There we can check that sequence numbers always increase and drop packets violating this rule. There is a validator in the pipeline after fec reader.

src/modules/roc_packet/units.h
```diff
@@ -87,6 +87,27 @@ static inline size_t num_channels(channel_mask_t ch_mask) {
...
 }
 
+//! FEC block number in a packet stream.
+typedef uint16_t blknum_t;
+
+//! FEC block numbers difference.
+typedef int16_t blknum_diff_t;
+
+//! Compute difference between two FEC block numbers.
+inline blknum_diff_t blknum_diff(blknum_t a, blknum_t b) {
```
gavv: Please add tests for blknum helpers to roc_packet/test_units.cpp

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -797,5 +755,186 @@ TEST(writer_reader, delayed_packets) {
     }
 }
 
+TEST(writer_reader, fill_sn_sbn) {
```
gavv: We have a test named "encode_source_id_and_seqnum" so I suggest to rename this one to "encode_sbn" and move it closer to that test.

src/modules/roc_packet/fec.h
```diff
@@ -21,10 +21,18 @@ namespace packet {
...
+    blknum_t source_block_number;
 
-    //! The number of source packet per block.
+    //! Number of source packets in the block to which this packet belongs to.
+    //!
+    //! @remarks
+    //!  Different blocks can have different number of source packets.
     size_t source_block_length;
 
     //! The index number of the repair packet in block.
```
gavv: I think we also should rename "repair_symbol_id" to "encoding_symbol_id", as it is called in the RFC.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -797,5 +755,186 @@ TEST(writer_reader, delayed_packets) {
...
+    CHECK(first_packet);
+    CHECK(first_packet->fec()->repair_symbol_id == 0);
+
+    const packet::blknum_t sbn = first_packet->fec()->source_block_number;
+
+    for (size_t n = 1; n < NumSourcePackets; ++n) {
+        packet::PacketPtr p = reader.read();
+        CHECK(p);
+        CHECK(p->fec()->source_block_number == sbn);
+        CHECK(p->fec()->repair_symbol_id == n);
```
gavv: It would be great to add two more checks to this test or maybe to a separate test:

* check source_block_length
* check source_block_number and repair_symbol_id of repair packets as well (currently we check only source packets)

src/modules/roc_fec/reader.cpp
dshil: Well, I agree that they are not outdated. Probably it will be better to name it:

* skip_prev_block()
* skip_packets_from_prev_block()
* skip_queued_packets()

I would prefer to use `skip_queued_packets`. Probably it is also not a good name, but it
tells the code reader the truth.

src/modules/roc_fec/reader.cpp
dshil: Should timestamps also increase?

dshil: Should timestamps also increase?

dshil: I also want to expand the log msgs produced by validator to:

```
[debug] roc_rtp: rtp validator: too long seqnum jump: prev=150 next=251 dist=101 limit=100
```

As we see that we have `too long jump` but it will be very convenient to see what the maximum allowed number.

src/modules/roc_fec/reader.cpp
dshil: Done

src/modules/roc_fec/reader.cpp
dshil: Done

src/modules/roc_packet/target_stdio/roc_packet/print.cpp
dshil: Done

src/tests/roc_fec/target_openfec/test_writer_reader.cpp

src/modules/roc_packet/units.h
dshil: I was very creative and just stole your tests for seqnum_t .

src/modules/roc_fec/reader.cpp
gavv: Timestamps are not required to be monotonic, IIRC. Depacketizer handles this.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
dshil: I added a separate test.

gavv: I'm looking at `encode_sbn` and `encode_sbn_validate_repair_packets` tests and I think it's not quite correct to use reader there. We want to test that the *writer* correctly fills sbn, sbl, and esi, right? However in current tests we can't be sure that these fields were filled by writer, not by reader. Note that `encode_source_id_and_seqnum` test doesn't use reader. I suggest to remove reader usage from these tests and also I think we can merge them because they test the same feature and look very similarly.

On the other hand, `encode_sbn_outdated_block` and `encode_sbn_repaired_block` actually test writer + reader cooperation, and not the encoding process itself. I think we should rename them to something like `drop_outdated_block` and `repaired_block_numbering`, respectively, and move back to the end of the file.

src/modules/roc_fec/reader.cpp
gavv: Well, the logic is good, but the naming is not quite true because this method does not simply drop all queued packets unconditionally. A correct name would be something like "drop_repair_packets_from_prev_blocks_()`.

We should also fix the log message:

```
        roc_log(LogTrace, "fec reader: dropping outdated repair packet,"
                          " decoding not started: cur_sbn=%lu pkt_sbn=%lu",
                          (unsigned long)cur_sbn_,
                          (unsigned long)fec->source_block_number);
```

src/modules/roc_fec/reader.cpp
dshil: As discussed we won't handle this case neither in validator nor in reader. We can resolve this discussion.

src/modules/roc_fec/reader.cpp
dshil: Fixed

src/modules/roc_packet/fec.h
dshil: Done

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
dshil: Fixed

PR #174, opened 2019-04-05
FEC: handle source block length change

dshil:
#48

### DIFF
--- a/src/modules/roc_fec/idecoder.h
+++ b/src/modules/roc_fec/idecoder.h
@@ -23,14 +23,25 @@ class IDecoder {
 public:
     virtual ~IDecoder();
 
+    //! Start block.
+    //!
+    //! @remarks
+    //!  Performs an initial setup for a block. Should be called before
+    //!  any operations for the block.
+    virtual bool begin(size_t sblen, size_t rblen) = 0;
+
     //! Store source or repair packet buffer for current block.
     virtual void set(size_t index, const core::Slice<uint8_t>& buffer) = 0;
 
     //! Repair source packet buffer.
     virtual core::Slice<uint8_t> repair(size_t index) = 0;
 
-    //! Reset current block.
-    virtual void reset() = 0;
+    //! Finish block.
+    //!
+    //! @remarks
+    //!  Cleanups the resources allocated for the block. Should be called after
+    //!  all operations for the block.
+    virtual void end() = 0;
 };
 
 } // namespace fec
--- a/src/modules/roc_fec/reader.cpp
+++ b/src/modules/roc_fec/reader.cpp
@@ -38,9 +38,6 @@ Reader::Reader(const Config& config,
     , has_source_(false)
     , source_(0)
     , n_packets_(0) {
-    if (!source_block_.resize(config.n_source_packets)) {
-        return;
-    }
     if (!repair_block_.resize(config.n_repair_packets)) {
         return;
     }
@@ -76,30 +73,69 @@ packet::PacketPtr Reader::read_() {
     fetch_packets_();
 
     if (!started_) {
-        packet::PacketPtr pp = source_queue_.head();
-        if (pp) {
-            if (!has_source_) {
-                source_ = pp->rtp()->source;
-                has_source_ = true;
-            }
-            cur_sbn_ = pp->fec()->source_block_number;
-            drop_repair_packets_from_prev_blocks_();
-        }
-
+        packet::PacketPtr pp = read_first_packet_();
         if (!pp || pp->fec()->encoding_symbol_id > 0) {
             return source_queue_.read();
         }
 
-        roc_log(LogDebug, "fec reader: got first packet in a block, start decoding:"
-                          " n_packets_before=%u sn=%lu sbn=%lu",
-                          n_packets_,
-                          (unsigned long)pp->rtp()->seqnum,
-                          (unsigned long)cur_sbn_);
+        roc_log(LogDebug,
+                "fec reader: got first packet in a block, start decoding:"
+                " n_packets_before=%u sn=%lu sbn=%lu",
+                n_packets_, (unsigned long)pp->rtp()->seqnum, (unsigned long)cur_sbn_);
 
         started_ = true;
     }
 
     return get_next_packet_();
+}
+
+packet::PacketPtr Reader::read_first_packet_() {
+    packet::PacketPtr pp = source_queue_.head();
+    if (pp) {
+        if (!has_source_) {
+            source_ = pp->rtp()->source;
+            has_source_ = true;
+        }
+
+        cur_sbn_ = pp->fec()->source_block_number;
+        if (!update_size_(pp->fec()->source_block_length)) {
+            return NULL;
+        }
+
+        drop_repair_packets_from_prev_blocks_();
+    }
+
+    return pp;
+}
+
+bool Reader::update_size_(size_t new_sblen) {
+    const size_t cur_sblen = source_block_.size();
+    if (cur_sblen == new_sblen) {
+        return true;
+    }
+
+    if (next_packet_ != 0) {
+        roc_log(LogDebug,
+                "fec reader: can't update source block size in the middle"
+                " of the current block, shutting down:"
+                " esi=%lu cur_sbl=%lu new_sbl=%lu",
+                (unsigned long)next_packet_, (unsigned long)cur_sblen,
+                (unsigned long)new_sblen);
+        return (alive_ = false);
+    }
+
+    if (!source_block_.resize(new_sblen)) {
+        roc_log(LogDebug,
+                "fec reader: can't resize source block, shutting down:"
+                " cur_sbl=%lu new_sbl=%lu",
+                (unsigned long)cur_sblen, (unsigned long)new_sblen);
+        return (alive_ = false);
+    }
+
+    roc_log(LogDebug, "fec reader: update sblen, cur_sbl=%lu new_sbl=%lu",
+            (unsigned long)cur_sblen, (unsigned long)new_sblen);
+
+    return true;
 }
 
 packet::PacketPtr Reader::get_next_packet_() {
@@ -162,6 +198,15 @@ void Reader::try_repair_() {
         return;
     }
 
+    if (!decoder_.begin(source_block_.size(), repair_block_.size())) {
+        roc_log(LogDebug,
+                "fec reader: can't begin decoder block, shutting down:"
+                " sbl=%lu rbl=%lu",
+                (unsigned long)source_block_.size(), (unsigned long)repair_block_.size());
+        alive_ = false;
+        return;
+    }
+
     for (size_t n = 0; n < source_block_.size(); n++) {
         if (!source_block_[n]) {
             continue;
@@ -207,7 +252,7 @@ void Reader::try_repair_() {
         source_block_[n] = pp;
     }
 
-    decoder_.reset();
+    decoder_.end();
     can_repair_ = false;
 }
 
@@ -288,16 +333,20 @@ void Reader::update_source_packets_() {
         n_fetched++;
 
         if (packet::blknum_lt(fec->source_block_number, cur_sbn_)) {
-            roc_log(LogTrace, "fec reader: dropping source packet from previous block:"
-                              " cur_sbn=%lu pkt_sbn=%lu pkt_sn=%lu",
-                              (unsigned long)cur_sbn_,
-                              (unsigned long)fec->source_block_number,
-                              (unsigned long)rtp->seqnum);
+            roc_log(LogTrace,
+                    "fec reader: dropping source packet from previous block:"
+                    " cur_sbn=%lu pkt_sbn=%lu pkt_sn=%lu",
+                    (unsigned long)cur_sbn_, (unsigned long)fec->source_block_number,
+                    (unsigned long)rtp->seqnum);
             n_dropped++;
             continue;
         }
 
         roc_panic_if(fec->source_block_number != cur_sbn_);
+        if (!update_size_(fec->source_block_length)) {
+            return;
+        }
+
         const size_t p_num = fec->encoding_symbol_id;
 
         if (!source_block_[p_num]) {
@@ -335,10 +384,10 @@ void Reader::update_repair_packets_() {
         n_fetched++;
 
         if (packet::blknum_lt(fec->source_block_number, cur_sbn_)) {
-            roc_log(LogTrace, "fec reader: dropping repair packet from previous block:"
-                              " cur_sbn=%lu pkt_sbn=%lu",
-                              (unsigned long)cur_sbn_,
-                              (unsigned long)fec->source_block_number);
+            roc_log(LogTrace,
+                    "fec reader: dropping repair packet from previous block:"
+                    " cur_sbn=%lu pkt_sbn=%lu",
+                    (unsigned long)cur_sbn_, (unsigned long)fec->source_block_number);
             n_dropped++;
             continue;
         }
@@ -380,9 +429,8 @@ void Reader::drop_repair_packets_from_prev_blocks_() {
 
         roc_log(LogTrace,
                 "fec reader: dropping repair packet from previous blocks,"
-                          " decoding not started: cur_sbn=%lu pkt_sbn=%lu",
-                          (unsigned long)cur_sbn_,
-                          (unsigned long)fec->source_block_number);
+                " decoding not started: cur_sbn=%lu pkt_sbn=%lu",
+                (unsigned long)cur_sbn_, (unsigned long)fec->source_block_number);
 
         repair_queue_.read();
         n_dropped++;
--- a/src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
+++ b/src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
@@ -23,8 +23,8 @@ OFDecoder::OFDecoder(const Config& config,
                      size_t payload_size,
                      core::BufferPool<uint8_t>& buffer_pool,
                      core::IAllocator& allocator)
-    : blk_source_packets_(config.n_source_packets)
-    , blk_repair_packets_(config.n_repair_packets)
+    : sblen_(0)
+    , rblen_(0)
     , payload_size_(payload_size)
     , of_sess_(NULL)
     , of_sess_params_(NULL)
@@ -36,19 +36,6 @@ OFDecoder::OFDecoder(const Config& config,
     , has_new_packets_(false)
     , decoding_finished_(false)
     , valid_(false) {
-    if (!buff_tab_.resize(blk_source_packets_ + blk_repair_packets_)) {
-        return;
-    }
-    if (!data_tab_.resize(blk_source_packets_ + blk_repair_packets_)) {
-        return;
-    }
-    if (!recv_tab_.resize(blk_source_packets_ + blk_repair_packets_)) {
-        return;
-    }
-    if (!status_.resize(blk_source_packets_ + blk_repair_packets_ + 2)) {
-        return;
-    }
-
     if (config.codec == ReedSolomon8m) {
         roc_log(LogDebug, "of decoder: initializing Reed-Solomon decoder");
 
@@ -68,12 +55,8 @@ OFDecoder::OFDecoder(const Config& config,
         roc_panic("of decoder: invalid codec");
     }
 
-    of_sess_params_->nb_source_symbols = (uint32_t)blk_source_packets_;
-    of_sess_params_->nb_repair_symbols = (uint32_t)blk_repair_packets_;
     of_sess_params_->encoding_symbol_length = (uint32_t)payload_size_;
     of_verbosity = 0;
-
-    OFDecoder::reset(); // non-virtual call from ctor
 
     valid_ = true;
 }
@@ -88,13 +71,28 @@ bool OFDecoder::valid() const {
     return valid_;
 }
 
+bool OFDecoder::begin(size_t sblen, size_t rblen) {
+    roc_panic_if_not(valid());
+
+    if (!resize_tabs_(sblen + rblen)) {
+        return false;
+    }
+
+    sblen_ = sblen;
+    rblen_ = rblen;
+
+    update_session_params_(sblen, rblen);
+    reset_session_();
+
+    return true;
+}
+
 void OFDecoder::set(size_t index, const core::Slice<uint8_t>& buffer) {
     roc_panic_if_not(valid());
 
-    if (index >= blk_source_packets_ + blk_repair_packets_) {
+    if (index >= sblen_ + rblen_) {
         roc_panic("of decoder: index out of bounds: index=%lu, size=%lu",
-                  (unsigned long)index,
-                  (unsigned long)(blk_source_packets_ + blk_repair_packets_));
+                  (unsigned long)index, (unsigned long)(sblen_ + rblen_));
     }
 
     if (!buffer) {
@@ -134,22 +132,46 @@ core::Slice<uint8_t> OFDecoder::repair(size_t index) {
     return buff_tab_[index];
 }
 
-void OFDecoder::reset() {
+void OFDecoder::end() {
     if (of_sess_ != NULL) {
         report_();
         destroy_session_();
     }
 
-    reset_session_();
+    reset_tabs_();
 
     has_new_packets_ = false;
     decoding_finished_ = false;
+}
 
+void OFDecoder::update_session_params_(size_t sblen, size_t rblen) {
+    of_sess_params_->nb_source_symbols = (uint32_t)sblen;
+    of_sess_params_->nb_repair_symbols = (uint32_t)rblen;
+}
+
+void OFDecoder::reset_tabs_() {
     for (size_t i = 0; i < buff_tab_.size(); ++i) {
         buff_tab_[i] = core::Slice<uint8_t>();
         data_tab_[i] = NULL;
         recv_tab_[i] = false;
     }
+}
+
+bool OFDecoder::resize_tabs_(size_t size) {
+    if (!buff_tab_.resize(size)) {
+        return false;
+    }
+    if (!data_tab_.resize(size)) {
+        return false;
+    }
+    if (!recv_tab_.resize(size)) {
+        return false;
+    }
+    if (!status_.resize(size + 2)) {
+        return false;
+    }
+
+    return true;
 }
 
 void OFDecoder::update_() {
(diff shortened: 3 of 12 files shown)


src/modules/roc_fec/idecoder.h
```diff
@@ -23,14 +23,24 @@ class IDecoder {
...
+
     //! Store source or repair packet buffer for current block.
     virtual void set(size_t index, const core::Slice<uint8_t>& buffer) = 0;
 
     //! Repair source packet buffer.
     virtual core::Slice<uint8_t> repair(size_t index) = 0;
 
-    //! Reset current block.
-    virtual void reset() = 0;
+    //! Perform housekeeping for current block.
```
gavv: I'd make these comments:

> Perform an initial setup for current block:

> Perform housekeeping for current block.

a bit more explicit:

```
Start block.

@remarks
 Performs an initial setup for a block. Should be called before <...>
```

```
Finish block.

@remarks
 Cleanups the resources allocated for the block. Should be called <...>
```

gavv: Also note that this comment:

```
    //! Perform an initial setup for current block:
    //!  * Set number of source packets and repair packets.
```

does not fit well with the autobrief doxygen mode which we're using currently.

src/modules/roc_fec/reader.cpp
```diff
@@ -82,26 +79,43 @@ packet::PacketPtr Reader::read_() {
                 source_ = pp->rtp()->source;
                 has_source_ = true;
             }
+
             cur_sbn_ = pp->fec()->source_block_number;
+            resize_(pp->fec()->source_block_length);
```
gavv: I suggest to extract the body of the `if (!started_) {` block into a method.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -902,5 +902,122 @@ TEST(writer_reader, repaired_block_numbering) {
     }
 }
 
+TEST(writer_reader, resize_souce_block_increase_decrease) {
```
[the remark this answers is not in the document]
gavv: There is also another concern with this test. If both writer and reader are broken, and the writer will correctly fill the SBL field but will not really change the block size, and the reader will just ignore the SBL field, this test will nevertheless happily succeed. It would be great if our test suite would also check that block sizes are *really* changing.

One way to do it is to add a test checking that the writer produces correct output when we're changing the block size. We set it to `<10, 5>` and it produces 10 source then 5 repair packets. We set it to `<15, 5>` and it produces 15 source then 5 repair packets.

In that way we would have 1) a test checking that writer produces an expected output 2) a test checking that writer + reader work as expected together (resize_souce_block_increase_decrease). This pair of tests would be enough to be sure that writer and reader work properly both individually and in pair.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
```diff
@@ -88,13 +71,35 @@ bool OFDecoder::valid() const {
     return valid_;
 }
 
+bool OFDecoder::begin(size_t sblen, size_t rblen) {
+    roc_panic_if_not(valid());
+
+    if (sblen_ == sblen && rblen_ == rblen) {
+        return true;
```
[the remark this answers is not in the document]
dshil: No. In most cases when we do not change a block size the workflow will be:

```
decoder.begin() 
   -> decoder.set() 
   -> decoder.end()  // here we will reset a session
```

Most of the time we will call `begin` with the same parameters and this `if` is some kind of optimization to avoid recreating FEC session params each time.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
```diff
@@ -88,13 +71,35 @@ bool OFDecoder::valid() const {
...
+
+    if (!resize_tabs_(sblen + rblen)) {
+        return false;
+    }
+
+    sblen_ = sblen;
+    rblen_ = rblen;
+
+    of_sess_params_->nb_source_symbols = (uint32_t)sblen_;
+    of_sess_params_->nb_repair_symbols = (uint32_t)rblen_;
```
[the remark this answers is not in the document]
dshil: > Maybe not in this PR though.

I want to do error handling separately because there are more things to do.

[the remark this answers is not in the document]
dshil: > I'd also extract it to a method like

Agree

src/modules/roc_fec/writer.cpp
```diff
@@ -68,51 +69,82 @@ void Writer::write(const packet::PacketPtr& pp) {
...
+void Writer::resize(size_t sblen) {
+    roc_log(LogDebug, "fec writer: update sblen, cur=%lu next=%lu",
+            (unsigned long)cur_sblen_, (unsigned long)sblen);
+
+    next_sblen_ = sblen;
+}
+
+void Writer::encode_source_packet_(const packet::FEC& fec) {
+    if (cur_packet_ == 0) {
+        if (!encoder_.begin(cur_sblen_, cur_rblen_)) {
```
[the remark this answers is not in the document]
dshil: I do not like this approach. `begin_block_` is an extra wrapper around `encoder.begin()` that only adds an error message to the log and no more.

I understand that you want to see actions that we perform when the block begins and actions when the blog ends.

Why you do not like the separation of encoding and writing?

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -902,5 +902,122 @@ TEST(writer_reader, repaired_block_numbering) {
...
+    CHECK(reader.valid());
+    CHECK(writer.valid());
+
+    const size_t block_sizes[] = { 20, 25, 35, 43, 33, 23, 13 };
+    CHECK(block_sizes[0] == NumSourcePackets);
+
+    packet::seqnum_t pack_n = 0;
+
+    for (size_t n = 0; n < ROC_ARRAY_SIZE(block_sizes) - 1; ++n) {
+        writer.resize(block_sizes[n + 1]);
```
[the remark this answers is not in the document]
dshil: Because we set the size of the next block.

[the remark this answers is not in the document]
dshil: Will the following comment help you understand `n + 1`?

```
We set the size of the source block before writing any packets into Writer. As a result, the updated source block size will be used only on the next block.
```

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -902,5 +902,122 @@ TEST(writer_reader, repaired_block_numbering) {
...
+            const packet::PacketPtr p = reader.read();
+
+            CHECK(p);
+            CHECK(p->fec());
+            CHECK(p->fec()->source_block_length == block_sizes[n]);
+        }
+    }
+}
+
+TEST(writer_reader, resize_souce_block_overlap) {
```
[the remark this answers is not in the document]
dshil: I would suggest:

```
* resize_souce_block_begin
* resize_source_block_middle
* resize_source_block_end
```

?

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
```diff
@@ -134,24 +139,44 @@ core::Slice<uint8_t> OFDecoder::repair(size_t index) {
...
 }
 
-void OFDecoder::reset() {
+void OFDecoder::end() {
     if (of_sess_ != NULL) {
         report_();
         destroy_session_();
     }
 
     reset_session_();
```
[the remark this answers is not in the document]
dshil: I kept the original behavior when decoder resets a session after each `end` call.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
gavv: Looks good.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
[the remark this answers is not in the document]
gavv: If I understand correctly, the original behaviour was to recreate the session exactly once per block, which is correct, and now it's possible that the session will be recreated twice, which is a bit worse.

src/modules/roc_fec/reader.cpp
```diff
@@ -82,26 +79,43 @@ packet::PacketPtr Reader::read_() {
...
+                n_packets_, (unsigned long)pp->rtp()->seqnum, (unsigned long)cur_sbn_);
 
         started_ = true;
     }
 
     return get_next_packet_();
 }
 
+void Reader::resize_(size_t sblen) {
+    const size_t cur_sblen = source_block_.size();
```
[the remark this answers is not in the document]
dshil: Agree

src/modules/roc_fec/reader.cpp
dshil: Agree.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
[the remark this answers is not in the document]
dshil: Well. We need to unify the responsibilities between `begin` and `end` because now it seems a little bit strange. Moreover, `decoder.end()` causes the session recreation but `encoder.end()` does not.

src/modules/roc_fec/idecoder.h
dshil: I got it. Fixed

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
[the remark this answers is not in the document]
gavv: Why is this important? Session (re)creation is an implementation detail hidden from the user. It can be created in begin, end, or even is set on demand -- as for me it doesn't change anything. On the other hand, recreating session twice is an early pessimization, and if I get it right it can be fixed trivially by removing this reset_session_() call. Isn't it?

src/modules/roc_fec/writer.cpp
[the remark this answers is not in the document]
gavv: > Why you do not like the separation of encoding and writing?

I'm OK with that, we can keep them in separate methods. E.g., you can extract the fill-and-compose part into a new method and we'll get something like:

```
    encoder_.set(cur_packet_, fec.payload);
    fill_packet_(pp);
    writer_.write(pp);
```

> begin_block_ is an extra wrapper around encoder.begin() that only adds an error message to the log and no more.

Yes, but I think it's not that bad because:

* Writer::begin_block_ and encoder.begin() are different abstraction layers. It's usually not recommended to mix different abstraction layers (google for "single level of abstraction principle") because it makes the code harder to understand.

* This wrapper removes some boilerplate (logging) from the Writer::write which allows us to keep that method more concise and high-level and the begin/fill/end flow becomes more clear.

If you still don't think this would be more readable, let's leave this as is. But in this case I suggest at least to hide the low-level code that increments and check the packet number:

```
    cur_packet_++;

    if (cur_packet_ == cur_sblen_) {
        handle_next_block_();
    }
```

inside the handle_next_block_ method (and maybe rename it accordingly).

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
gavv: > As a result the updated source block size will be used only on the next block.

Why is this so even if we call writer.resize *before* writing the block? I would expect that if the resize() is called before writing the first packet in a block, it will take effect starting from *this* block.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
gavv: > One way to do it is to add a test checking that the writer produces correct output when we're changing the block size.

BTW the test I suggested should be added to test_encoder_decoder instead of test_writer_reader.

src/modules/roc_fec/reader.cpp
```diff
@@ -76,32 +73,55 @@ packet::PacketPtr Reader::read_() {
...
-                source_ = pp->rtp()->source;
-                has_source_ = true;
-            }
-            cur_sbn_ = pp->fec()->source_block_number;
-            drop_repair_packets_from_prev_blocks_();
-        }
+        packet::PacketPtr pp = read_first_();
 
         if (!pp || pp->fec()->encoding_symbol_id > 0) {
             return source_queue_.read();
```
gavv: Do we still need this condition?

```
pp->fec()->encoding_symbol_id > 0
```

It probably made sense only when we relied on the marker bit? Not sure but anyway it looks unnecessary.

If we don't need it anymore, this check:

```
        if (!pp || pp->fec()->encoding_symbol_id > 0) {
            return source_queue_.read();
        }
```

could be reduced to:

```
        if (!pp) {
            return NULL;
        }
```

and we'll get the ability to repair lost packets event in the very fist block.

gavv:
https://github.com/roc-project/roc/issues/48#issuecomment-486686125

Please rebase and re-target on the develop branch back :)

src/modules/roc_fec/reader.cpp
dshil: Hm.. Why we can simply return NULL instead of `return source_queue_.read()`?
We have an `OR` condition in the mentioned if block.

gavv: Because there is no difference between this:

```
        packet::PacketPtr pp = source_queue_.head();

        if (!pp) {
            return source_queue_.read();
        }
```

and this:

```
        packet::PacketPtr pp = source_queue_.head();

        if (!pp) {
            return NULL;
        }
```

(well, there is no *practical* difference)

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
dshil: You are right but it was implemented in a slightly different way. resize just promises you that next block will have the provided size. I guess that we need also to handle the case when resize is called at the beginning of the block.

src/modules/roc_fec/writer.cpp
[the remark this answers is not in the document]
dshil: I have reworked it a little bit. I still do not like it but it seems better than it was...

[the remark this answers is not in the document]
gavv: Looks pretty good.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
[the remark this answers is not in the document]
dshil: According to https://github.com/roc-project/roc/pull/174#discussion_r277382695 I removed this if block.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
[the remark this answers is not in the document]
dshil: It is not important at all and you are right. The extra call to `reset_session_` has removed.

src/modules/roc_fec/reader.cpp
dshil: Agree. Lets extract it into separate PR? I also want to cover this with test.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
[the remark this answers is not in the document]
dshil: I think I will forget about it. Lets keep it unresolved until I will create an issue for it.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
dshil: It is too earlier to resolve it because the tests are in progress.

src/modules/roc_fec/reader.cpp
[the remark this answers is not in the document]
dshil: I have unresolved it because I am not sure was this fixed or not.

src/modules/roc_fec/reader.cpp
dshil: https://github.com/roc-project/roc/pull/174#discussion_r283099439

src/modules/roc_fec/reader.cpp
gavv: Sure.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
dshil: Tests were added.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
dshil: I have reworked. FYI: https://github.com/roc-project/roc/pull/174#discussion_r283100657

src/modules/roc_fec/reader.cpp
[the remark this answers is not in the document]
dshil: Done

src/modules/roc_fec/reader.cpp
dshil: Done

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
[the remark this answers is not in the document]
dshil: https://github.com/roc-project/roc/issues/178

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
dshil: > It would be great if our test suite would also check that block sizes are really changing.
> We set it to <10, 5> and it produces 10 source then 5 repair packets. We set it to <15, 5> and it produces 15 source then 5 repair packets.

Note, that test validates that the source block size has really changed:

```
        for (size_t i = 0; i < block_sizes[n]; ++i) {
            const packet::PacketPtr p = reader.read();

            CHECK(p);
            CHECK(p->fec());
            CHECK(p->fec()->source_block_length == block_sizes[n]);
        }
```

We just need to ensure that we have the same behavior in case of repair packets.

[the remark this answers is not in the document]
dshil: > BTW the test I suggested should be added to test_encoder_decoder instead of test_writer_reader.

It is not clear why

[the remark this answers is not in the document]
dshil: I've added a test `TEST(writer_reader, lossy_resize_source_block)` that somehow verifies the correctness of the block resizing. I think it is enough for the current PR.

src/modules/roc_fec/reader.cpp
```diff
@@ -162,6 +181,11 @@ void Reader::try_repair_() {
         return;
     }
 
+    if (!decoder_.begin(source_block_.size(), repair_block_.size())) {
+        alive_ = false;
```
gavv: We should log the reason every time we're shutting down, e.g.:

```
        roc_log(LogDebug,
                "fec reader: can't begin decoder block, shutting down:"
                " sbl=%lu rbl=%lu", ...);
```

Also, we are we shutting down here? Shouldn't we skip a block instead?

src/modules/roc_fec/reader.cpp
```diff
@@ -288,16 +312,18 @@ void Reader::update_source_packets_() {
...
+                    "fec reader: dropping source packet from previous block:"
+                    " cur_sbn=%lu pkt_sbn=%lu pkt_sn=%lu",
+                    (unsigned long)cur_sbn_, (unsigned long)fec->source_block_number,
+                    (unsigned long)rtp->seqnum);
             n_dropped++;
             continue;
         }
 
         roc_panic_if(fec->source_block_number != cur_sbn_);
+        update_size_(fec->source_block_length);
```
gavv: As discussed, we should do this only at the beginning of a block.

src/modules/roc_fec/reader.cpp
gavv: #186

src/modules/roc_fec/target_openfec/roc_fec/of_encoder.cpp
```diff
@@ -99,10 +104,10 @@ void OFEncoder::set(size_t index, const core::Slice<uint8_t>& buffer) {
     buff_tab_[index] = buffer;
 }
 
-void OFEncoder::commit() {
+void OFEncoder::fill() {
     roc_panic_if_not(valid());
 
-    for (size_t i = blk_source_packets_; i < blk_source_packets_ + blk_repair_packets_;
+    for (size_t i = sblen_; i < sblen_ + rblen_;
          ++i) {
```
gavv: Please run `scons fmt`

src/modules/roc_fec/writer.cpp
```diff
@@ -68,51 +69,85 @@ void Writer::write(const packet::PacketPtr& pp) {
         } while (source_ == pp->rtp()->source);
     }
 
+    if (cur_packet_ == 0) {
+        if (!begin_block_()) {
+            return;
```
gavv: As discussed, we should either set alive_ to false here or better skip current block.

src/modules/roc_fec/writer.cpp
```diff
@@ -68,51 +69,85 @@ void Writer::write(const packet::PacketPtr& pp) {
...
 
-void Writer::handle_next_block_() {
+void Writer::next_block_() {
+    cur_block_repair_sn_ += cur_rblen_;
+    cur_sbn_++;
+
+    cur_packet_ = 0;
+    cur_sblen_ = next_sblen_;
+
     roc_log(LogTrace, "fec writer: next block: sbn=%lu", (unsigned long)cur_sbn_);
```
gavv: Please also add sblen and rblen to the log message.

src/modules/roc_fec/writer.cpp
```diff
@@ -68,51 +69,85 @@ void Writer::write(const packet::PacketPtr& pp) {
...
 
-    if (cur_packet_ == n_source_packets_) {
-        handle_next_block_();
+void Writer::end_block_() {
+    make_repair_packets_();
 
-        cur_block_repair_sn_ += n_repair_packets_;
-        cur_sbn_++;
+    encode_repair_packets_();
+    encoder_.fill();
```
gavv: I think we should either move `encoder_.fill()` call inside  encode_repair_packets_() or rename encode_repair_packets_() to something like `prepare_repair_packets_()`, because currently it does no encoding.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -898,5 +898,199 @@ TEST(writer_reader, repaired_block_numbering) {
...
+            CHECK(p);
+            CHECK(p->fec());
+            CHECK(p->fec()->source_block_length == prev_sblen);
+        }
+
+        prev_sblen = block_sizes[n];
+    }
+}
+
+TEST(writer_reader, resize_source_block_end) {
```
gavv: Isn't this test equivalent to resize_source_block_begin?

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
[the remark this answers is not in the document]
gavv: > Note, that the test validates that the source block size has really changed:

It check the SBL *field*, but not the *real* block size.

> It is not clear why.

My bad, I was wrong with that.

> I've added a test TEST(writer_reader, lossy_resize_source_block) that somehow verifies the correctness of the block resizing. I think it is enough for the current PR.

LGTM.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -898,5 +898,199 @@ TEST(writer_reader, repaired_block_numbering) {
...
+            CHECK(p->fec());
+            CHECK(p->fec()->source_block_length == prev_sblen);
+        }
+
+        prev_sblen = block_sizes[n];
+        writer.resize(block_sizes[n]);
+    }
+}
+
+TEST(writer_reader, lossy_resize_source_block) {
```
gavv: The name is strange, I see no packet losses in this test.

I suggest to rename it to something like "write_resize_block".

src/modules/roc_fec/reader.cpp
gavv: (see #180)

src/modules/roc_fec/reader.cpp
dshil: `begin` is usually failed when there is no enough memory. I am not sure if skipping a block or restarting a session will actually help us. I guess restarting a session will be more suitable because it requires more time and probably some memory will be released.

gavv: My logic was the following: we can skip the "bad" block and try again next block, or next, or next.. If there are only a few bad blocks, we'll continue playing soon. And if not, watchdog will detect the silence and kill the session.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
dshil: Not exactly, in `_begin` we resize writer `writer.resize(block_sizes[n])` before writing any packets, on the other side in `_end` we resize writer after writing all packets for current block.

They are equal according to the implementation details. BTW I want to keep them both because of their conceptual difference.

src/modules/roc_fec/writer.cpp
dshil: I have moved `encoder_.fill()` into `encode_repair_packets_`. As a result the remaining function sequence looks neat and readable, IMO:

```
void Writer::end_block_() {
    make_repair_packets_();
    encode_repair_packets_();
    write_repair_packets_();

    encoder_.end();
}
```

src/modules/roc_fec/writer.cpp
dshil: Done

src/tests/roc_fec/target_openfec/test_writer_reader.cpp

src/modules/roc_fec/target_openfec/roc_fec/of_encoder.cpp
dshil: Done

src/modules/roc_fec/writer.cpp
dshil: As discussed the panic was added :)

src/modules/roc_fec/reader.cpp
dshil: Discussed to do it separately during #174.

gavv: #179

src/modules/roc_fec/writer.cpp
gavv: Will be fixed in #179.

src/modules/roc_fec/reader.cpp
dshil: Discussed to set `alive_ = false` in the current PR.

PR #188, opened 2019-05-16
FEC: add test for resize + losses

gavv:
Minor improvements in existing tests plus one new test for resize + losses.

Changes are splitted into commits for easy reviewing.

### DIFF
--- a/src/tests/roc_fec/target_openfec/test_writer_reader.cpp
+++ b/src/tests/roc_fec/target_openfec/test_writer_reader.cpp
@@ -1054,5 +1054,58 @@ TEST(writer_reader, resize_block_middle) {
     UNSIGNED_LONGS_EQUAL(wr_sn, rd_sn);
 }
 
+TEST(writer_reader, resize_block_losses) {
+    OFEncoder encoder(config, FECPayloadSize, allocator);
+    OFDecoder decoder(config, FECPayloadSize, buffer_pool, allocator);
+
+    CHECK(decoder.valid());
+    CHECK(encoder.valid());
+
+    PacketDispatcher dispatcher;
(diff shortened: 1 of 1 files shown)


src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -898,7 +898,7 @@ TEST(writer_reader, repaired_block_numbering) {
     }
 }
 
-TEST(writer_reader, resize_source_block_begin) {
+TEST(writer_reader, writer_block_sizes) {
```
dshil: As we do not use reader we can remove decoder as well.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -941,11 +983,14 @@ TEST(writer_reader, resize_source_block_begin) {
             CHECK(p);
             CHECK(p->fec());
             CHECK(p->fec()->source_block_length == block_sizes[n]);
+
+            check_audio_packet(p, rd_sn);
+            rd_sn++;
```
dshil: It is worth to check if `wr_sn == rd_sn` at the end.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
gavv: Done.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
gavv: Indeed, done.

Issue #178, opened 2019-05-11
FEC: validate source block length


dshil:
## Issue

Due to #174 now it is possible to change the source block dynamically. As a
result, invalid value for the source block can be set and we will get an error
from the OpenFEC library because different schemes have different limits for
the maximum value of the source block size. ROC panics because of this error.

## Solution

**Writer**

If the size of a source block exceeds the limit of the selected scheme, 
writer should not update the size of source block.

**Reader**

If the size of a source block exceeds the limit of the selected scheme, 
the reader should ignore packets from this block.

For more info see:

* https://github.com/roc-project/roc/pull/174#discussion_r277381255
* https://www.freelists.org/post/roc/fec-writer-alive

gavv:
We can add IEncoder::max_source_block() and IDecoder::max_source_block(), report them from OFEncoder and OFDecoder depending on the selected scheme and and check them in Writer and Reader.

PR #189, opened 2019-05-17
FEC: validate source block length

dshil:
Closes: #178

### DIFF
--- a/src/modules/roc_fec/idecoder.h
+++ b/src/modules/roc_fec/idecoder.h
@@ -23,6 +23,9 @@ class IDecoder {
 public:
     virtual ~IDecoder();
 
+    //! Get the maximum number of encoding symbols for the scheme being used.
+    virtual size_t max_block_length() const = 0;
+
     //! Start block.
     //!
     //! @remarks
--- a/src/modules/roc_fec/reader.cpp
+++ b/src/modules/roc_fec/reader.cpp
@@ -175,6 +175,19 @@ bool Reader::update_block_size_(size_t new_sblen) {
         return false;
     }
 
+    const size_t new_blen = new_sblen + repair_block_.size();
+
+    if (new_blen > decoder_.max_block_length()) {
+        roc_log(LogDebug,
+                "fec reader: can't update block length,"
+                " maximum value exceeded: shutting down:"
+                " cur_sbl=%lu cur_rbl=%lu new_sbl=%lu max_blen=%lu",
+                (unsigned long)cur_sblen, (unsigned long)repair_block_.size(),
+                (unsigned long)new_sblen, (unsigned long)decoder_.max_block_length());
+        return (alive_ = false);
+    }
+
+
     if (!source_block_.resize(new_sblen)) {
         roc_log(LogDebug,
                 "fec reader: can't allocate source block table, shutting down:"
--- a/src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
+++ b/src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
@@ -71,6 +71,25 @@ bool OFDecoder::valid() const {
     return valid_;
 }
 
+size_t OFDecoder::max_block_length() const {
+    roc_panic_if_not(valid());
+
+    size_t ret = 0;
+
+    switch((int)codec_id_) {
+    case OF_CODEC_REED_SOLOMON_GF_2_M_STABLE:
+        ret = OF_REED_SOLOMON_MAX_NB_ENCODING_SYMBOLS_DEFAULT;
+        break;
+    case OF_CODEC_LDPC_STAIRCASE_STABLE:
+        ret = OF_LDPC_STAIRCASE_MAX_NB_ENCODING_SYMBOLS_DEFAULT;
+        break;
+    default:
+        roc_panic("unexpected fec scheme");
+    }
+
+    return ret;
+}
+
 bool OFDecoder::begin(size_t sblen, size_t rblen) {
     roc_panic_if_not(valid());
 
--- a/src/modules/roc_fec/target_openfec/roc_fec/of_encoder.h
+++ b/src/modules/roc_fec/target_openfec/roc_fec/of_encoder.h
@@ -52,6 +52,9 @@ public:
     //! Get buffer alignment requirement.
     virtual size_t alignment() const;
 
+    //! Get the maximum number of encoding symbols for the scheme being used.
+    virtual size_t max_block_length() const;
+
     //! Start block.
     //!
     //! @remarks
--- a/src/modules/roc_fec/writer.h
+++ b/src/modules/roc_fec/writer.h
@@ -58,7 +58,7 @@ public:
     bool alive() const;
 
     //! Set number of source packets per block.
-    void resize(size_t sblen);
+    bool resize(size_t sblen);
 
     //! Write packet.
     //! @remarks
--- a/src/tests/roc_fec/target_openfec/test_encoder_decoder.cpp
+++ b/src/tests/roc_fec/target_openfec/test_encoder_decoder.cpp
@@ -188,5 +188,20 @@ TEST(encoder_decoder, load_test) {
     }
 }
 
+TEST(encoder_decoder, max_source_block) {
+    size_t test_cases[] = {
+        OF_REED_SOLOMON_MAX_NB_ENCODING_SYMBOLS_DEFAULT,
+        OF_LDPC_STAIRCASE_MAX_NB_ENCODING_SYMBOLS_DEFAULT
+    };
+
+    for (size_t n_scheme = 0; n_scheme < Test_n_fec_schemes; ++n_scheme) {
+        config.scheme = Test_fec_schemes[n_scheme];
+        Codec code(config);
+
+        CHECK(code.encoder().max_block_length() == test_cases[n_scheme]);
+        CHECK(code.decoder().max_block_length() == test_cases[n_scheme]);
+    }
+}
+
 } // namespace fec
 } // namespace roc
--- a/src/tests/roc_fec/target_openfec/test_writer_reader.cpp
+++ b/src/tests/roc_fec/target_openfec/test_writer_reader.cpp
@@ -1256,7 +1256,7 @@ TEST(writer_reader, writer_resize_blocks) {
         packet::seqnum_t pack_n = 0;
 
         for (size_t n = 0; n < ROC_ARRAY_SIZE(block_sizes); ++n) {
-            writer.resize(block_sizes[n]);
+            CHECK(writer.resize(block_sizes[n]));
 
             core::Array<packet::PacketPtr> packets(allocator);
             packets.resize(block_sizes[n]);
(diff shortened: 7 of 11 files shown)


src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
```diff
@@ -71,6 +71,20 @@ bool OFDecoder::valid() const {
     return valid_;
 }
 
+size_t OFDecoder::max_source_block() const {
+    roc_panic_if_not(valid());
+
+    if (codec_id_ == OF_CODEC_REED_SOLOMON_GF_2_M_STABLE) {
+        return OF_REED_SOLOMON_MAX_NB_SOURCE_SYMBOLS_DEFAULT;
```
gavv: I've checked OpenFEC sources and it appears that these defines are just defaults (as the name suggests).

OpenFEC provides two function to get and set these and other codec parameters:

* of_get_control_parameter
* of_set_control_parameter

However, of_set_control_parameter() is not implemented for LDPC and RS currently.

I suggest to do the following:

* use of_get_control_parameter() instead of defines;

* create an issue (in Roc) for:
  * crating a patch for OpenFEC that implements of_set_control_parameter()
  * using of_set_control_parameter() in Roc sender with the block sizes provided by user (to remove any limitation)
  * using of_set_control_parameter() in Roc receiver either with the block sizes provided by session (to remove any limitation) OR with the maximum block sizes set by user via the receiver config (to keep some limit possible but make it configurable)

src/modules/roc_fec/reader.cpp
```diff
@@ -114,6 +114,15 @@ bool Reader::update_size_(size_t new_sblen) {
         return true;
     }
 
+    if (new_sblen > decoder_.max_source_block()) {
```
gavv: The documentation for max_source_block() states that it can return zero, but there is no check here. We should either fix the documentation or the usage.

If we will use of_get_control_parameter(), we probably can just panic on its error (it should not happen) and fix documentation.

src/modules/roc_fec/writer.h
```diff
@@ -55,7 +55,7 @@ class Writer : public packet::IWriter, public core::NonCopyable<> {
     bool valid() const;
 
     //! Set number of source packets per block.
-    void resize(size_t sblen);
+    bool resize(size_t sblen);
```
gavv: We should also replace `resize(n)` with `CHECK(resize(n))` everywhere in tests.

src/tests/roc_fec/target_openfec/test_encoder_decoder.cpp
```diff
@@ -182,5 +191,10 @@ TEST(encoder_decoder, load_test) {
     }
 }
 
+TEST(encoder_decoder, max_source_block) {
+    check_max_source_block(ReedSolomon8m, OF_REED_SOLOMON_MAX_NB_SOURCE_SYMBOLS_DEFAULT);
+    check_max_source_block(LDPCStaircase, OF_LDPC_STAIRCASE_MAX_NB_SOURCE_SYMBOLS_DEFAULT);
```
gavv: Other encoder-decoder tests use a cycle over all codecs:

```
for (int type = ReedSolomon8m; type != CodecTypeMax; ++type) {
```

It's better to use it here as well. And thus we can also remove (inline) the check_max_source_block() method.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -1043,5 +1154,22 @@ TEST(writer_reader, writer_resize_block) {
     }
 }
 
+TEST(writer_reader, source_block_max) {
+    check_writer_source_block_max(rs_source_composer, rs_repair_composer, ReedSolomon8m,
```
gavv: I suggest to use the same approach as in encoder-decoder tests:

* create a cycle over all codec types in every test;
* inline test code (check_writer_source_block_max) in the test itself.

Pros:

* will automatically work with new codec if we add them;
* a single test-case code is not dispersed in different parts of the file.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
gavv: Discussed that we still need to validate that block size fits into FECFRAME packet limitations (1 or 2 bytes for ESI) so we still need a check. Leaving it as is for now, we'll add of_set_control_parameter() support later.

src/modules/roc_fec/idecoder.h
```diff
@@ -23,14 +23,31 @@ class IDecoder {
 public:
     virtual ~IDecoder();
 
+    //! Get the maximum value of a source block for a used scheme.
+    //!
+    //! @returns
+    //!  0 when there are no limits.
+    virtual size_t max_source_block() const = 0;
```
gavv: Discussed to replace it with max_block_length().

src/modules/roc_fec/writer.h
dshil: Done

src/modules/roc_fec/idecoder.h
dshil: Done

src/tests/roc_fec/target_openfec/test_writer_reader.cpp

src/tests/roc_fec/target_openfec/test_encoder_decoder.cpp
dshil: Done

src/modules/roc_fec/reader.cpp
dshil: Reworked.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
gavv: #193

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
```diff
@@ -71,6 +71,16 @@ bool OFDecoder::valid() const {
...
 }
 
+size_t OFDecoder::max_block_length() const {
+    roc_panic_if_not(valid());
+
+    if (codec_id_ == OF_CODEC_REED_SOLOMON_GF_2_M_STABLE) {
+        return OF_REED_SOLOMON_MAX_NB_SOURCE_SYMBOLS_DEFAULT;
+    }
+
+    return OF_LDPC_STAIRCASE_MAX_NB_SOURCE_SYMBOLS_DEFAULT;
```
gavv: I suggest a switch with a panic in default branch, so that we notice the problem if we add a new FEC scheme ad forget to patch this method.

src/modules/roc_fec/target_openfec/roc_fec/of_encoder.h
```diff
@@ -52,6 +52,9 @@ class OFEncoder : public IEncoder, public core::NonCopyable<> {
     //! Get buffer alignment requirement.
     virtual size_t alignment() const;
 
+    //! Get the maximum value of a source block for a used scheme.
+    virtual size_t max_block_length() const;
```
gavv: Comment is outdated.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -1653,5 +1653,129 @@ TEST(writer_reader, error_reader_decode_packet) {
...
+                      repair_composer(), packet_pool, buffer_pool, allocator);
+
+        Reader reader(config, decoder, dispatcher.source_reader(),
+                      dispatcher.repair_reader(), rtp_parser, packet_pool, allocator);
+
+        CHECK(writer.valid());
+        CHECK(reader.valid());
+
+        // try to resize writer with an invalid value
+        CHECK(!writer.resize(encoder.max_block_length() * 2));
```
gavv: I suggest to try `encoder.max_block_length() + 1` instead to catch off-by-one errors.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -1653,5 +1653,129 @@ TEST(writer_reader, error_reader_decode_packet) {
...
+        // ensure that the block size was not updated
+        fill_all_packets(0);
+
+        // write packets to dispatcher
+        for (size_t i = 0; i < NumSourcePackets; ++i) {
+            writer.write(source_packets[i]);
+        }
+
+        // deliver packets from dispatcher to reader
+        dispatcher.push_written();
```
gavv: Let's check queue sizes before calling dispatcher to ensure that writer generated a block of correct size:

```
        CHECK(dispatcher.source_size() == NumSourcePackets);
        CHECK(dispatcher.repair_size() == NumRepairPackets);
```

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -1653,5 +1653,129 @@ TEST(writer_reader, error_reader_decode_packet) {
...
+
+        // write packets from queue to dispatcher
+        for (size_t i = 0; i < NumSourcePackets + NumRepairPackets; ++i) {
+            packet::PacketPtr p = queue.read();
+            CHECK(p);
+            if (i == 5) {
+                // violates: SBL + RBL > MAX_NES (for source packets)
+                p->fec()->source_block_length = encoder.max_block_length() * 2;
+                // reset block length to ensure that our packet won't be dropped
+                p->fec()->block_length = 0;
```
gavv: I think we need a test for both cases when block_length is zero and non-zero.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
```diff
@@ -1653,5 +1653,129 @@ TEST(writer_reader, error_reader_decode_packet) {
...
+            }
+            if (i == NumSourcePackets + 3) {
+                // violates: SBL + RBL > MAX_NES (for repair packets)
```
gavv: `SBL + RBL <= MAX_NES`

also I think you need to remove " (for repair packets)"

src/modules/roc_fec/target_openfec/roc_fec/of_encoder.h
dshil: s/for a used scheme/for a scheme being used/

dshil: Done

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
dshil: Done

src/tests/roc_fec/target_openfec/test_writer_reader.cpp

src/tests/roc_fec/target_openfec/test_writer_reader.cpp
dshil: Done. Also additional block is processed.

src/tests/roc_fec/target_openfec/test_writer_reader.cpp

Issue #181, opened 2019-05-15
FEC: dynamic configuration


gavv:
- [x] Detect FEC scheme from port.

  Currently, the receiver is configured to handle a single FEC scheme which is used for every session. Instead, we should allow to bind multiple ports employing different FEC schemes and automatically select the appropriate scheme when creating a new session, depending on which port we got its packets. We also need a test for that.

- [x] Implement port notation.

  Currently tools are operating address notations ("host:port", e.g. "127.0.0.1:123"). This notation does not allow to specify arbitrary protocol for every port. We should replace it with a new port notation ("proto:host:port", e.g. "ldpc:127.0.0.1:123"). We should use it in roc-send and roc-recv.

- [x] Support for multiple ports in roc-recv.

  After removing FEC configuration from the receiver, we should improve the roc-recv command-line interface to allow the user to bind multiple ports with different protocols, e.g. two Reed-Solomon ports, two LDPC ports, and one bare RTP port.

- [x] Remove static configuration from API and roc-recv.

  After finishing dynamic block and packet size support (#180), we should remove the following parameters from the API: packet_length, fec_code, fec_block_source_packets, fec_block_repair_packets.

Issue #180, opened 2019-05-15
FEC: handle dynamic packet and block sizes


gavv:
- [x] Handle dynamic # of source packets.

  - [x] Retrieve it from the SBL field of FECFRAME header. Apply at the beginning of every block.

  - [x] If it's changed in the middle of a block, drop packets with mismatching block length until next block.

- [x] Handle dynamic # of repair packets:

  - [x] Retrieve it from Number Encoding Symbol field of FECFRAME headers. If it's changed in the middle of a block, drop packets with mismatching block length until next block.

  - [x] For Reed-Solomon, always use 255 as the block length.

- [x] Handle dynamic packet size. Retrieve it from payload size of the FECFRAME packet. Apply at the beginning of every block. If it's changed in the middle of a block, drop packets with mismatching block length until next block.

- [x] Add a test for allocator error in writer (error_writer_resize_block).

- [x] Add a test for the very specific case when # of repair symbols is zero (i.e. source block length == number encoding symbols). We should correctly work in this case.

PR #195, opened 2019-05-26
FEC: final fixes for dynamic config

gavv:
Fixes and tests for #180 and #181. See commit messages.

### DIFF
--- a/src/modules/roc_fec/reader.cpp
+++ b/src/modules/roc_fec/reader.cpp
@@ -36,16 +36,13 @@ Reader::Reader(const ReaderConfig& config,
     , next_packet_(0)
     , cur_sbn_(0)
     , payload_size_(0)
-    , block_resized_(false)
+    , source_block_resized_(false)
+    , repair_block_resized_(false)
     , payload_resized_(false)
     , has_source_(false)
     , source_(0)
     , n_packets_(0)
     , max_sbn_jump_(config.max_sbn_jump) {
-    // FIXME: remove n_repair_packets from config
-    if (!repair_block_.resize(config.n_repair_packets)) {
-        return;
-    }
     valid_ = true;
 }
 
@@ -100,18 +97,20 @@ packet::PacketPtr Reader::get_first_packet_() {
         return NULL;
     }
 
-    packet::FEC& fec = *pp->fec();
-    packet::RTP& rtp = *pp->rtp();
+    const packet::RTP& rtp = *pp->rtp();
+    const packet::FEC& fec = *pp->fec();
 
-    cur_sbn_ = fec.source_block_number;
-    if (!update_block_size_(fec.source_block_length)) {
+    if (!process_source_packet_(pp)) {
         roc_log(LogTrace,
-                "fec reader: dropping first source packet: can't update block size: "
-                "esi=%lu sblen=%lu blen=%lu",
+                "fec reader: dropping leading source packet:"
+                " esi=%lu sblen=%lu blen=%lu payload_size=%lu",
                 (unsigned long)fec.encoding_symbol_id,
-                (unsigned long)fec.source_block_length, (unsigned long)fec.block_length);
+                (unsigned long)fec.source_block_length, (unsigned long)fec.block_length,
+                (unsigned long)fec.payload.size());
         return NULL;
     }
+
+    cur_sbn_ = fec.source_block_number;
 
     if (!has_source_) {
         source_ = rtp.source;
@@ -124,7 +123,7 @@ packet::PacketPtr Reader::get_first_packet_() {
 }
 
 packet::PacketPtr Reader::get_next_packet_() {
-    update_block_();
+    fill_block_();
 
     packet::PacketPtr pp = source_block_[next_packet_];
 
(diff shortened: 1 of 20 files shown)


src/modules/roc_fec/reader.cpp
```diff
@@ -374,7 +322,7 @@ void Reader::update_source_block_() {
 
         if (packet::blknum_lt(fec.source_block_number, cur_sbn_)) {
             roc_log(LogTrace,
-                    "fec reader: dropping source packet: belongs to previous block:"
+                    "fec reader: dropping source packet from  previous block:"
```
dshil: Extra space between `from` and `previous`.

src/modules/roc_fec/reader.cpp
```diff
@@ -385,39 +333,17 @@ void Reader::update_source_block_() {
         // should not happen: we have handled preceding and following blocks above
         roc_panic_if_not(fec.source_block_number == cur_sbn_);
 
-        if (!validate_incoming_source_packet_(pp)) {
+        if (!process_source_packet_(pp)) {
```
dshil: s/process_source_packet_/handle_source_packet_ ?
The same for the repair packets.

src/modules/roc_fec/reader.cpp
```diff
@@ -461,7 +387,7 @@ void Reader::update_repair_block_() {
 
         if (packet::blknum_lt(fec.source_block_number, cur_sbn_)) {
             roc_log(LogTrace,
-                    "fec reader: dropping repair packet: belongs to previous block:"
+                    "fec reader: dropping repair packet from to previous block:"
```
dshil: s/from to/from previous block/

src/modules/roc_fec/reader.cpp
```diff
@@ -552,19 +429,87 @@ void Reader::update_repair_block_() {
...
+
+    if (!check_payload_size_(fec.payload.size())) {
+        return false;
+    }
+
+    if (!check_source_block_size_(fec.source_block_length)) {
+        return false;
+    }
+
+    if (!update_payload_size_(fec.payload.size())) {
```
dshil: It is still possible to update payload size and fail to update source block size.

src/modules/roc_fec/reader.cpp
```diff
@@ -626,6 +579,157 @@ bool Reader::validate_sbn_sequence_(const packet::PacketPtr& pp) {
...
+    if (source_block_resized_) {
+        roc_log(LogDebug,
+                "fec reader: can't change source block size in the middle of a block:"
+                " next_esi=%lu cur_sblen=%lu new_sblen=%lu",
+                (unsigned long)next_packet_, (unsigned long)cur_sblen,
+                (unsigned long)new_sblen);
+        return false;
+    }
+
+    if (new_sblen > decoder_.max_block_length()) {
```
dshil: As we do not allow zero number of repair packets we should ensure that `new_sblen` also does not equal to `decoder_.max_block_length()`.

src/modules/roc_fec/reader.cpp
```diff
@@ -585,6 +534,10 @@ bool Reader::validate_incoming_repair_packet_(const packet::PacketPtr& pp) {
         }
     }
 
+    if (fec.payload.size() == 0) {
```
dshil: Could we move this check into `check_payload_size_` ?

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
```diff
@@ -109,7 +109,7 @@ void OFDecoder::set(size_t index, const core::Slice<uint8_t>& buffer) {
     }
 
     if (buffer.size() == 0 || buffer.size() != payload_size_) {
-        roc_panic("of decoder: invalid payload size: size=%lu, expected=%lu",
+        roc_panic("of decoder: invalid payload size: new=%lu cur=%lu",
```
dshil: It is more intuitive to read `cur=%lu new=%lu`, IMO.

src/modules/roc_fec/writer.cpp
```diff
@@ -147,6 +146,23 @@ void Writer::write(const packet::PacketPtr& pp) {
     }
 }
 
+bool Writer::resize_repair_block_(size_t rblen) {
+    if (repair_packets_.size() == rblen) {
+        return true;
+    }
+
+    if (repair_packets_.resize(rblen)) {
```
dshil: Lets rename `repair_packets` into `repair_block`? Because field name and function name seems to be inconsistent a little bit.

src/modules/roc_fec/writer.cpp
```diff
@@ -256,6 +277,21 @@ void Writer::encode_repair_packets_() {
     encoder_.fill();
 }
 
+void Writer::compose_repair_packets_() {
+    for (packet::seqnum_t i = 0; i < cur_rblen_; i++) {
```
dshil: I guess that we can use `size_t` instead of `packet::seqnum_t`.

src/modules/roc_fec/writer.cpp
dshil: Well, we can leave it as is for now because we have some functions that have `block` in their name and other functions that have `packets` in their name, but packets and block are the same concept in the current context.

src/modules/roc_fec/reader.cpp
gavv: What's the difference?

src/modules/roc_fec/reader.cpp
gavv: If update_source_block_size_() or update_repair_block_size_() fails, alive_ is always set to false, so that's ok.

src/modules/roc_fec/reader.cpp
gavv: This is already handled in validate_incoming_repair_packet_(). We have a test for this BTW.

src/modules/roc_fec/reader.cpp
gavv: Why? This is a part of packet validation.

In validate_incoming_source_packet_() and validate_incoming_repair_packet_() we drop invalid packets.

In check_payload_size_(), check_source_block_size_(), and check_repair_block_size_() we check whether the block or packet size update may be applied.

We probably can rename them somehow.

gavv: I'll rename `check_*` to `can_update_*`.

src/modules/roc_fec/reader.cpp
gavv: Fixed.

src/modules/roc_fec/reader.cpp
gavv: Fixed.

src/modules/roc_fec/reader.cpp
gavv: Renamed.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
gavv: Agree, fixed.

src/modules/roc_fec/writer.cpp
gavv: Renamed for consistency with reader.

src/modules/roc_fec/writer.cpp
gavv: Done.

PR #202, opened 2019-05-31
API: cleanup comments

dshil:


gavv:
It seems that travis didn't catch the branch change and was running this on master.

Issue #186, opened 2019-05-16
FEC: don't drop leading packets in FEC reader


gavv:
Currently FEC reader drops leading packets until it receives the first packet in a block.

This is bad for two reasons:

* It may unintentionally decrease the latency up to SBL-1 packets, because we first accumulate the desired latency in the queue and than start the FEC reader which drops these packets.

* It may unintentionally increase the latency the start time if several block are missing the first packet or two.

We should replace this code in reader:

        if (!pp || pp->fec()->encoding_symbol_id > 0) {
            return source_queue_.read();
        }

with:

        if (!pp) {
            return NULL;
        }

On the other hand, it *may* make sense to zeroize the first (incomplete) FEC block because it potentially has lower quality of service than other blocks (because FEC decoder has less packets that can be used to restore losses).

However, the right way to do it is one of the following:

* either take FEC block boundaries into account in packet::Delayer, before fec::Reader;
* or zeroize leading frames (if they are incomplete) in the frame pipeline after fec::Reader and audio::Depacketizer.

Both of this approaches will not affect the latency.

I think we should first fix fec::Reader and then test the new behavior for a while. If we'll find  problems with it, we can implement one of these approaches.

gavv:
> or zeroize leading frames (if they are incomplete) in the frame pipeline after fec::Reader and audio::Depacketizer.

After some though, it appears that this is a bad option. An unrecovered loss in the first block does not always led to incomplete *leading* frame. First few frames may be complete, and then we can get an incomplete one.

Thus, it seems that we can't zeroize bad first block at the frame level (after Depacketizer) without introducing additional latency. So we need to do it at the packet level (before Depacketizer and likely before fec::Reader).

gavv:
One idea is to make DelayedReader to drop packets until it meets the block boundary. More exactly, DelayedReader should drop the very first packet until it is either the first packet in a block (esi=0), or it belongs to the next block after the block of the previously dropped packet (cur.sbn > prev.sbn). This way we'll increase the cold latency (session startup time), but won't affect the runtime latency.

The "dropping" part should be done smart enough. We should take into account that when a delayed  packet is received, it can change our decision where the block boundary is. Thus we likely should not just blindly drop packets until the session startup; instead, we should somehow mark them as dropped (e.g. store in a separate queue) and exclude from latency calculations; but we should reconsider whether they should be really dropped whenever a delayed packet is received.

[...]

PR #203, opened 2019-05-31
FEC: do not drop leading packets in reader

dshil:
Closes: #186

### DIFF
--- a/src/modules/roc_fec/reader.cpp
+++ b/src/modules/roc_fec/reader.cpp
@@ -76,8 +76,8 @@ packet::PacketPtr Reader::read_() {
 
     if (!started_) {
         packet::PacketPtr pp = get_first_packet_();
-        if (!pp || pp->fec()->encoding_symbol_id > 0) {
-            return source_queue_.read();
+        if (!pp) {
+            return NULL;
         }
 
         roc_log(LogDebug,
--- a/src/tests/roc_fec/target_openfec/test_writer_reader.cpp
+++ b/src/tests/roc_fec/target_openfec/test_writer_reader.cpp
@@ -327,20 +327,23 @@ TEST(writer_reader, lost_first_packet_in_first_block) {
         for (size_t i = 0; i < NumSourcePackets; ++i) {
             writer.write(source_packets[i]);
         }
+
+        UNSIGNED_LONGS_EQUAL(NumSourcePackets * 2 - 1, dispatcher.source_size());
+        UNSIGNED_LONGS_EQUAL(NumRepairPackets * 2, dispatcher.repair_size());
+
         dispatcher.push_stocks();
 
-        // Receive every sent packet except the first one.
-        for (size_t i = 1; i < NumSourcePackets * 2; ++i) {
+        // Receive every sent packet including the first one.
+        for (size_t i = 0; i < NumSourcePackets * 2; ++i) {
             packet::PacketPtr p = reader.read();
-            if (i < NumSourcePackets) {
-                CHECK(!reader.started());
-            } else {
-                CHECK(reader.started());
-            }
+            CHECK(reader.started());
+
             check_audio_packet(p, i);
-            check_restored(p, false);
+            check_restored(p, i == 0);
         }
-        CHECK(dispatcher.source_size() == 0);
+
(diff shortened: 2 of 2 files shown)


gavv:
LGTM. How did you test this?

dshil:
> LGTM. How did you test this?

I just run the tests but I did not test in the real scenario when ROC looses a lot of packets
and accidentally meets this case.

Probably I can put a hack in which I will drop each first packet in a block and listen how the ROC behaves.

gavv:
I think you can just start the sender in the background and then restart the receiver frequently. This way receiver will be usually started in a middle of a block.

The first thing is just to check that it works.

The second thing is to do this on a bit noisy channel, where packets are lost regularly, and see whether this patch increases the probability of glitches on session start, or not.

dshil:
Well, several checks was performed:

* As you mentioned before. I started sender in the background and randomly
  started the receiver. All worked as expected.

* Emulate the noisy channel:

```sh
# Setup 250ms latency + 15% of packet loss.
sudo tc qdisc add dev lo root handle 1:0 netem delay 250msec loss 15%

# Restore the initial behaviour.
sudo tc qdisc del dev lo root
```

I've tested the channel with parameters above and did not see any problems.

P.S. I would recommend you to try to increase the % of packet loss.
ROC recovers packets very well.

I think we can merge it.

gavv:
Well, you tested it not hard enough.

I've patched the sender to drop every 7th packet to emulate frequent drops:

```
--- a/src/modules/roc_netio/target_uv/roc_netio/udp_sender.cpp
+++ b/src/modules/roc_netio/target_uv/roc_netio/udp_sender.cpp
@@ -127,6 +127,9 @@ void UDPSender::write(const packet::PacketPtr& pp) {
         roc_panic("udp sender: unexpected packet w/o data");
     }
 
+    static int C = 0;
+    if (C++ % 7 == 0) return;
+
     {
```

Then I started sender in one terminal:

```
roc-send -vv -s rtp+rs8m:127.0.0.1:10001 -r rs8m:127.0.0.1:10002 -i ~/stash/loituma.wav
```

And executed this script in another one:

```
for x in `seq 1 30`; do timeout 1 roc-recv -vv -s rtp+rs8m::10001 -r rs8m::10002 ; done |& grep --line-buffered 'of decoder: repaired' | stdbuf -oL cut -d' ' -f7- | grep --line-buffered X
```

This script restarts roc-recv 30 times every one second and greps for packets that were lost and were not recovered.

Normally this loss ration is okay and roc recovers all losses. On develop branch the output is empty.

However on you branch the output is the following:

```
XXXXXXXXXXXX.....X.. ....x....
XXXXXXXXXXXXXX...X.. ....x....
XXXXXXXXXXXXXXXX.X.. ....x....
XXXXXXX.X......X.... ..x.....
XXXXXXXXXX.....X.... ..x.....
XXXXXXXXXXXXX..X.... ..x.....
XXXXXXXXXXXXXX.X.... ..x.....
XXXXXXXXXXXXXXXXXX.. ..x.....
XXXXXXXXX....X...... x......x.
XXXXXXXXXXXX.X...... x......x.
XXXXXXXXXXXXXXXX.... x......x.
XXXXXXXXXXXXXXXXXXX. x......x.
XXXXXXXXX..X......X. .....x...
XXXXXXXXXX.X......X. .....x...
XXXXXXXXXXXXXX....X. .....x...
XXXXXXXXXXXXXXXXX.X. .....x...
XXXXXXXX.X......X... ...x.....
XXXXXXXXXXXX....X... ...x.....
XXXXXXXXXXXXXXX.X... ...x.....
```

19 runs from total 30 runs had unrecovered losses.

Let's look at this line closer:

```
XXXXXXXXXXXX.....X.. ....x....
```

The leading `XXXXXXXXXXXX` indicates that we've started in the middle of a block. This is okay, and these `X` corresponds to the silence preceding the first played packet.

However the second half of the source block (`.....X..`) indicates that we were not able to recover one lost packet, obviously because we had not enough  packets to feed the FEC decoder.

So, *there is* a degradation. It seems that we need some smarter approach.

gavv:
BTW, when we fix this, I think we also need a pipeline-level unit test for this situation.

gavv:
(Postponed)

Issue #206, opened 2019-06-01
Don't panic on bind errors


dshil:
During the IPv6 testing #15 I found the following problem.

```sh
# Specify the port number <1024. IPv4 also can be used in this case.
roc-recv -vv -s rtp+rs8m:[::1]:10001 -r rs8m:[::1]:1002
```

And got:

```
[debug] roc_sndio: pulseaudio sink: opening sink: device=(null)
[debug] roc_sndio: pulseaudio sink: opening stream: device=(null) n_channels=2 sample_rate=44100
[debug] roc_sndio: pulseaudio sink: stream_latency=0
[info] roc_netio: udp receiver: opened port [::1]:10001
[error] roc_netio: udp receiver: uv_udp_bind(): [EACCES] permission denied
[error] roc_netio: transceiver: can't add port [::1]:1002: can't start receiver

src/modules/roc_netio/target_uv/roc_netio/udp_receiver.cpp:36: error: roc_panic()

ERROR: roc_netio: udp receiver: receiver was not fully closed before calling destructor
```

`EACCESS` error is expected because we tried to use the port number <1024. Only
root can do that.

I would expect ROC to print the error message and normally shutting down.

gavv:
@dshil Just to be sure, I've re-tested this after merging your PR; the issue is fixed.

```
$ roc-recv -s rtp+rs8m:[::1]:10001 -r rs8m:[::1]:1002                                 
15:17:38.839 [err] roc_netio: udp receiver: uv_udp_bind(): [EACCES] permission denied
15:17:38.839 [err] roc_netio: transceiver: can't add port [::1]:1002: can't start receiver
15:17:38.839 [err] roc_recv: can't bind repair port: rs8m:[::1]:1002
```

PR #207, opened 2019-06-04
roc_netio: properly handle errors

dshil:
Closes: #206


src/modules/roc_netio/target_uv/roc_netio/udp_receiver.cpp
```diff
@@ -42,6 +42,13 @@ void UDPReceiver::destroy() {
 }
 
 bool UDPReceiver::start(packet::Address& bind_address) {
+    if (bind_address.port() > 0 && bind_address.port() < 1024) {
+        roc_log(LogError,
+                "udp receiver: can't open port %s: ports (0, 1024] are reserved",
```
gavv: Umm.. This is wrong. Ports below 1024 are not "reserved", they are privileged. You *can* bind to such a port, but this requires root or CAP_NET_BIND_SERVICE. This is up to the user to choose ports and privileges and we should not force any policy here. Instead, we should correctly handle errors from libuv and exit with non-zero code instead of a panic.

dshil: Did you try to run ROC under the root user? I tried and it does not work. As a result I considered that we shouldn't allow user to choose a port number in this range.

gavv: If it doesn't work, it's a bug. You can test udp ports below 1024 with sudo nc, for example.

gavv: Just checked - works for me, e.g.:

```
sudo roc-recv -vv -s rtp+rs8m::101 -r rs8m::102 -d wav -o /dev/null
```

What error did you get? Probably you were running roc-recv under sudo and pulseaudio server was not running under the root user?

dshil: >  Probably you were running roc-recv under sudo and pulseaudio server was not running under the root user

Correct

dshil: > Instead, we should correctly handle errors from libuv

Reworked.

We need to ensure that uv HANDLE will be properly closed on any errors
during UDPReceiver::start call.

We did not release it if there is an error on:

* uv_udp_bind
* uv_udp_getsockname

gavv: ### Problem 1

Unfortunately, it seems that you didn't fix the problem.

First, take a look at this code:

```
bool Transceiver::add_udp_receiver_(Task& task) {
    ...

    core::SharedPtr<UDPReceiver> rp = new (allocator_)
        UDPReceiver(loop_, *task.writer, packet_pool_, buffer_pool_, allocator_);

    ...

    if (!rp->start(*task.address)) {
        roc_log(LogError, "transceiver: can't add port %s: can't start receiver",
                packet::address_to_str(*task.address).c_str());
        return false;
    }

    receivers_.push_back(*rp);
    num_ports_++;

    return true;
}
```

If UDPReceiver::start() returns false, the UDPReceiver object is immediately destroyed.

Second, close operations are asynchronous in libuv. When you call UDPReceiver::close_() and it calls uv_close(), the handle is not closed immediately. Instead, it is marked as *closing* and will be fully closed in the next event loop iteration.

This means that in case of an error we destroy an object that is still used by libuv.

### Problem 2

Don't we have the same problem in UDPSender?

### Problem 3

I've looked at the trx and receiver code in more detail and found another related problem (not caused by this PR):

```
bool Transceiver::remove_port_(Task& task) {
    for (core::SharedPtr<UDPReceiver> rp = receivers_.front(); rp;
         rp = receivers_.nextof(*rp)) {
        if (rp->address() == *task.address) {
            rp->remove(receivers_);
            num_ports_--;
            return true;
        }
    }

    ...
}

void UDPReceiver::remove(core::List<UDPReceiver>& container) {
    roc_panic_if(container_);

    if (handle_initialized_) {
        stop();
        container_ = &container;
        address_ = packet::Address();
    } else {
        container.remove(*this);
    }
}
```

This piece correctly handles asynchronous closing, but it has a problem. It schedules the receiver for removing, keeps it in the list, and then immediately returns success to the caller.

Then the receiver will be closed in the background and removed from the list shortly. However, if the caller will be fast enough to try to rebind this port before the old receiver is closed, it will get an unexpected error because such a port is still in the list.

So there is a race. We report that the port is removed while it is not yet removed. Instead, we should wait until it is removed before returning success to the user.

### Solution

I suggest to address all these problems simultaneously.

I suggest to introduce a new list for pending-removal objects in trx. This how it should be used:

* when an object is created and started successfully, it is added to receivers_ or senders_ list, as before
* when an object is created but can't be started, asynchronous close is issued and the object is added to pending_removal_
* when a port is removed by user, asynchronous close is issued and the object is moved to pending_removal_
* UDPReceiver::remove and UDPSender::remove should accept a callback argument instead of a container; they should call that callback when the asynchronous close is finished
* Transceiver should pass them a callback that removes the object from pending_removal_ and signals the condition variable
* Transceiver::remove_port() should first wait until the task is finished and then wait until the asynchronous close is finished by looping on the condition an checking the pending_removal_ list

gavv: Also, we need unit tests for the case when bind fails. The simplest way to reproduce this error is to specify an already bound port, which can be achieved by creating the second trx and binding the same port on it.

BTW, if we had such tests already, they would detect the problem in this PR when running under clang sanitizers.

dshil: Oh..It took quite a long time.

----

* Transceiver::start(), Transceiver::join() were removed,
  now Transceiver::stop() waits all senders and receivers
  to be removed and joins the background thread.

* Transceiver automatically calls Transceiver::stop() in
  destructor if it was not called explicitly.

* Transceiver::remove_port waits until port will be actually removed.

* Transceiver::add_udp_receiver, Transceiver::add_udp_sender 
  Transceiver::remove_port should not be called after Transceiver::stop(). 
  It simplifies the internals but makes the API more strict.

* UDPSender::remove, UPDReceiver::remove were removed.

src/modules/roc_netio/target_uv/roc_netio/handle.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef ROC_NETIO_HANDLE_H_
+#define ROC_NETIO_HANDLE_H_
+
+#include "roc_packet/address.h"
+
+namespace roc {
+namespace netio {
+
+//! Network handle.
+struct Handle {
```
gavv: The name is confusing. A "handle" is usually an (opaque) reference to some resource, e.g. file handle or window handle.. This struct is something more like a callback/delegate or notification handleR.

But I suggest to get rid of this struct at all and replace it with an interface like:

```
class ICloseHandler {
public:
    virtual void closed(...) = 0;
}
```

Then Transceiver can (privately!) implement this interface and pass itself to UDPReceiver and UDPSender.

This approach is (IMHO) a bit cleaner and also easier to analyze because you can quickly check what classes implement ICloseHandler.

src/modules/roc_netio/target_uv/roc_netio/handle.h
```diff
@@ -0,0 +1,38 @@
...
+    Handle()
+        : data(NULL)
+        , fn(NULL) {
+    }
+
+    //! Data to call callback with.
+    void* data;
+
+    //! Handle callback.
+    void (*fn)(void*, packet::Address&);
```
gavv: Using address as a sender/receiver identifier is probably not very good because:

* we may want to allow sender and receiver to share the same UDP address
* we may have UDP and TCP sender/receiver with the same IP and port

I think it would be better to use `void *` as an identifier.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -214,12 +198,88 @@ void Transceiver::stop_sem_cb_(uv_async_t* handle) {
...
+            roc_panic("transceiver: can't remove unknown port %s",
+                      packet::address_to_str(address).c_str());
+        } else {
+            self.senders_.remove(*sp);
+        }
+    } else {
+        self.receivers_.remove(*rp);
+    }
+
+    self.num_ports_--;
```
gavv: I suggest to get rid of `num_ports_` field at all. Instead We can implement `num_ports()` method as follows:

```
size_t Transceiver::num_ports() const {

    return senders_.size() + receivers_.size();
}
```

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -214,12 +198,88 @@ void Transceiver::stop_sem_cb_(uv_async_t* handle) {
...
+    if (!rp) {
+        core::SharedPtr<UDPSender> sp = self.get_sender_(address);
+        if (!sp) {
+            roc_panic("transceiver: can't remove unknown port %s",
+                      packet::address_to_str(address).c_str());
+        } else {
+            self.senders_.remove(*sp);
+        }
+    } else {
+        self.receivers_.remove(*rp);
```
gavv: The nesting is a bit confusing. It would be cleaner to extract a new method that uses the same nesting for consecutive operations:

```
    core::SharedPtr<UDPReceiver> rp = self.get_receiver_(address);
    if (rp) {
        self.senders_.remove(*sp);
        return;
    }

    core::SharedPtr<UDPSender> sp = self.get_sender_(address);
    if (sp) {
        self.receivers_.remove(*sp);
        return;
    }

    roc_panic(...);
```

gavv: However, there is a better option. Current approach has one problem: we should add code for every new element type (e.g. when we'll add TCPSender and TCPReceiver) here and in a few other places. Instead, we can do the following:

* add a new interface, say, `IElement`
* make UDPReceiver and UDPSender to inherit this interface
* use a single list for senders and receivers and unify the removing code

Something like this:

```
class IElement : public RefCnt<IElement> {
public:
    virtual void stop() = 0;
};
```

This will also allow to use `IElement` instead of `void*` in `ICloseHandler`.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -152,7 +124,15 @@ bool Transceiver::add_udp_receiver(packet::Address& bind_address,
 
     run_task_(task);
 
-    return task.result;
+    if (task.flags & Task::PortError) {
+        if (task.flags & Task::PortAdded) {
+            wait_port_removed_(bind_address);
```
gavv: This approach with flags looks a bit dirty. What do you think about replacing the `flags` field with `bool success` and `IElement* element` fields and using them like this?

```
if (!tesk.success) {
    if (task.element) {
        wait_element_removed_(task.element);
    }
}

return task.success;
```

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -214,12 +198,88 @@ void Transceiver::stop_sem_cb_(uv_async_t* handle) {
     roc_panic_if_not(handle);
 
     Transceiver& self = *(Transceiver*)handle->data;
-    self.stop_();
+    self.stop_all_();
     self.close_();
     self.process_tasks_();
 }
 
+void Transceiver::close_cb_(uv_handle_t* handle) {
```
gavv: I suggest to rename some methods to make it clear what objects they are working with:

* close_cb_() -> close_sem_cb_()
* wait_stopped_() -> wait_elements_stopped_()
* wait_closed_() -> wait_sems_closed_()
* stop_all_() -> stop_all_elements_()
* close_() -> close_sems_()

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -214,12 +198,88 @@ void Transceiver::stop_sem_cb_(uv_async_t* handle) {
...
+
+void Transceiver::wait_stopped_() {
+    while (receivers_.size() || senders_.size()) {
+        cond_.wait();
+    }
+}
+
+void Transceiver::wait_closed_() {
+    while (task_sem_initialized_ || stop_sem_initialized_) {
+        cond_.wait();
```
gavv: Probably it would better to inline wait_stopped_() and wait_closed_() into stop_(). We can even unite the two loops into one.

src/tests/roc_lib/test_sender_receiver.cpp
```diff
@@ -465,13 +456,9 @@ TEST(sender_receiver, fec_with_losses) {
     Sender sender(context, sender_conf, proxy.source_addr(), proxy.repair_addr(), samples,
                   TotalSamples, FrameSamples, Flags);
 
-    proxy.start();
-
     sender.start();
     receiver.run();
     sender.join();
```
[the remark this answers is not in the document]
dshil: `receiver.run()` is a sync call, while `sender.start()` is async call.

src/tests/roc_netio/test_transceiver.cpp
```diff
@@ -36,177 +36,104 @@ packet::Address make_address(const char* ip, int port) {
 
 TEST_GROUP(transceiver) {};
 
-TEST(transceiver, noop) {
+TEST(transceiver, start) {
```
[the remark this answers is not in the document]
dshil: I would suggest to remove any `init` mentions.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -51,85 +52,56 @@ Transceiver::Transceiver(packet::PacketPool& packet_pool,
...
+    stop_handle_.fn = remove_port_cb_;
+
+    started_ = Thread::start();
 }
 
 Transceiver::~Transceiver() {
-    if (joinable()) {
-        roc_panic("transceiver: thread is not joined before calling destructor");
+    if (started_ && !stopped_) {
+        stop();
```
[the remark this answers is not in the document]
dshil: It is a destructor. Having this in mind we can guarantee that `trx.stop()` was not called, thus no one can modify this variable.

src/lib/src/context.cpp
```diff
@@ -42,7 +42,7 @@ roc_context* roc_context_open(const roc_context_config* config) {
 
     roc_log(LogInfo, "roc_context: starting context");
```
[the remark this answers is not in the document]
dshil: done

src/lib/src/context.cpp
```diff
@@ -42,7 +42,7 @@ roc_context* roc_context_open(const roc_context_config* config) {
...
 
-    if (!context->trx.start()) {
+    if (!context->trx.valid()) {
         roc_log(LogError, "roc_context_start: can't start thread");
```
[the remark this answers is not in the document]
dshil: done

src/tests/roc_netio/test_transceiver.cpp
```diff
@@ -36,177 +36,104 @@ packet::Address make_address(const char* ip, int port) {
...
-
-    trx.stop();
-    trx.join();
-
     trx.remove_port(tx_addr);
     trx.remove_port(rx_addr);
 }
 
-TEST(transceiver, start_add_stop) {
+TEST(transceiver, start_bind_duplicate) {
```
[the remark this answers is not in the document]
dshil: done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -51,85 +52,56 @@ Transceiver::Transceiver(packet::PacketPool& packet_pool,
...
     }
 
-    if (num_ports_ != 0) {
-        remove_all_ports_();
-    }
-
-    close_();
-
     if (loop_initialized_) {
```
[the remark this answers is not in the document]
dshil: done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: done

src/modules/roc_netio/target_uv/roc_netio/handle.h
dshil: done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: done

src/tests/roc_lib/test_sender_receiver.cpp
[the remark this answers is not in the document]
gavv: Yes, but it's easy to reverse this I think.

src/tests/roc_netio/test_transceiver.cpp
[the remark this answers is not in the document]
gavv: OK.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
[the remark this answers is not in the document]
gavv: The case:

1. thread1: call trx::trx()
2. thread2: call trx.stop() which sets stopped_ to true
3. thread1: call trx::~trx()

All three steps are performed sequentially (no concurrent trx calls).

The value set in `2.` (thread 2) is not guaranteed to be visible in `3.` (thread 1) because `3.` has no memory barrier (mutex or atomic).

src/modules/roc_netio/target_uv/roc_netio/iclose_handler.h
```diff
@@ -0,0 +1,36 @@
...
+class ICloseHandler {
+public:
+    virtual ~ICloseHandler();
+
+    //! Asynchronous close for network element.
+    //!
+    //! @remarks
+    //!  - Closed element should not be used.
+    //!  - Should be called from the event loop thread.
+    virtual void close(IElement&) = 0;
```
gavv: I suggest "closeD()" or "handle_closed()" because "close()" implies that the closing operation should be implemented by ICloseHandler but this is wrong.

src/modules/roc_netio/target_uv/roc_netio/ielement.h
```diff
@@ -0,0 +1,45 @@
...
+
+//! Network element interface.
+class IElement : public core::RefCnt<IElement>, public core::ListNode {
+public:
+    virtual ~IElement();
+
+    //! Asynchronous stop.
+    //! @remarks
+    //!  Should be called from the event loop thread.
+    virtual void stop() = 0;
```
gavv: If the stop is asynchronous, it would better to name it "async_stop()" or "schedule_stop()" or something. Or even better "async_close()" or "schedule_close()".

src/modules/roc_netio/target_uv/roc_netio/ielement.h
```diff
@@ -0,0 +1,45 @@
...
+
+    //! Get bind address.
+    virtual const packet::Address& address() const = 0;
+
+private:
+    friend class core::RefCnt<IElement>;
+
+    virtual void destroy() = 0;
```
gavv: It seems that IElement is more than just an interface: it's an interface plus referenc counting plus a node for list management.

Thus it probably makes sense to make two further steps:

* rename IElement to BasicElement or something like this;

* add a constructor BasicElement(allocator); make destroy() non-virtual; implement it by calling allocator_.destroy(this); remove destroy() implementation from child classes;

This will eliminate one virtual call (currently both destroy and destructor are virtual; after this change only destructor will be virtual) and arguably will make the code cleaner.

Not sure if this will be really better. Just an idea.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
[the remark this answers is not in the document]
gavv: The new implementation has no race.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
gavv: It seems that close_cb_ was not renamed.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -303,111 +299,138 @@ void Transceiver::process_tasks_() {
...
     }
 
-    if (!sp->start(*task.address)) {
-        roc_log(LogError, "transceiver: can't add port %s: can't start sender",
-                packet::address_to_str(*task.address).c_str());
+    return sp;
+}
+
+bool Transceiver::has_port_(const packet::Address& address) const {
+    return get_element_(address) != NULL;
```
gavv: I thinks we can remove this method (inline in into two invocation points).

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -203,6 +175,26 @@ void Transceiver::run() {
     roc_log(LogDebug, "transceiver: finishing event loop");
 }
 
+void Transceiver::close(IElement& element) {
+    roc_log(LogDebug, "transceiver: removed element: port %s",
```
gavv: I'd change the log message to something like "asynchronous close finished".

src/modules/roc_netio/target_uv/roc_netio/ielement.h
gavv: And, accordingly, rename "stop_all_elements_()" to something like "async_close_all_elements_()".

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
gavv: BTW, do we really need this check in make_udp_receiver_() and make_udp_sender_()?

src/modules/roc_netio/target_uv/roc_netio/ielement.h
gavv: And thereby also rename UDPReceiver::start() UDPSender::start() to open().

src/modules/roc_netio/target_uv/roc_netio/ielement.h
gavv: We can also replace term "element" with "port". I see it like this:

```
class BasicPort ... {
public:
    virtual const packet::Address& address() const = 0;

    virtual bool open(packet::Address& bind_address) = 0;

    virtual void async_close() = 0;
};

class UDPSendingPort : public BasicPort { ... };
class UDPReceivingPort : public BasicPort { ... };
```

src/modules/roc_netio/target_uv/roc_netio/udp_receiver.cpp
```diff
@@ -42,59 +44,20 @@ void UDPReceiver::destroy() {
...
 
-    if (int err = uv_udp_recv_start(&handle_, alloc_cb_, recv_cb_)) {
-        roc_log(LogError, "udp receiver: uv_udp_recv_start(): [%s] %s", uv_err_name(err),
-                uv_strerror(err));
-        return false;
+    if (handle_initialized_) {
+        stop();
+    } else {
+        close_handler_.close(*this);
     }
```
gavv: It probably makes sense to move this code:

```
    if (handle_initialized_) {
        stop();
    } else {
        close_handler_.close(*this);
    }
```

from UDPSender and UDPReceiver to Transceiver by requiring it to call stop() even when start() failed:

```
if (!rp->start()) {
    rp->stop();
    return false;
}
```

and by adding this to stop() implementation:

```
if (!handle_initialized_) {
    close_handler_.close(*this);
}
```

Why? It's quite expected that ICloseHandler may be called iniside UDPSender::stop() (at least if we'll rename it to async_close()) but it's quite unexpected that it may be called inside UDPSender::start().

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -214,76 +206,80 @@ void Transceiver::stop_sem_cb_(uv_async_t* handle) {
...
     }
+}
 
-    if (stop_sem_initialized_) {
-        uv_close((uv_handle_t*)&stop_sem_, NULL);
-        stop_sem_initialized_ = false;
+void Transceiver::stop_all_elements_() {
+    for (core::SharedPtr<IElement> pp = elements_.front(); pp;
+         pp = elements_.nextof(*pp)) {
+        pp->stop();
```
gavv: I suggest to rework this method and add a guarantee that it will work even if `pp->stop()` will invoke `Transceiver::close()`. We can do it like this:

```
void Transceiver::stop_all_elements_() {
    core::SharedPtr<IElement> curr = elements_.front();
    while (curr) {
        core::SharedPtr<IElement> next = elements_.nextof(*curr);
        curr->stop();
        curr = next;
    }
}
```

src/tests/roc_lib/test_sender_receiver.cpp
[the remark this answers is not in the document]
gavv: We probably want a separate issue/PR for this.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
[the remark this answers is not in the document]
dshil: > The case:
> 
>     1. thread1: call trx::trx()
> 
>     2. thread2: call trx.stop() which sets stopped_ to true
> 
>     3. thread1: call trx::~trx()
> 
> 
> All three steps are performed sequentially (no concurrent trx calls).
> 
> The value set in `2.` (thread 2) is not guaranteed to be visible in `3.` (thread 1) because `3.` has no memory barrier (mutex or atomic).

Well, it seems as a bug in a user code.
As `thread1` creates an object it should ensure that it will be valid in all other threads. If `thread2` sleeps a little more than `trx::~trx()` in `thread1` we will get `heap use after free`.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Reworked

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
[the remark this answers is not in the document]
gavv: All three steps are performed sequentially. There is no concurrent trx calls. Every step is executed only after the previous step is finished. Thus ~trx() in thread1 is called only after stop() returned in thread2. And even if there is no concurrency, the result of a completed operation A from thread2 is not guaranteed to be seen by thread1 if there is no memory barrier (due to per-CPU core cache synchronization).

src/modules/roc_netio/target_uv/roc_netio/udp_receiver.cpp
dshil: Probably, it was done in a such non-obvious way because it is a little bit
strange to call stop() on non-started object.

dshil: Also, it was complicated because it was required to initialize UDPSender(Receiver)
address even if the start() was failed (due to stop logic that depends on the address).
Now we put an object directly to callback and do not have such problems.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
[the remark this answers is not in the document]
dshil: Even if `trx.stop()` is wait until background thread is finished?

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: One more time.

src/modules/roc_netio/target_uv/roc_netio/iclose_handler.h
dshil: Fixed

src/modules/roc_netio/target_uv/roc_netio/ielement.h
dshil: Done

src/modules/roc_netio/target_uv/roc_netio/ielement.h
dshil: Done

src/modules/roc_netio/target_uv/roc_netio/udp_receiver.cpp
dshil: Reworked

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: How did you notice it :) Fixed

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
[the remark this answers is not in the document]
gavv: Yes. Because an operation performed on one CPU core is not guaranteed to be visible on another CPU core until that another CPU core issues a memory barrier.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
gavv: It's a standard idiom: deleting an element inside a loop iterating the container.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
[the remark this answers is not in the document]
dshil: Ok, I got it.

src/lib/src/context.cpp
```diff
@@ -65,7 +63,6 @@ int roc_context_close(roc_context* context) {
     }
 
     context->trx.stop();
```
gavv: I guess we can also remove this call.

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
```diff
@@ -0,0 +1,57 @@
...
+//! @brief basic network port.
+
+#ifndef ROC_NETIO_BASIC_PORT_H_
+#define ROC_NETIO_BASIC_PORT_H_
+
+#include "roc_core/iallocator.h"
+#include "roc_core/list_node.h"
+#include "roc_core/refcnt.h"
+
+#include "roc_packet/address.h"
```
gavv: We usually use a single block (without newlines) for all includes for src/modules. We use separate blocks for external includes, src/modules includes, and src/tests includes.

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
```diff
@@ -0,0 +1,57 @@
...
+    //! Destroy.
+    virtual ~BasicPort();
+
+    //! Get bind address.
+    virtual const packet::Address& address() const = 0;
+
+    //! Open port.
+    //! @remarks
+    //!  Should be called from the event loop thread.
+    virtual bool open(packet::Address& bind_address) = 0;
```
gavv: It's worth to mention in the comment that open() modifies `@p bind_address` if the port is zero.

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
```diff
@@ -0,0 +1,57 @@
...
+namespace netio {
+
+//! Basic port interface.
+class BasicPort : public core::RefCnt<BasicPort>, public core::ListNode {
+public:
+    //! Initialize.
+    explicit BasicPort(core::IAllocator&);
+
+    //! Destroy.
+    virtual ~BasicPort();
```
gavv: We can mark ctor and dtor protected.

src/modules/roc_netio/target_uv/roc_netio/iclose_handler.h
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_netio/target_uv/roc_netio/iclose_handler.h
+//! @brief close handler.
```
gavv: Nitpick: we usually start a `@brief` with a capital letter.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -214,76 +205,82 @@ void Transceiver::stop_sem_cb_(uv_async_t* handle) {
...
-        if (int err = uv_async_send(&task_sem_)) {
-            roc_panic("transceiver: uv_async_send(): [%s] %s", uv_err_name(err),
-                      uv_strerror(err));
-        }
+    if (int err = uv_async_send(&task_sem_)) {
+        roc_panic("transceiver: uv_async_send(): [%s] %s", uv_err_name(err),
+                  uv_strerror(err));
     }
 
     while (!task.done) {
```
gavv: I see no place when `done` is set. How is it working? Why tests are not failing?

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
```diff
@@ -0,0 +1,57 @@
...
+
+    //! Open port.
+    //! @remarks
+    //!  Should be called from the event loop thread.
+    virtual bool open(packet::Address& bind_address) = 0;
+
+    //! Asynchronous close.
+    //! @remarks
+    //!  Should be called from the event loop thread.
+    virtual void async_close() = 0;
```
gavv: It's worth to mention in the comment that it is allowed to call async_close() multiple times (Transceiver can call async_close() in stop() for ports that are already being closed). In this case, the port should guarantee that handle_close() will be invoked only a single time (Transceiver will panic otherwise).

BTW, is this guarantee provided? Please re-check.

gavv: If possible, it would be great to add a test for this.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -214,76 +205,82 @@ void Transceiver::stop_sem_cb_(uv_async_t* handle) {
...
-    for (core::SharedPtr<UDPSender> sp = senders_.front(); sp;
-         sp = senders_.nextof(*sp)) {
-        sp->stop();
+    if (int err = uv_async_send(&stop_sem_)) {
+        roc_panic("transceiver: uv_async_send(): [%s] %s", uv_err_name(err),
+                  uv_strerror(err));
     }
+
+    wait_stopped_();
+    stopped_ = true;
```
gavv: It seems that there is a problem if stop() will be called to times concurrently:

* thread 1 acquires the mutex
* thread 2 sleeps on the mutex
* thread 1 calls uv_async_send() and calls wait_stopped_(), which temporary releases the mutex and goes to sleep
* thread 2 wakes up, acquires the mutex, and calls uv_async_send()

This way, uv_async_send() is called twice. But stop_sem_cb_() is not intended to be called more than once. For example, close_sems_() will incorrectly call uv_close() more than once.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -185,14 +158,32 @@ void Transceiver::remove_port(packet::Address bind_address) {
...
+        if (pp.get() == &port) {
+            ports_.remove(*pp);
+            cond_.broadcast();
+
+            return;
+        }
     }
 
+    roc_panic("transceiver: can't remove unknown port %s",
+              packet::address_to_str(port.address()).c_str());
```
gavv: Since all this stuff with multiple async_close() and handle_closed() calls is quite complicated, I would remove this panic and just allow multiple handle_closed() calls.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
gavv: BTW, why at all we need the wait_stopped_() call? We anyway call Thread::join() after calling stop_() so it looks unnecessary. If we remove wait_stopped_() the problem will gone.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -303,112 +300,118 @@ void Transceiver::process_tasks_() {
 }
```
gavv: I suggest to inline task->execute() call (i.e. remove execute() method). I've spent about ten minutes to find out why we are waiting `task.done` to become true but are never setting it :)

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -303,112 +300,118 @@ void Transceiver::process_tasks_() {
...
 bool Transceiver::add_udp_receiver_(Task& task) {
-    if (stopped_) {
-        roc_log(LogError, "transceiver: can't add port %s: transceiver is stopped",
-                packet::address_to_str(*task.address).c_str());
-        return false;
-    }
+    roc_panic_if(stopped_);
 
-    if (has_port_(*task.address)) {
+    if (get_port_(*task.address)) {
```
gavv: Do we need this check? Can we instead rely on OS to handle duplicate addresses?

* If we don't need it, we can remove it.
* If we need it, then 1) could you explain why? 2) we have a problem with current implementation.

The problem.

Step 1:

* the user calls add_udp_receiver("0.0.0.0:12345")
* trx creates UDPReceiverPort and adds it to ports_ list
* trx calls UDPReceiverPort::open("0.0.0.0:12345") but it fails for some reason and returns before setting UDPReceiverPort::address_
* trx calls UDPReceiverPort::async_close()
* the asynchronous close is still in progress...

Step 2:

* the user calls add_udp_receiver("0.0.0.0:0") (zero port means random port)
* trx calls get_port_("0.0.0.0:0") and it returns the UDPReceiverPort constructed at step 1 (because its address is not initialized and so is also "0.0.0.0:0")
* trx incorrectly complains that the port already exists and returns the user an error

If can't just remove the check, we can:

* ensure that address_ is already initialized (as in your previous revision)
* OR (it's probably a cleaner approach) place ports that are uninitialized/being closed into a separate list

gavv: We probably have this check because Transceiver::remove_port() assumes that there is always only one port for given address.

This approach will fail when we'll add TCP (because Address is IP + port, but it has no info about protocol - TCP or UDP).

So it seems that we anyway should rework the remove_port() interface somehow. If this is the only reason why we have this check, we can just do it now.

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
dshil: We can't:

```cpp
src/modules/roc_core/iallocator.h:38:17: error: '~BasicPort' is a protected member of 'roc::netio::BasicPort'
        object.~T();
                ^
src/modules/roc_netio/target_uv/roc_netio/basic_port.cpp:22:16: note: in instantiation of function template specialization 'roc::core::IAllocator::destroy<roc::netio::BasicPort>' requested
      here
    allocator_.destroy(*this);
               ^
src/modules/roc_netio/target_uv/roc_netio/basic_port.h:45:13: note: declared protected here
    virtual ~BasicPort();
```

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: >  why at all we need the wait_stopped_() call

It is an ancestor of the time when `trx.stop()` did not wait the background thread to be finished.

src/lib/src/context.cpp
dshil: Agree.

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
dshil: Fixed

src/modules/roc_netio/target_uv/roc_netio/iclose_handler.h
dshil: Fixed

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Reworked

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: I do not agree. Why not to go to task->execute method and understand what it does?
I also do not like the idea of in-lining because we hide the ugliness of calling a pointer to member function inside this function.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: > Do we need this check? Can we instead rely on OS to handle duplicate addresses?

Yes, we can. This check was introduced before this PR, thus I was trying to
keep the same behaviour.

> trx calls get_port_("0.0.0.0:0") and it returns the UDPReceiverPort
> constructed at step 1 (because its address is not initialized and so is also "0.0.0.0:0")

It is another problem, isn't it? I would expect that zero value for address
means invalid address, thus port also should have an invalid value.

> This approach will fail when we'll add TCP (because Address is IP + port, but it has no info about protocol - TCP or UDP).

This approach won't fail if we add an info about an address wheather it TCP or UDP,
thus the following check will find a correct address:

```cpp
core::SharedPtr<BasicPort> Transceiver::get_port_(const packet::Address& address) const {
    for (core::SharedPtr<BasicPort> pp = ports_.front(); pp; pp = ports_.nextof(*pp)) {
        if (pp->address() == address) {
            return pp;
        }
    }

    return NULL;
}
```

Now we can do the following:

- rely on OS to handle duplicate address
- leave remove_port() implementation as is and rework address structure later

src/modules/roc_netio/target_uv/roc_netio/udp_sender_port.cpp
```diff
@@ -15,45 +15,47 @@
...
+    if (int err = uv_async_init(&loop_, &write_sem_, write_sem_cb_)) {
+        roc_log(LogError, "udp sender: uv_async_init(): [%s] %s", uv_err_name(err),
                 uv_strerror(err));
         return false;
     }
 
-    handle_.data = this;
-    handle_initialized_ = true;
+    write_sem_.data = this;
+    write_sem_initialized_ = true;
```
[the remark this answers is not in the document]
dshil: Rolled back. The initial idea was to unify the initialization pipeline in udp_receiver and udp_sender.

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
dshil: done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Fixed

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
dshil: > BTW, is this guarantee provided? Please re-check.

No, there is no such guarantee. I do not see the reason why we should add it.
Now, trx guarantees that `trx.stop()` will be called only once, thus, port will be closed only once and only in one place, otherwise we will need to add a logic for multiple stop to each port implementation.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Fixed

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
dshil: In both, udp_receiver_port, udp_sender_port we have the following code:

```cpp
if (!handle_initialized_) {
    close_handler_.handle_closed(*this);
    return;
}
```

as a result in trx we will have the following:

```cpp
void Transceiver::handle_closed(BasicPort& port) {

    for (core::SharedPtr<BasicPort> pp = ports_.front(); pp; pp = ports_.nextof(*pp)) {
        if (pp.get() != &port) {
            continue;
        }

        roc_log(LogDebug,
                "transceiver: asynchronous close finished: port %s",
                packet::address_to_str(port.address()).c_str());

        ports_.remove(*pp);
        cond_.broadcast();

        break;
    }
}
```

and now it is safe to call async_close multiple times but this safety is very
implicit and can be easily broken.

dshil: Also we have the following contract in iclose_handler:

```cpp
//! Handle asynchronously closed port.
//!
//! @remarks
//!  - After this call closed port should not be used.
//!  - Should be called from the event loop thread.
virtual void handle_closed(BasicPort&) = 0;
```

Well, closed port should not be used because it is usually will be destroyed (freed)

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
gavv: Because all other manipulations with Task fields are in Transceiver and are not encapsulated in Task, so I expected to see setting "done" in Transceiver and it was a surprise that this specific case is implemented somewhere else.

On the other hand, this is not very important, we can keep it as is too.

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
gavv: async_close() is called twice if trx calls it in add_udp_receiver_ / add_udp_sender_ / remove_port_, the close is still not finished and the port is still in the ports_ list, and the user calls stop().

gavv:
```
void UDPSenderPort::close_() {
    if (!handle_initialized_) {
        close_handler_.handle_closed(*this);
        return;
    }

    if (handle_initialized_ && !uv_is_closing((uv_handle_t*)&handle_)) {
        roc_log(LogInfo, "udp sender: closing port %s",
                packet::address_to_str(address_).c_str());

        uv_close((uv_handle_t*)&handle_, close_cb_);
    }

    if (write_sem_initialized_ && !uv_is_closing((uv_handle_t*)&write_sem_)) {
        uv_close((uv_handle_t*)&write_sem_, close_cb_);
    }
}
```

handle_closed() will be incorrectly closed if handle_initialized_ is false but write_sem_initialized_ is true (if semaphore handle was successfully initialized, but udp handle initialization was failed).

gavv: We have this piece:

```
    if (!handle_initialized_) {
        close_handler_.handle_closed(*this);
        return;
    }
```

both in UDPReceiverPort and UDPSenderPort.

It has three problems:

* incorrect handle_closed() close described above
* multiple (spurious) handle_closed() calls are possible; not very harmful, but a bit dirty
* it was quite hard for me to figure it out when handle_closed() is called and when it's not

I suggest to add an explicit boolean field closed_ and use it like this:

```
void UDPSenderPort::close_() {
    if (closed_) {
        return; // means that handle_closed() was already called
    }

    if (!handle_initialized_ && !write_sem_initialized_) {
        closed_ = true;
        close_handler_.handle_closed(*this);
        return;
    }

    if (handle_initialized_ && !uv_is_closing((uv_handle_t*)&handle_)) {
        uv_close((uv_handle_t*)&handle_, close_cb_);
    }

    if (write_sem_initialized_ && !uv_is_closing((uv_handle_t*)&write_sem_)) {
        uv_close((uv_handle_t*)&write_sem_, close_cb_);
    }
}

void UDPSenderPort::close_cb_(uv_handle_t* handle) {
    ...

    if (handle == (uv_handle_t*)&self.handle_) {
        self.handle_initialized_ = false;
    } else {
        self.write_sem_initialized_ = false;
    }

    if (self.handle_initialized_ || self.write_sem_initialized_) {
        return;
    }

    closed_ = true;
    self.close_handler_.handle_closed(self);
}
```

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Well, I inlined it.

src/modules/roc_netio/target_uv/roc_netio/basic_port.h
dshil: > handle_closed() will be incorrectly called if handle_initialized_ is false but write_sem_initialized_ is true (if semaphore handle was successfully initialized, but udp handle initialization was failed).

I remember why I rearrange the handle initialization with `write_sem_`. I want to provide a guarantee (very week) that if handle is not initialized all other sems are not initialized too. As a result I can rely on this check.

dshil: > We have this piece:
> 
> ```c++
>     if (!handle_initialized_) {
>         close_handler_.handle_closed(*this);
>         return;
>     }
> ```
> 
> both in UDPReceiverPort and UDPSenderPort.
> 
> It has three problems:
> 
>     * incorrect handle_closed() close described above
> 
>     * multiple (spurious) handle_closed() calls are possible; not very harmful, but a bit dirty
> 
>     * it was quite hard for me to figure it out when handle_closed() is called and when it's not
> 
> 
> I suggest to add an explicit boolean field closed_ and use it like this:
> 
> ```c++
> void UDPSenderPort::close_() {
>     if (closed_) {
>         return; // means that handle_closed() was already called
>     }
> 
>     if (!handle_initialized_ && !write_sem_initialized_) {
>         closed_ = true;
>         close_handler_.handle_closed(*this);
>         return;
>     }

What if handle was not initialized but write_sem was initialized?

dshil: `closed_` added

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
```diff
@@ -91,65 +96,57 @@ bool UDPReceiver::start(packet::Address& bind_address) {
...
-        roc_log(LogError, "udp receiver: uv_udp_recv_stop(): [%s] %s", uv_err_name(err),
-                uv_strerror(err));
+    if (recv_started_) {
+        if (int err = uv_udp_recv_stop(&handle_)) {
+            roc_log(LogError, "udp receiver: uv_udp_recv_stop(): [%s] %s",
+                    uv_err_name(err), uv_strerror(err));
+        }
     }
 
     uv_close((uv_handle_t*)&handle_, close_cb_);
```
gavv: I suggest to guard this with `if (!uv_is_closing(...))`. Just in case. The flow is quite complicated so I would be twice as careful here. Also, I would add `recv_started_ = false` to the above `if` which calls `uv_udp_recv_stop`.

With this two changes calling UDPReceiverPort::async_close() twice will not be harmful (even it's not intended to do so).

BTW note that UDPSenderPort::async_close() *has* such guards.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -51,92 +80,50 @@ Transceiver::Transceiver(packet::PacketPool& packet_pool,
...
 
 Transceiver::~Transceiver() {
-    if (joinable()) {
-        roc_panic("transceiver: thread is not joined before calling destructor");
+    if (stop_sem_initialized_) {
+        if (int err = uv_async_send(&stop_sem_)) {
+            roc_panic("transceiver: uv_async_send(): [%s] %s", uv_err_name(err),
+                      uv_strerror(err));
+        }
+    } else if (task_sem_initialized_) {
```
gavv: Let's replace `else if (task_sem_initialized_) {` with just `else {`:

* close_sems_() already has all necessary checks so we don't need to duplicate them;
* if more handles / semaphores will be added, the code will continue to work.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -51,92 +80,50 @@ Transceiver::Transceiver(packet::PacketPool& packet_pool,
...
+        close_sems_();
     }
 
-    if (num_ports_ != 0) {
-        remove_all_ports_();
+    if (started_) {
+        Thread::join();
+    } else {
```
gavv: And here we should replace `else {` with `else if (loop_initialized_) {`. Or, alternatively, we can rewrite this piece as:

```
if (loop_initialized_) {
    if (started_) {
        Thread::join();
    } else {
        // If the thread was never started we should manually run the loop to
        // wait all opened handles to be closed. Otherwise, uv_loop_close()
        // will fail with EBUSY.
        Transceiver::run(); // non-virtual call from dtor
    }

    
    if (int err = uv_loop_close(&loop_)) {
        roc_panic("transceiver: uv_loop_close(): [%s] %s", uv_err_name(err),
                  uv_strerror(err));
    }
}
```

dshil: What if loop was not initialized but stop_sem and task_sem were initialized?

gavv: It's not possible. To initialize any handle, we need an initialized loop.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -296,118 +284,114 @@ void Transceiver::process_tasks_() {
...
+    return true;
 }
 
-bool Transceiver::has_port_(const packet::Address& address) const {
-    for (core::SharedPtr<UDPReceiver> rp = receivers_.front(); rp;
-         rp = receivers_.nextof(*rp)) {
-        if (rp->address() == address) {
-            return true;
-        }
+void Transceiver::wait_active_port_removed_(const BasicPort& port) const {
```
gavv: We can merge wait_active_port_removed_() and wait_failed_port_removed_() into one method by adding a list argument to it. We can also inline has_port() into that method because it is the only place when it is called.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
gavv: Also, I think we should replace `if (stop_sem_initialized_) {` with `if (started_) {`. We should call uv_async_send() only if the thread was started; otherwise, we should close the semaphores in place.

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -296,118 +284,114 @@ void Transceiver::process_tasks_() {
...
-            return true;
-        }
+    core::SharedPtr<BasicPort> pp = get_port_(*task.address);
+    if (!pp) {
+        return false;
     }
 
-    return false;
+    task.port = pp.get();
+    pp->async_close();
```
gavv: I suggest one cosmetic improvement:

* rename active_ports_ to open_ports_
* rename failed_ports_ to closing_ports_
* here, in remove_port_, move the port from open_ports_ to closing_ports_

This approach will have the following benefits:

* wait_port_removed_() will always use closing_ports_; there will be no need to parametrize it with a list

* handle_closed() will always scan closing_ports_; there will be no need to parametrize try_remove_port() with a list and try two lists

* we will be 100% sure that async_close_ports_() will never try to close an already closed port

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
dshil: Done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
```diff
@@ -214,76 +200,63 @@ void Transceiver::stop_sem_cb_(uv_async_t* handle) {
...
+
+    self.cond_.broadcast();
 }
 
-void Transceiver::close_() {
-    if (task_sem_initialized_) {
-        uv_close((uv_handle_t*)&task_sem_, NULL);
-        task_sem_initialized_ = false;
-    }
+void Transceiver::async_close_ports_() {
```
gavv: async_close_ports_() should acquire a mutex because it modifies the lists.

Also, now, when it removes ports from open_ports_, it can be simplified a bit:

```
    while (core::SharedPtr<BasicPort> port = open_ports_.front()) {
        open_ports_.remove(*port);
        closing_ports_.push_back(*port);

        port->async_close();
    }
```

dshil: Done

src/modules/roc_netio/target_uv/roc_netio/transceiver.cpp
dshil: Done

dshil:
ping @gavv

src/tests/roc_lib/test_sender_receiver.cpp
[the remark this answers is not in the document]
gavv: Added to my TODO.

gavv:
LGTM!

Issue #15, opened 2015-12-15
IPv6 support


gavv:
Current implementation unconditionally uses IPv4:
- uses 4 bytes for IP address;
- calls ipv4 functions from libuv to parse and bind address.

We can add support for IPv6, activated if it's available at build time and if user provided an IPv6 address.

gavv:
Almost done, needs testing.

gavv:
How to test:

```
roc-send -vv -s rtp+rs8m:[::1]:10001 -r rs8m:[::1]:10002 -i ./file.wav
roc-recv -vv -s rtp+rs8m::10001 -r rs8m::1000
```

```
roc-send -vv -s rtp+rs8m:[::1]:10001 -r rs8m:[::1]:10002 -i ./file.wav
roc-recv -vv -s rtp+rs8m:[::1]:10001 -r rs8m:[::1]:1000
```

gavv:
Done.

PR #209, opened 2019-06-05
roc_packet: properly compare IPv6 addresses

dshil:
#15

### DIFF
--- a/src/modules/roc_packet/target_posix/roc_packet/address.cpp
+++ b/src/modules/roc_packet/target_posix/roc_packet/address.cpp
@@ -129,7 +129,8 @@ bool Address::operator==(const Address& other) const {
         break;
 
     case AF_INET6:
-        if (sa_.addr6.sin6_addr.s6_addr != other.sa_.addr6.sin6_addr.s6_addr) {
+        if (memcmp(sa_.addr6.sin6_addr.s6_addr, other.sa_.addr6.sin6_addr.s6_addr,
+                   sizeof(sa_.addr6.sin6_addr.s6_addr)) != 0) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


src/modules/roc_packet/target_posix/roc_packet/address.cpp
```diff
@@ -129,7 +129,8 @@ bool Address::operator==(const Address& other) const {
         break;
 
     case AF_INET6:
-        if (sa_.addr6.sin6_addr.s6_addr != other.sa_.addr6.sin6_addr.s6_addr) {
+        if (memcmp(sa_.addr6.sin6_addr.s6_addr,
+                   other.sa_.addr6.sin6_addr.s6_addr, sizeof(in6_addr)) != 0) {
```
gavv: There is a small problem here. `sa_.addr6.sin6_addr.s6_addr` is not of type in6_addr, `sa_.addr6.sin6_addr` is. We either need:

```
memcmp(sa_.addr6.sin6_addr.s6_addr, other.sa_.addr6.sin6_addr.s6_addr,
    sizeof(sa_.addr6.sin6_addr.s6_addr)
```

or:

```
memcmp(&sa_.addr6.sin6_addr, &other.sa_.addr6.sin6_addr,
    sizeof(in6_addr)
```

I think the first variant is preferred, because:

* it specified in POSIX that there a field s6_addr of type uint8_t[16], so it's ok to rely that it's present and that it's an array
* it's not specified in POSIX whether other auxiliary fields are present, and if they are, we likely don't want to include them into comparison

(see http://pubs.opengroup.org/onlinepubs/009695399/basedefs/netinet/in.h.html)

dshil: Fixed

PR #245, opened 2019-08-30
Fixes #241 Add git SHA to build version info

### LINKED ISSUE #241: Add git sha to version info
Currently Roc tools (roc-recv, roc-send, roc-conv) can report version. The version is read from the `version` file during build time and passed to the tools using a preprocessor.

It would be very useful to report git commit sha as well. This is particularly useful when there are several boxes with binaries built from different commits yet from the same version.

The commit sha can be retrieved by:

```
git rev-parse HEAD
```

However, I think it would be better to avoid using git executable and retrieve it manually. Otherwise we will add build-time dependency on git, which is not really necessary and is inconvenient e.g. when building on an embedded host. Example:

```
cat .git/$(cat .git/HEAD | awk '{print $2}')
```

We should retrieve the sha SConstruct, add a preprocessor definition with the sha to the build, and report it in tools.

malcops:
Took a stab at this.. let me know if it's what you had in mind. 

`paul@thinkpad ~/roc-dev (feature/malcops/git-sha-to-version-info) $ ./bin/x86_64-pc-linux-gnu/roc-conv --version
roc-conv 0.1.2-e09a148c7b`
`paul@thinkpad ~/roc-dev (feature/malcops/git-sha-to-version-info) $ ./bin/x86_64-pc-linux-gnu/roc-recv --version
roc-recv 0.1.2-e09a148c7b`
`paul@thinkpad ~/roc-dev (feature/malcops/git-sha-to-version-info) $ ./bin/x86_64-pc-linux-gnu/roc-conv --version
roc-conv 0.1.2-e09a148c7b`

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -548,6 +548,8 @@ build_dir = 'build/%s/%s' % (
 
 env['ROC_BINDIR'] = '#bin/%s' % host
 env['ROC_VERSION'] = open(env.File('#.version').path).read().strip()
+env['ROC_SHA'] = os.popen("cat .git/$(cat .git/HEAD | awk '{print $2}') | cut -b1-10").read().strip()
+env['ROC_VERSION_STR'] = env['ROC_VERSION'] + "-" + env['ROC_SHA']
 
 abi_version = '.'.join(env['ROC_VERSION'].split('.')[:2])
 
--- a/SConstruct
+++ b/SConstruct
@@ -548,8 +548,8 @@ build_dir = 'build/%s/%s' % (
 
 env['ROC_BINDIR'] = '#bin/%s' % host
 env['ROC_VERSION'] = open(env.File('#.version').path).read().strip()
-env['ROC_SHA'] = os.popen("cat .git/$(cat .git/HEAD | awk '{print $2}') | cut -b1-10").read().strip()
-env['ROC_VERSION_STR'] = env['ROC_VERSION'] + "-" + env['ROC_SHA']
+env['ROC_SHA'] = env.ParseGitHead()
+env['ROC_VERSION_STR'] = env['ROC_VERSION'] + '-' + env['ROC_SHA']
 
 abi_version = '.'.join(env['ROC_VERSION'].split('.')[:2])
 
--- a/site_scons/site_tools/roc/parsers.py
+++ b/site_scons/site_tools/roc/parsers.py
@@ -14,9 +14,13 @@ def ParseVersion(env, command):
 def ParseGitHead(env):
     with open('.git/HEAD') as hf:
         head_ref = hf.read().strip().split()[-1]
-        with open('.git/%s' % (head_ref)) as hrf:
-            sha = hrf.read().strip()[:10]
-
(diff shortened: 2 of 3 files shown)


SConstruct
```diff
@@ -548,6 +548,8 @@ build_dir = 'build/%s/%s' % (
 
 env['ROC_BINDIR'] = '#bin/%s' % host
 env['ROC_VERSION'] = open(env.File('#.version').path).read().strip()
+env['ROC_SHA'] = os.popen("cat .git/$(cat .git/HEAD | awk '{print $2}') | cut -b1-10").read().strip()
+env['ROC_VERSION_STR'] = env['ROC_VERSION'] + "-" + env['ROC_SHA']
```
dshil: @gavv what is the policy of using single quotes or doubles quotes for strings in python?

As I might see almost everywhere in scons you use double quotes for human-readable lines (such strings do not contain any formatting) and single quotes in all other places.

SConstruct
```diff
@@ -548,6 +548,8 @@ build_dir = 'build/%s/%s' % (
 
 env['ROC_BINDIR'] = '#bin/%s' % host
 env['ROC_VERSION'] = open(env.File('#.version').path).read().strip()
+env['ROC_SHA'] = os.popen("cat .git/$(cat .git/HEAD | awk '{print $2}') | cut -b1-10").read().strip()
```
gavv: Thanks for the PR! Yes, this is what I was thinking about.

I'd change the cat/awk/cut pipeline into a pure-Python code to avoid build-time dependency on these tools and to make it portable across non-unix platforms. We will support Windows some day and I usually try to avoid unix-specific code in our build scripts, to make this future porting easier. And also, IIRC, awk is not installed by default on some minimalistic Linux distros like Alpine.

(And if this can't be done in a one-liner, I'd extract it into a function in [parsers.py](https://github.com/roc-project/roc/blob/master/site_scons/site_tools/roc/parsers.py#L119), something like env.ParseGitHead).

SConstruct
gavv: No any special policy, I think. Just my personal habit.

SConstruct
malcops: Good thinking.. updated things to be straight Python with file opens, and extracted to a function in parsers.py. 

I'm not sure about the Travis failures..
On my machine, head_ref is equal to "refs/heads/feature/malcops/git-sha-to-version-info" and then opening ".git/refs/heads/feature/malcops/git-sha-to-version-info" provides the SHA.

gavv: That's because the HEAD is detached on Travis.

You can reproduce it like this:

```
$ git checkout HEAD^
...
$ git status
HEAD detached at 020f7b9
nothing to commit, working tree clean
$ cat .git/refs/heads/master
a16b575580bbe70f06cc998541a46770cabbd885
```

I think we should correctly handle both situations.

malcops: Thanks for the hint!

gavv:
LGTM. Thanks!

PR #247, opened 2019-09-08
Add multicast support

dshil:
#253

### DIFF
--- a/docs/sphinx/manuals/roc_recv.rst
+++ b/docs/sphinx/manuals/roc_recv.rst
@@ -19,8 +19,8 @@ Options
 -v, --verbose             Increase verbosity level (may be used multiple times)
 -o, --output=OUTPUT       Output file or device
 -d, --driver=DRIVER       Output driver
--s, --source=PORT         Source port triplet (may be used multiple times)
--r, --repair=PORT         Repair port triplet (may be used multiple times)
+-s, --source=PORT         Source port triplet
+-r, --repair=PORT         Repair port triplet
 --sess-latency=STRING     Session target latency, TIME units
 --min-latency=STRING      Session minimum latency, TIME units
 --max-latency=STRING      Session maximum latency, TIME units
@@ -82,8 +82,6 @@ For example:
 
 If FEC is enabled on sender, a pair of a source and repair ports should be used for communication between sender and receiver. If FEC is disabled, a single source port should be used instead.
 
-Receiver can listen on multiple source and repair ports of different protocols simultaneously. This allows multiple senders which use different protocols and FEC schemes to connect to a single receiver.
-
 Supported protocols for source ports:
 
 - rtp (bare RTP, no FEC scheme)
@@ -104,6 +102,12 @@ Time
 EXAMPLES
 ========
 
+Listen on one bare RTP port on all IPv4 interfaces:
+
+.. code::
+
+    $ roc-recv -vv -s rtp::10001
+
 Listen on two ports on all IPv4 interfaces (but not IPv6):
 
 .. code::
@@ -116,23 +120,17 @@ Listen on two ports on all IPv6 interfaces (but not IPv4):
 
     $ roc-recv -vv -s rtp+rs8m:[::]:10001 -r rs8m:[::]:10002
 
+Listen on two ports on all IPv4 interfaces (using LDPC scheme)
+
+.. code::
+
+    $ roc-recv -vv -s rtp+ldpc::10001 -r ldpc::10002
+
 Listen on two ports on a particular interface:
 
 .. code::
 
     $ roc-recv -vv -s rtp+rs8m:192.168.0.3:10001 -r rs8m:192.168.0.3:10002
-
-Listen on two Reed-Solomon ports, two LDPC ports, and one bare RTP port:
-
-.. code::
-
-    $ roc-recv -vv -s rtp+rs8m::10001 -r rs8m::10002 -s rtp+ldpc::10003 -r ldpc::10004 -s rtp::10005
-
-Listen on two ports on all IPv4 interfaces and on two ports on all IPv6 interfaces:
-
-.. code::
-
-    $ roc-recv -vv -s rtp+rs8m::10001 -r rs8m::10002 -s rtp+rs8m:[::]:10001 -r rs8m:[::]:10002
 
 Output to the default ALSA device:
 
--- a/docs/sphinx/manuals/roc_recv.rst
+++ b/docs/sphinx/manuals/roc_recv.rst
@@ -21,6 +21,7 @@ Options
 -d, --driver=DRIVER       Output driver
 -s, --source=PORT         Source port triplet
 -r, --repair=PORT         Repair port triplet
+--miface=IFACE_IP         IP address of the network interface on which to join the multicast group
 --sess-latency=STRING     Session target latency, TIME units
 --min-latency=STRING      Session minimum latency, TIME units
 --max-latency=STRING      Session maximum latency, TIME units
@@ -132,6 +133,24 @@ Listen on two ports on a particular interface:
 
     $ roc-recv -vv -s rtp+rs8m:192.168.0.3:10001 -r rs8m:192.168.0.3:10002
 
+Listen on two ports on a particular multicast address (but not join to a multicast group):
+
+.. code::
+
+    $ roc-recv -vv -s rtp+rs8m:225.1.2.3:10001 -r rs8m:225.1.2.3:10002
+
+Listen on two ports on a particular multicast address and join to a multicast group on all interfaces:
+
+.. code::
+
+    $ roc-recv -vv -s rtp+rs8m:225.1.2.3:10001 -r rs8m:225.1.2.3:10002 --miface 0.0.0.0
+
+Listen on two ports on a particular multicast address and join to a multicast group on a particular interface:
+
+.. code::
+
+    $ roc-recv -vv -s rtp+rs8m:225.1.2.3:10001 -r rs8m:225.1.2.3:10002 --miface 192.168.0.3
+
 Output to the default ALSA device:
 
 .. code::
@@ -180,6 +199,13 @@ Select resampler profile:
 .. code::
 
     $ roc-recv -vv -s rtp+rs8m::10001 -r rs8m::10002 --resampler-profile=high
+
+Multicast
+---------
+
+*IFACE_IP* should be an IP address of the network interface on which to join to the multicast group. It may be "0.0.0.0" (for IPv4) or "[::]" (for IPv6) to join the multicast group on all available interfaces.
+
+If *IFACE_IP* is not specified and multicast address is used, the user is responsible for joining the multicast group manually.
 
 SEE ALSO
 ========
--- a/src/lib/include/roc/address.h
+++ b/src/lib/include/roc/address.h
@@ -37,7 +37,7 @@ typedef enum roc_family {
 
 enum {
     /** Address struct size. */
-    ROC_ADDRESS_SIZE = 64
+    ROC_ADDRESS_SIZE = 256
 };
 
 /** Network address.
--- a/src/lib/src/address.cpp
+++ b/src/lib/src/address.cpp
@@ -53,13 +53,13 @@ int roc_address_init(roc_address* address, roc_family family, const char* ip, in
     packet::Address& pa = *new (address_payload(address)) packet::Address;
 
     if (family == ROC_AF_AUTO || family == ROC_AF_IPv4) {
-        if (pa.set_ipv4(ip, port)) {
+        if (pa.set_host_ipv4(ip, port)) {
             return 0;
         }
     }
 
     if (family == ROC_AF_AUTO || family == ROC_AF_IPv6) {
-        if (pa.set_ipv6(ip, port)) {
+        if (pa.set_host_ipv6(ip, port)) {
             return 0;
         }
     }
@@ -97,7 +97,7 @@ const char* roc_address_ip(const roc_address* address, char* buf, size_t bufsz) 
 
     const packet::Address& pa = get_address(address);
 
-    if (!pa.get_ip(buf, bufsz)) {
+    if (!pa.get_host(buf, bufsz)) {
         return NULL;
     }
 
--- a/src/modules/roc_netio/target_libuv/roc_netio/udp_receiver_port.cpp
+++ b/src/modules/roc_netio/target_libuv/roc_netio/udp_receiver_port.cpp
@@ -26,6 +26,7 @@ UDPReceiverPort::UDPReceiverPort(ICloseHandler& close_handler,
     , close_handler_(close_handler)
     , loop_(event_loop)
     , handle_initialized_(false)
+    , multicast_group_joined_(false)
     , recv_started_(false)
     , closed_(false)
     , address_(address)
@@ -89,6 +90,12 @@ bool UDPReceiverPort::open() {
         return false;
     }
 
+    if (address_.multicast() && address_.has_miface()) {
+        if (!join_multicast_group_()) {
+            return false;
+        }
+    }
+
     if (int err = uv_udp_recv_start(&handle_, alloc_cb_, recv_cb_)) {
         roc_log(LogError, "udp receiver: uv_udp_recv_start(): [%s] %s", uv_err_name(err),
                 uv_strerror(err));
@@ -125,6 +132,10 @@ void UDPReceiverPort::async_close() {
         }
 
         recv_started_ = false;
+    }
+
+    if (multicast_group_joined_) {
+        leave_multicast_group_();
     }
 
     if (!uv_is_closing((uv_handle_t*)&handle_)) {
@@ -261,6 +272,43 @@ void UDPReceiverPort::recv_cb_(uv_udp_t* handle,
     pp->set_data(core::Slice<uint8_t>(*bp, 0, (size_t)nread));
 
     self.writer_.write(pp);
+}
+
+bool UDPReceiverPort::join_multicast_group_() {
+    char host[packet::Address::MaxStrLen];
+    address_.get_host(host, sizeof(host));
+
+    char miface[packet::Address::MaxStrLen];
+    address_.get_miface(miface, sizeof(miface));
+
+    if (int err = uv_udp_set_membership(&handle_, host, miface, UV_JOIN_GROUP)) {
+        roc_log(LogError, "udp receiver: uv_udp_set_membership(): [%s] %s",
+                uv_err_name(err), uv_strerror(err));
+        return false;
+    }
+
+    roc_log(LogDebug, "udp receiver: joined multicast group for port %s",
+            packet::address_to_str(address_).c_str());
+
+    return (multicast_group_joined_ = true);
+}
+
+void UDPReceiverPort::leave_multicast_group_() {
+    multicast_group_joined_ = false;
+
+    char host[packet::Address::MaxStrLen];
+    address_.get_host(host, sizeof(host));
+
+    char miface[packet::Address::MaxStrLen];
+    address_.get_miface(miface, sizeof(miface));
+
+    if (int err = uv_udp_set_membership(&handle_, host, miface, UV_LEAVE_GROUP)) {
+        roc_log(LogError, "udp receiver: uv_udp_set_membership(): [%s] %s",
+                uv_err_name(err), uv_strerror(err));
+    }
+
+    roc_log(LogDebug, "udp receiver: left multicast group for port %s",
+            packet::address_to_str(address_).c_str());
 }
 
 } // namespace netio
--- a/src/modules/roc_packet/target_posix/roc_packet/address.cpp
+++ b/src/modules/roc_packet/target_posix/roc_packet/address.cpp
@@ -13,8 +13,10 @@
 namespace roc {
 namespace packet {
 
-Address::Address() {
+Address::Address()
+    : miface_family_(AF_UNSPEC) {
     memset(&sa_, 0, sizeof(sa_));
+    memset(&miface_, 0, sizeof(miface_));
 }
 
 bool Address::valid() const {
@@ -32,7 +34,7 @@ bool Address::set_saddr(const sockaddr* sa) {
     return true;
 }
 
-bool Address::set_ipv4(const char* ip_str, int port) {
+bool Address::set_host_ipv4(const char* ip_str, int port) {
     in_addr addr;
     if (inet_pton(AF_INET, ip_str, &addr) != 1) {
         return false;
@@ -45,7 +47,7 @@ bool Address::set_ipv4(const char* ip_str, int port) {
     return true;
 }
 
-bool Address::set_ipv6(const char* ip_str, int port) {
+bool Address::set_host_ipv6(const char* ip_str, int port) {
     in6_addr addr;
     if (inet_pton(AF_INET6, ip_str, &addr) != 1) {
         return false;
@@ -54,6 +56,30 @@ bool Address::set_ipv6(const char* ip_str, int port) {
     sa_.addr6.sin6_family = AF_INET6;
     sa_.addr6.sin6_addr = addr;
     sa_.addr6.sin6_port = htons(uint16_t(port));
+
+    return true;
+}
+
+bool Address::set_miface_ipv4(const char* iface) {
+    in_addr addr;
+
+    if (inet_pton(AF_INET, iface, &addr) != 1) {
+        return false;
+    }
+    miface_.addr4 = addr;
+    miface_family_ = AF_INET;
+
+    return true;
+}
+
+bool Address::set_miface_ipv6(const char* iface) {
+    in6_addr addr;
+
+    if (inet_pton(AF_INET6, iface, &addr) != 1) {
+        return false;
+    }
+    miface_.addr6 = addr;
+    miface_family_ = AF_INET6;
 
     return true;
 }
@@ -103,7 +129,11 @@ bool Address::multicast() const {
     }
 }
 
-bool Address::get_ip(char* buf, size_t bufsz) const {
+bool Address::has_miface() const {
+    return miface_family_ == AF_INET || miface_family_ == AF_INET6;
+}
+
+bool Address::get_host(char* buf, size_t bufsz) const {
     switch (family_()) {
     case AF_INET:
         if (!inet_ntop(AF_INET, &sa_.addr4.sin_addr, buf, (socklen_t)bufsz)) {
@@ -115,6 +145,29 @@ bool Address::get_ip(char* buf, size_t bufsz) const {
         if (!inet_ntop(AF_INET6, &sa_.addr6.sin6_addr, buf, (socklen_t)bufsz)) {
             return false;
         }
+        break;
+
+    default:
+        return false;
+    }
+
+    return true;
+}
+
+bool Address::get_miface(char* buf, size_t bufsz) const {
+    switch (miface_family_) {
+    case AF_INET:
+        if (!inet_ntop(AF_INET, &miface_.addr4, buf, (socklen_t)bufsz)) {
+            return false;
+        }
+
+        break;
+
+    case AF_INET6:
+        if (!inet_ntop(AF_INET6, &miface_.addr6, buf, (socklen_t)bufsz)) {
+            return false;
+        }
+
         break;
 
     default:
@@ -148,6 +201,30 @@ bool Address::operator==(const Address& other) const {
         if (sa_.addr6.sin6_port != other.sa_.addr6.sin6_port) {
             return false;
         }
+        break;
+
+    default:
+        break;
+    }
+
+    if (miface_family_ != other.miface_family_) {
+        return false;
+    }
+
+    switch (miface_family_) {
+    case AF_INET:
+        if (miface_.addr4.s_addr != other.miface_.addr4.s_addr) {
+            return false;
+        }
+
+        break;
+    case AF_INET6:
+        if (memcmp(miface_.addr6.s6_addr, other.miface_.addr6.s6_addr,
+                   sizeof(miface_.addr6.s6_addr))
+            != 0) {
+            return false;
+        }
+
         break;
 
     default:
--- a/src/modules/roc_packet/target_posix/roc_packet/address.h
+++ b/src/modules/roc_packet/target_posix/roc_packet/address.h
@@ -32,11 +32,17 @@ public:
     //! Set address from sockaddr struct.
     bool set_saddr(const sockaddr* sa);
 
-    //! Set IPv4 address.
-    bool set_ipv4(const char* ip, int port);
+    //! Set IPv4 host address.
+    bool set_host_ipv4(const char* ip, int port);
 
-    //! Set IPv6 address.
-    bool set_ipv6(const char* ip, int port);
+    //! Set IPv6 host address.
+    bool set_host_ipv6(const char* ip, int port);
+
+    //! Set IPv4 address of the interface on which to join to the multicast group.
+    bool set_miface_ipv4(const char* iface);
+
+    //! Set IPv6 address of the interface on which to join to the multicast group.
+    bool set_miface_ipv6(const char* iface);
 
     //! Get sockaddr struct.
     sockaddr* saddr();
@@ -56,14 +62,26 @@ public:
     //! Check whether this is multicast address.
     bool multicast() const;
 
-    //! Get IP address.
-    bool get_ip(char* buf, size_t bufsz) const;
+    //! Check whether IP address of the interface on which to join to the
+    //! multicast group is initialized.
+    bool has_miface() const;
+
+    //! Get host IP address.
+    bool get_host(char* buf, size_t bufsz) const;
+
+    //! Get IP address of the interface on which to join to the multicast group.
+    bool get_miface(char* buf, size_t bufsz) const;
 
     //! Compare addresses.
     bool operator==(const Address& other) const;
 
     //! Compare addresses.
     bool operator!=(const Address& other) const;
+
+    enum {
+        // An estimate maximum length of a string representation of an address.
+        MaxStrLen = 128
+    };
 
 private:
     static socklen_t sizeof_(sa_family_t family);
@@ -74,6 +92,13 @@ private:
         sockaddr_in addr4;
         sockaddr_in6 addr6;
     } sa_;
+
+    union {
+        in_addr addr4;
+        in6_addr addr6;
+    } miface_;
+
+    sa_family_t miface_family_;
 };
 
 } // namespace packet
--- a/src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
+++ b/src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
@@ -19,13 +19,42 @@ address_to_str::address_to_str(const Address& addr) {
 
     switch (addr.version()) {
     case 4: {
-        if (!addr.get_ip(buffer_, sizeof(buffer_))) {
+        if (!addr.get_host(buffer_, sizeof(buffer_))) {
             roc_log(LogError, "address to str: can't format ip");
+            strcpy(buffer_, "<error>");
+
+            return;
         }
 
-        const size_t blen = strlen(buffer_);
+        size_t blen = strlen(buffer_);
+
         if (snprintf(buffer_ + blen, sizeof(buffer_) - blen, ":%d", addr.port()) < 0) {
             roc_log(LogError, "address to str: can't format port");
+            strcpy(buffer_, "<error>");
+
+            return;
+        }
+
+        if (!addr.has_miface()) {
+            return;
+        }
+
+        blen = strlen(buffer_);
+
+        if (snprintf(buffer_ + blen, sizeof(buffer_) - blen, " miface ") < 0) {
+            roc_log(LogError, "address to str: can't format miface");
+            strcpy(buffer_, "<error>");
+
+            return;
+        }
+
+        blen = strlen(buffer_);
+
+        if (!addr.get_miface(buffer_ + blen, sizeof(buffer_) - blen)) {
+            roc_log(LogError, "address to str: can't format miface");
+            strcpy(buffer_, "<error>");
+
+            return;
         }
 
         break;
@@ -33,19 +62,56 @@ address_to_str::address_to_str(const Address& addr) {
     case 6: {
         buffer_[0] = '[';
 
-        if (!addr.get_ip(buffer_ + 1, sizeof(buffer_) - 1)) {
+        if (!addr.get_host(buffer_ + 1, sizeof(buffer_) - 1)) {
             roc_log(LogError, "address to str: can't format ip");
+            strcpy(buffer_, "<error>");
+
+            return;
         }
 
-        const size_t blen = strlen(buffer_);
+        size_t blen = strlen(buffer_);
         if (snprintf(buffer_ + blen, sizeof(buffer_) - blen, "]:%d", addr.port()) < 0) {
             roc_log(LogError, "address to str: can't format port");
+            strcpy(buffer_, "<error>");
+
+            return;
+        }
+
+        if (!addr.has_miface()) {
+            return;
+        }
+
+        blen = strlen(buffer_);
+
+        if (snprintf(buffer_ + blen, sizeof(buffer_) - blen, " miface [") < 0) {
+            roc_log(LogError, "address to str: can't format miface");
+            strcpy(buffer_, "<error>");
+
+            return;
+        }
+
+        blen = strlen(buffer_);
+
+        if (!addr.get_miface(buffer_ + blen, sizeof(buffer_) - blen)) {
+            roc_log(LogError, "address to str: can't format miface");
+            strcpy(buffer_, "<error>");
+
+            return;
+        }
+
+        blen = strlen(buffer_);
+
+        if (snprintf(buffer_ + blen, sizeof(buffer_) - blen, "]") < 0) {
+            roc_log(LogError, "address to str: can't format miface");
+            strcpy(buffer_, "<error>");
+
+            return;
         }
 
         break;
     }
     default:
-        strcpy(buffer_, "none");
+        strcpy(buffer_, "<none>");
         break;
     }
 }
--- a/src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_port.cpp
+++ b/src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_port.cpp
@@ -121,7 +121,7 @@ bool parse_port(PortType type, const char* input, PortConfig& result) {
             return false;
         }
 
-        char addr6[128] = {};
+        char addr6[packet::Address::MaxStrLen] = {};
         if (addrlen - 2 > sizeof(addr6) - 1) {
             roc_log(LogError, "parse port: bad IPv6 address: address too long");
             return false;
@@ -129,12 +129,12 @@ bool parse_port(PortType type, const char* input, PortConfig& result) {
 
         memcpy(addr6, addr + 1, addrlen - 2);
 
-        if (!result.address.set_ipv6(addr6, (int)port_num)) {
+        if (!result.address.set_host_ipv6(addr6, (int)port_num)) {
             roc_log(LogError, "parse port: bad IPv6 address: %s", addr6);
             return false;
         }
     } else {
-        if (!result.address.set_ipv4(addr, (int)port_num)) {
+        if (!result.address.set_host_ipv4(addr, (int)port_num)) {
             roc_log(LogError, "parse port: bad IPv4 address: %s", addr);
             return false;
         }
--- a/src/tools/roc_recv/cmdline.ggo
+++ b/src/tools/roc_recv/cmdline.ggo
@@ -12,11 +12,9 @@ section "Options"
 
     option "driver" d "Output driver" typestr="DRIVER" string optional
 
-    option "source" s "Source port triplet (may be used multiple times)"
-        typestr="PORT" string optional multiple
+    option "source" s "Source port triplet" typestr="PORT" string optional
 
-    option "repair" r "Repair port triplet (may be used multiple times)"
-        typestr="PORT" string optional multiple
+    option "repair" r "Repair port triplet" typestr="PORT" string optional
 
     option "sess-latency" - "Session target latency, TIME units"
         string optional
--- a/src/tools/roc_recv/cmdline.ggo
+++ b/src/tools/roc_recv/cmdline.ggo
@@ -16,6 +16,9 @@ section "Options"
 
     option "repair" r "Repair port triplet" typestr="PORT" string optional
 
+    option "miface" - "IP address of the network interface on which to join the multicast group"
+        typestr="IFACE_IP" string optional
+
     option "sess-latency" - "Session target latency, TIME units"
         string optional
 
@@ -77,6 +80,9 @@ DRIVER is the type of the output file or device, e.g.:
 
 PORT is a triplet PROTOCOL:IPADDR:PORTNUM, e.g.:
   rtp+rs8m::10001; rtp+rs8m:127.0.0.1:10001; rtp+rs8m:[::1]:10001;
+
+IFACE_IP is an IPv4 or IPv6 address, e.g.:
+  0.0.0.0; 192.168.0.3; [::], [2001:db8::1];
 
 TIME is an integer number with a suffix, e.g.:
   123ns; 123us; 123ms; 123s; 123m; 123h;
--- a/src/tools/roc_recv/main.cpp
+++ b/src/tools/roc_recv/main.cpp
@@ -263,39 +263,44 @@ int main(int argc, char** argv) {
         return 1;
     }
 
-    if (args.source_given == 0 && args.repair_given == 0) {
-        roc_log(LogError, "at least one --source or --repair port should be specified");
+    if (!args.source_given) {
+        roc_log(LogError, "source port must be specified");
         return 1;
     }
 
-    for (size_t n = 0; n < args.source_given; n++) {
+    if (args.source_given) {
         pipeline::PortConfig port;
-        if (!pipeline::parse_port(pipeline::Port_AudioSource, args.source_arg[n], port)) {
-            roc_log(LogError, "can't parse source port: %s", args.source_arg[n]);
+
+        if (!pipeline::parse_port(pipeline::Port_AudioSource, args.source_arg,
+                                  port)) {
+            roc_log(LogError, "can't parse source port: %s", args.source_arg);
             return 1;
         }
+
         if (!trx.add_udp_receiver(port.address, receiver)) {
-            roc_log(LogError, "can't bind source port: %s", args.source_arg[n]);
+            roc_log(LogError, "can't bind source port: %s", args.source_arg);
             return 1;
         }
         if (!receiver.add_port(port)) {
-            roc_log(LogError, "can't initialize source port: %s", args.source_arg[n]);
+            roc_log(LogError, "can't initialize source port: %s", args.source_arg);
             return 1;
         }
     }
 
-    for (size_t n = 0; n < args.repair_given; n++) {
+    if (args.repair_given) {
         pipeline::PortConfig port;
-        if (!pipeline::parse_port(pipeline::Port_AudioRepair, args.repair_arg[n], port)) {
-            roc_log(LogError, "can't parse repair port: %s", args.repair_arg[n]);
+
+        if (!pipeline::parse_port(pipeline::Port_AudioRepair, args.repair_arg,
+                                  port)) {
+            roc_log(LogError, "can't parse repair port: %s", args.repair_arg);
             return 1;
         }
         if (!trx.add_udp_receiver(port.address, receiver)) {
-            roc_log(LogError, "can't bind repair port: %s", args.repair_arg[n]);
+            roc_log(LogError, "can't bind repair port: %s", args.repair_arg);
             return 1;
         }
         if (!receiver.add_port(port)) {
-            roc_log(LogError, "can't initialize repair port: %s", args.repair_arg[n]);
+            roc_log(LogError, "can't initialize repair port: %s", args.repair_arg);
             return 1;
         }
     }
--- a/src/tools/roc_recv/main.cpp
+++ b/src/tools/roc_recv/main.cpp
@@ -16,6 +16,8 @@
 #include "roc_core/scoped_destructor.h"
 #include "roc_core/unique_ptr.h"
 #include "roc_netio/transceiver.h"
+#include "roc_packet/address_to_str.h"
+#include "roc_packet/parse_address.h"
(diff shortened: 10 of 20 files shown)


src/modules/roc_packet/target_stdio/roc_packet/ip_str_formatter.h
```diff
@@ -0,0 +1,39 @@
...
+#include "roc_packet/address.h"
+
+namespace roc {
+namespace packet {
+
+//! Format ip string of network address.
+class ip_str_formatter : public core::NonCopyable<> {
+public:
+    //! Construct.
+    explicit ip_str_formatter(const Address&);
```
gavv: It would be better to:

* rename this to `ip_to_str`, like we already have `address_to_str`, `errno_to_str`, etc, etc

* reuse `ip_to_str` in `address_to_str`, like we reuse `address_to_str` in `port_to_str`

    https://github.com/roc-project/roc/blob/3fe1421cf6c7f8ee8399e824259a894e2b2e5eda/src/modules/roc_pipeline/target_stdio/roc_pipeline/port_to_str.cpp#L45-L53

gavv:
It would be nice if you explain these changes in the commit messages.

gavv:
A few questions:

* What happens if a single process joins the same group multiple times?
* What happens if multiple processes join the same group?
* What happens if the process crashes without leaving a group?

gavv:
Do you have examples of other software that automatically join/leave multicast group if the address is multicast? Is it a common practice?

gavv:
May the user want to control the interface on which to join the group?

dshil:
> Do you have examples of other software that automatically join/leave multicast group if the address is multicast? 

vlc

> Is it a common practice?

I guess yes since I saw this in 2 projects (vlc is one of them)

dshil:
> May the user want to control the interface on which to join the group?

Well, there are 2 most common use cases of multicast:

* Multiple roc-recv consume data on a multicast address and probably writing to a file or somewhere else
* One roc-send streams data using the multicast address and multiple clients consume this data.

I am not sure if the user even want to care about join/leave mechanism. All the user wants to do is send data over the network.

dshil:
> A few questions:
> 
>     * What happens if a single process joins the same group multiple times?
> 

10:15:03.324 [err] roc_netio: udp receiver: uv_udp_set_membership(): [EADDRINUSE] address already in use

>     * What happens if multiple processes join the same group?
> 

If more than one process joins the same multicast group on the same interface, they will all receive the datagrams sent to that group via that interface. The host remains member of the group, until all the processes decide to leave the group, a.k.a. ref counting mechanism. The funny thing that if we leave the multicast group but keep binding to the socket and there is at least one process that already has joined to the same multicast group we will still receive multicast traffic.

>     * What happens if the process crashes without leaving a group?

It depends how process handle the signals and abnormal exit. I tried to kill roc via SIGKILL and after that check if multicast group still exists with `netstat -gn`. The group was released. 

We can even omit the leave_multicast_group() call and when process finishes the group will be released (if it is not used by other processes).

gavv:
> 10:15:03.324 [err] roc_netio: udp receiver: uv_udp_set_membership(): [EADDRINUSE] address already in use

How is it reproduced?

I just checked out your branch and run this command:

```
roc-recv -vv -s rtp+rs8m:224.0.0.1:10001 -r rs8m:224.0.0.1:10002 -s rtp+ldpc:224.0.0.1:10003 -r ldpc:224.0.0.1:10004
```

The receiver joined the group four times and I got no error.

gavv:
> If more than one process joins the same multicast group on the same interface, they will all receive the datagrams sent to that group via that interface. The host remains member of the group, until all the processes decide to leave the group, a.k.a. ref counting mechanism. The funny thing that if we leave the multicast group but keep binding to the socket and there is at least one process that already has joined to the same multicast group we will still receive multicast traffic.

Great.

> I tried to kill roc via SIGKILL and after that check if multicast group still exists with netstat -gn. The group was released.

Great.

> We can even omit the leave_multicast_group() call and when process finishes the group will be released (if it is not used by other processes).

We shouldn't: the user may want to unbind port without terminating the process. We don't have unbind() in the API so far, but we'll add it later.

gavv:
> I am not sure if the user even want to care about join/leave mechanism. All the user wants to do is send data over the network.

I think the user may want to control the interface for security reasons. Imagine that the user has two interfaces: one "public", connected to the Internet, and one "private", connected to the private network. Then the user starts Roc receiver and wants to ensure that Roc will accept only packets from the private network and will be protected (at the kernel level) from the Internet.

In case of unicast, the user just specifies the address of the private interface as the bind address, and it works as expected.

But in case of multicast, if I understand correctly, your implementation will automatically join the group on all interfaces and in our example it will be vulnerable to packets coming from the public interface? Is it true? If you're not sure, could you please test it?

If it's true, I think we need a way to restrict the interface. BTW, it seems that VLC has `--miface` option for that.

We could also add such an option, but it would be better to be able to control the interface on per-port basic, so that you can bind some ports to one interface, and some ports to another.

To solve this, we could implement a syntax like:

```
roc-recv -vv -s rtp+rs8m:224.0.0.1@eth0:10001 -r rs8m:224.0.0.1@eth0:10002
```

It should be also possible to specify interface address instead of interface name, like:

```
roc-recv -vv -s rtp+rs8m:224.0.0.1@192.168.0.101:10001 -r rs8m:224.0.0.1@192.168.0.101:10002
```

In case of unicast, the `@...` part should be omitted. In case of multicast, the `@...` part likely should be strictly required for security reasons.

It should be also possible to explicitly request to join group on all interfaces, like:

```
rtp+rs8m:224.0.0.1@0.0.0.0:10001
```

Finally, if the entire IP address is empty:

```
rtp+rs8m::10001
```

...which means that we should bind to all interfaces, we should join the group on all interfaces too.

dshil:
> > 10:15:03.324 [err] roc_netio: udp receiver: uv_udp_set_membership(): [EADDRINUSE] address already in use
> 
> How is it reproduced?
> 
> I just checked out your branch and run this command:
> 
> ```
> roc-recv -vv -s rtp+rs8m:224.0.0.1:10001 -r rs8m:224.0.0.1:10002 -s rtp+ldpc:224.0.0.1:10003 -r ldpc:224.0.0.1:10004
> ```
> 
> The receiver joined the group four times and I got no error.

I was wrong. I got this error because I tried to join to the multicast group on the same interface + on the same port.

> What happens if a single process joins the same group multiple times?

I guess that nothing special. We will only see one record in `/proc/net/igmp`

dshil:
> We shouldn't: the user may want to unbind port without terminating the process. We don't have unbind() in the API so far, but we'll add it later.

OK. As I wrote previously in case of multicast group we will have a some kind UB for the user who is not well-informed about multicast usage. If you unbind the port but still there is at least one process joined to this group all previously joined processes will still receive the multicast traffic.

I guess that providing the `unbind` API is not enough. Lets keep in mind it.

dshil:
> > I am not sure if the user even want to care about join/leave mechanism. All the user wants to do is send data over the network.
>
> I think the user may want to control the interface for security reasons. Imagine that the user has two interfaces: one "public", connected to the Internet, and one "private", connected to the private network. Then the user starts Roc receiver and wants to ensure that Roc will accept only packets from the private network and will be protected (at the kernel level) from the Internet.
>
> In case of unicast, the user just specifies the address of the private interface as the bind address, and it works as expected.
>
> But in case of multicast, if I understand correctly, your implementation will automatically join the group on all interfaces and in our example it will be vulnerable to packets coming from the public interface? Is it true? If you're not sure, could you please test it?

You are right.

I join to the multicast group as:

```cpp
bool UDPReceiverPort::join_multicast_group_() {
    if (int err = uv_udp_set_membership(
            &handle_, packet::ip_to_str(address_).c_str(), NULL, UV_JOIN_GROUP)) {
        roc_log(LogError, "udp receiver: uv_udp_set_membership(): [%s] %s",
                uv_err_name(err), uv_strerror(err));
        return false;
    }

    return (multicast_group_joined_ = true);
}
```

And if we look into the libuv code we see the following:

```cpp
UV_EXTERN int uv_udp_set_membership(uv_udp_t* handle,
                                    const char* multicast_addr,
                                    const char* interface_addr,
                                    uv_membership membership);

static int uv__udp_set_membership4(uv_udp_t* handle,
                                   const struct sockaddr_in* multicast_addr,
                                   const char* interface_addr,
                                   uv_membership membership) {
  struct ip_mreq mreq;
  int optname;
  int err;

  memset(&mreq, 0, sizeof mreq);

  if (interface_addr) {
    err = uv_inet_pton(AF_INET, interface_addr, &mreq.imr_interface.s_addr);
    if (err)
      return err;
  } else {
    mreq.imr_interface.s_addr = htonl(INADDR_ANY);
  }
  
  // ...
}
```

As you might see we join to the multicast group on all available interfaces.

dshil:
> It would be nice if you explain these changes in the commit messages. It would be also nice to add the PR number to the commit title (since we have no github issue for this).

I extracted it into the separate issue.

src/modules/roc_packet/target_stdio/roc_packet/ip_str_formatter.h
dshil: Reworked

gavv:
> > We shouldn't: the user may want to unbind port without terminating the process. We don't have unbind() in the API so far, but we'll add it later.
> 
> OK. As I wrote previously, in case of multicast group we will have a some kind UB for the user who is not well-informed about multicast usage. If you unbind the port but still there is at least one process joined to this group all previously joined processes will still receive the multicast traffic (even after `unbind` call).
> 
> I guess that providing the `unbind` API is not enough. Lets keep in mind it.

I don't get it. If our unbind() implementation will leave the group and then close the socket, what problem exactly do we have?

gavv:
> > What happens if a single process joins the same group multiple times?
> 
> I guess that nothing special. We will only see one record in `/proc/net/igmp`

Is reference counting working as expected inside a single process too? I mean if a single process joins the same group two times, it should leave it two times, not a single time, right?

src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
```diff
@@ -18,28 +19,12 @@ address_to_str::address_to_str(const Address& addr) {
...
-
-        const size_t blen = strlen(buffer_);
-        if (snprintf(buffer_ + blen, sizeof(buffer_) - blen, ":%d", addr.port()) < 0) {
-            roc_log(LogError, "address to str: can't format port");
-        }
-
-        break;
-    }
+    case 4:
     case 6: {
```
gavv: We can replace this switch with a single check `if (addr.valid())` to avoid duplicating this code.

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
```diff
@@ -263,5 +275,29 @@ void UDPReceiverPort::recv_cb_(uv_udp_t* handle,
     self.writer_.write(pp);
 }
 
+bool UDPReceiverPort::join_multicast_group_() {
+    if (int err = uv_udp_set_membership(&handle_, packet::ip_to_str(address_).c_str(),
```
gavv: I suggest to add a debug message that we're joining/leaving group - it may be an important event for the user.

src/modules/roc_packet/target_stdio/roc_packet/ip_to_str.cpp
```diff
@@ -0,0 +1,47 @@
...
+        if (!addr.get_ip(buffer_, sizeof(buffer_))) {
+            roc_log(LogError, "ip_str_formatter: can't format ip");
+        }
+
+        break;
+    }
+    case 6: {
+        buffer_[0] = '[';
+
+        if (!addr.get_ip(buffer_ + 1, sizeof(buffer_) - 1)) {
```
gavv: Oh, by the way shouldn't it be `sizeof(buffer_) - 2` instead of `sizeof(buffer_) - 1`? Because of the closing `]`.

dshil:
> > > We shouldn't: the user may want to unbind port without terminating the process. We don't have unbind() in the API so far, but we'll add it later.
> > 
> > 
> > OK. As I wrote previously, in case of multicast group we will have a some kind UB for the user who is not well-informed about multicast usage. If you unbind the port but still there is at least one process joined to this group all previously joined processes will still receive the multicast traffic (even after `unbind` call).
> > I guess that providing the `unbind` API is not enough. Lets keep in mind it.
> 
> I don't get it. If our unbind() implementation will leave the group and then close the socket, what problem exactly do we have?

I missed that you mentioned that `unbind` will also leave a group. I didn't assume it (don't know why). Never mind/

gavv:
> I missed that you mentioned that unbind will also leave a group. I didn't assume it (don't know why). Never mind.

Actually unbind() should close and remove the port. It's an opposite of adding port. It's called unbind only because "adding port" is called "bind" in terms of our C API.

src/modules/roc_packet/target_stdio/roc_packet/ip_to_str.cpp
dshil: Previously we specified that we have `sizeof(buffer) - 1` bytes and expected that 2snprintf` would properly format the remaining part.

Agree, that in my implementation in some very rare cases we won't have enough space in the underlying buffer for the remaining `]`.

dshil: Also, it does not a big deal but for the ip part we can use a smaller buffer because the string representation of the ipv6 address does not exceed 64 bytes. On Linux we have `INET6_ADDRSTRLEN`.

src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
dshil: fixed

src/modules/roc_packet/target_stdio/roc_packet/ip_to_str.cpp
gavv: > On Linux we have INET6_ADDRSTRLEN.

We should avoid using it here because this will make ip_to_str platform-dependent.

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
dshil: Done

dshil:
> > > What happens if a single process joins the same group multiple times?
> > 
> > 
> > I guess that nothing special. We will only see one record in `/proc/net/igmp`
> 
> Is reference counting working as expected inside a single process too? I mean if a single process joins the same group two times, it should leave it two times, not a single time, right?

Correct

SConstruct
```diff
@@ -1091,6 +1091,7 @@ if compiler == 'clang':
             '-Wno-variadic-macros',
             '-Wno-disabled-macro-expansion',
             '-Wno-c++11-long-long',
+            '-Wno-documentation-unknown-command',
```
gavv: Why do we need it? These warnings are usually helpful.

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
```diff
@@ -263,5 +275,39 @@ void UDPReceiverPort::recv_cb_(uv_udp_t* handle,
     self.writer_.write(pp);
 }
 
+bool UDPReceiverPort::join_multicast_group_() {
+    if (!address_.miface()) {
```
gavv: So the real condition is `address_.multicast() && address_.miface()`. I'd keep these two together. Maybe move all checks inside join_multicast_group_()/leave_multicast_group_()?

src/modules/roc_packet/target_posix/roc_packet/address.h
```diff
@@ -56,6 +59,13 @@ class Address {
     //! Check whether this is multicast address.
     bool multicast() const;
 
+    //! Return multicast interface address.
+    //!
+    //! @returns
+    //!  NULL if address is not multicast
+    //!  NULL if automatic multicast group joining/leaving is disabled
```
gavv: > NULL if address is not multicast

We don't have this check in code.

I think it'd better to replace both lines with a single "NULL if multicast interface is not set".

src/tests/roc_pipeline/target_stdio/test_port.cpp
```diff
@@ -186,5 +186,42 @@ TEST(port, bad_port_number) {
     CHECK(!parse_port(Port_AudioSource, "rtp:1.2.3.4:999999999999999", port));
 }
 
+TEST(port, multicast_ipv4) {
```
gavv: We also need a check that if address is multicast, but the `@...` part is missing, the parsing fails. Both for ipv4 and ipv6.

gavv: We also need to check that both interface name and IP address can be used after `@`.

src/lib/include/roc/address.h
```diff
@@ -85,6 +85,19 @@ typedef struct roc_address {
 ROC_API int
 roc_address_init(roc_address* address, roc_family family, const char* ip, int port);
 
+/** Set multicast interface address.
+ *
+ * @b Parameters
+ *  - @p address should point to a properly initialized address struct
+ *  - @p miface should point to a zero-terminated string with a valid IPv4 or
+ *       IPv6 multicast address
```
gavv: ...or network interface name?

src/lib/include/roc/address.h
```diff
@@ -126,6 +139,18 @@ ROC_API const char* roc_address_ip(const roc_address* address, char* buf, size_t
...
+ *
+ * @b Parameters
+ *  - @p address should point to a properly initialized address struct
+ *
+ * @b Returns
+ *  - returns zero-terminated string representation of the multicast address
+ *  - returns NULL if the argument is invalid
+ *  - returns NULL if the multicast address was not set
+ */
+ROC_API const char* roc_address_miface(const roc_address* address);
```
gavv: This API assumes that roc_address stores the the raw string somewhere. This is true now, but may change in future (e.g. we may want to store a parsed address or a some interface handle instead). So I suggest to remove this assumption from the API and instead copy the miface to the user-provided buffer, like we do it for IP address:

https://github.com/roc-project/roc/blob/a29d8b5edf4f29922c506f5f27fe12dbd752c171/src/lib/include/roc/address.h#L129

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
```diff
@@ -263,5 +275,39 @@ void UDPReceiverPort::recv_cb_(uv_udp_t* handle,
...
+
+    if (int err = uv_udp_set_membership(&handle_, packet::ip_to_str(address_).c_str(),
+                                        address_.miface(), UV_JOIN_GROUP)) {
+        roc_log(LogError, "udp receiver: uv_udp_set_membership(): [%s] %s",
+                uv_err_name(err), uv_strerror(err));
+        return false;
+    }
+
+    roc_log(LogDebug, "udp receiver: join multicast group: address=%s interface=%s",
+            packet::address_to_str(address_).c_str(), address_.miface());
```
gavv: Does address_to_str() also formats the `@...` part? If it does we don't need to explicitly print interface here. If it doesn't, I guess we should fix it?

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
```diff
@@ -263,5 +275,39 @@ void UDPReceiverPort::recv_cb_(uv_udp_t* handle,
     self.writer_.write(pp);
 }
 
+bool UDPReceiverPort::join_multicast_group_() {
+    if (!address_.miface()) {
+        return true;
+    }
+
+    if (int err = uv_udp_set_membership(&handle_, packet::ip_to_str(address_).c_str(),
```
gavv: Do I understand correctly that uv_udp_set_membership() accepts both interface name and IP address? Is this behavior portable accross platforms?

src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_port.cpp
```diff
@@ -47,97 +47,140 @@ bool match_proto(PortType type, const char* str, PortProtocol& proto) {
...
-                "parse port: bad format: expected PROTO:ADDR:PORT or PROTO::PORT");
+    PortProtocol protocol = Proto_None;
+    if (!match_proto(type, proto_buf, protocol)) {
+        return Proto_None;
+    }
+
+    return protocol;
+}
+
+long parse_port_num(const char* port) {
```
gavv: s/long/int/

The returned port is guaranteed to be in the correct range.

src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_port.cpp
```diff
@@ -47,97 +47,140 @@ bool match_proto(PortType type, const char* str, PortProtocol& proto) {
...
+    return true;
+}
 
-    char proto_buf[16] = {};
-    if (size_t(lcolon - input) > sizeof(proto_buf) - 1) {
-        roc_log(LogError, "parse port: bad protocol: too long");
+bool parse_ipv6_addr(const char* addr, long port_num, PortConfig& result) {
+    const size_t addrlen = strlen(addr);
+
+    if (addr[addrlen - 1] != ']') {
```
gavv: It would be better to work with `[` and `]` in the same function instead of spreading it across several. I suggest to work with brackets in parse_addr() and pass begin/end pointers to parse_ipv6_addr().

gavv: Actually, to unify all these helpers, I would suggest to pass begin/end pointers to *every* helper. Currently some of them accept begin/end pointers, and some accept null-terminated strings, which is a bit confusing.

src/lib/include/roc/address.h
```diff
@@ -85,6 +85,19 @@ typedef struct roc_address {
...
+ * @b Parameters
+ *  - @p address should point to a properly initialized address struct
+ *  - @p miface should point to a zero-terminated string with a valid IPv4 or
+ *       IPv6 multicast address
+ *
+ * @b Returns
+ *  - returns zero if no error occurred
+ *  - returns a negative value if the arguments are invalid
+ */
+ROC_API int roc_address_set_miface(roc_address* address, const char* miface);
```
gavv: Let's be explicit and rename it to `roc_address_set_multicast_interface`.

SConstruct
dshil: When I use `@INTERFACE` in the comments clang complains about unknown tag.

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
dshil: It accepts only interface address. I will convert interface name to interface address during port parsing.

SConstruct
gavv: Then you should try to escape or quote it.

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
dshil: I would prefer to keep the both checks on the upper layer. Having unconditional `join_multicast_group_()` seems a little bit odd.

```cpp
    if (address_.multicast() && address_.miface()) {
        if (!join_multicast_group_()) {
            return false;
        }
    }
```

gavv: Also good.

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
gavv: I've created #258. Take a look at the description of two new functions: `parse_endpoint_url()` and `resolve_address()`.

That task should be done separately from this PR, but the idea is that the resolving should be separated from parsing (at lease because resolving should stay in netio and depends on the network engine we use, and parsing is independent from that).

So for now, in this PR, I suggest to choose one of the two options:

* either don't perform interface name resolving at all and leave it for #258

* or create a simplified version of `netio::Transceiver::resolve_address()` that resolves interface name but doesn't resolve hostname; and we will finish it in #258

SConstruct
dshil: When I use `@INTERFACE` in the comments clang complains about unknown tag.

gavv: `\@INTERFACE`

src/lib/include/roc/address.h
dshil: As discussed privately supporting interface name is not a good idea because:

* interface name can be mapped to the multiple addresses and who knows which one should be returned

* interface name can be mapped to the ivp4 or ipv6 versions of the same address and which version should be used is the next question

To make things simple we've decided to prohibit interface name usage.

SConstruct
dshil: Wow, it's a kind of magic :)

src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_port.cpp
dshil: well, internally we are working with the port as a long number why should we return int and explicitly cast long to int?

SConstruct
dshil: Fixed

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
dshil: > Does address_to_str() also formats the @... part?

It does not. I fixed it. I think we should use `ip@iface` instead of `ip@iface:port` here.

src/lib/include/roc/address.h
dshil: Done

src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_port.cpp
gavv: We're casting port to int anyway. I'm just suggesting to make the cast earlier because in parse_port_num() we already checked that the port is in [0; 2^16) so the cast is valid.

src/modules/roc_netio/target_uv/roc_netio/udp_receiver_port.cpp
gavv: Agree.

gavv: But I'd expand the log message to make things explicit, something like:

```
udp receiver: joined multicast group %s at interface %s
```

src/modules/roc_packet/target_posix/roc_packet/address.h
dshil: > I think it'd better to replace both lines with a single "NULL if multicast interface is not set".

Done

src/lib/include/roc/address.h
dshil: Fixed

src/tests/roc_pipeline/target_stdio/test_port.cpp
dshil: > We also need to check that both interface name and IP address can be used after @.

https://github.com/roc-project/roc/pull/247#discussion_r326856574

src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_port.cpp
dshil: Fixed

dshil:
As discussed privately we won't introduce per port multicast interface configuration due to 4fc2b5723e7ad28b331514cf50f86c055606d51c.

New option was added to `roc-recv`, `--miface`, that allows to configure multicast interface for both source and repair ports:

* It is prohibited (in API and in cmdline interface) to combine IPv4 address and  IPv6 multicast interface address.

```
roc-recv -s rtp:225.1.2.3:10001 --miface ffee::
```

* It is prohibited (in API and in cmdline interface) to combine IPv6 address and IPv4 multicast interface address:

```
roc-recv -s rtp:[ffee::]:10001 --miface 0.0.0.0
```

* It is prohibited (in API and in cmdline interface) to use `--miface` option for non-multicast address:

```
roc-recv -s rtp:192.169.18.1:10001 --miface 0.0.0.0
```

* If --miface option is not set, user will require to manually join to a multicast group.

dshil:
Due to #258 we need to discuss how we will merge to address to string formatting. Now I use the old variants of network address format as:

```
PROTO:ADDR@MIFACE:PORT
```

dshil:
@gavv,

Could you please test ipv6 multicast?

* Start receiver:

```
roc-recv -s rtp+rs8m:\[ff00::\]:10001 -r rs8m:\[ff00::\]:10002 --miface \:\: -vv
```

* Ensure that we successfully joined to the multicast group:

```
16:42:28.475 [dbg] roc_netio: udp receiver: joined multicast group ff00:: at interface ::
16:42:28.475 [inf] roc_netio: udp receiver: opened port [ff00::]@[::]:10001
16:42:28.475 [inf] roc_pipeline: receiver: adding port rtp+rs8m:[ff00::]@[::]:10001
16:42:28.475 [dbg] roc_netio: udp receiver: joined multicast group ff00:: at interface ::
16:42:28.475 [inf] roc_netio: udp receiver: opened port [ff00::]@[::]:10002
16:42:28.475 [inf] roc_pipeline: receiver: adding port rs8m:[ff00::]@[::]:10002
```

* You can also validate output from `netstat -gn`:

```
arch% netstat -gn
IPv6/IPv4 Group Memberships
Interface       RefCnt Group
--------------- ------ ---------------------
wlp2s0          2      ff00::
```

* Start sender

```
roc-send -s rtp+rs8m:\[ff00::\]:10001 -r rs8m:\[ff00::\]:10002 -i loituma.wav -vv
```

And I get the silence...

docs/sphinx/manuals/roc_recv.rst
```diff
@@ -122,17 +118,23 @@ Listen on two ports on a particular interface:
...
     $ roc-recv -vv -s rtp+rs8m:192.168.0.3:10001 -r rs8m:192.168.0.3:10002
 
-Listen on two Reed-Solomon ports, two LDPC ports, and one bare RTP port:
+Listen on two ports on a particular multicast interface (but not join to a multicast group):
+
+.. code::
+
+    $ roc-recv -vv -s rtp+rs8m:225.1.2.3:10001 -r rs8m:225.1.2.3:10002
+
+Listen on two ports on a particular multicast interface and join to a multicast group on all interfaces:
```
gavv: AFAIK there is no such thing as "multicast interface".

`multicast interface` -> `multicast address`

docs/sphinx/manuals/roc_recv.rst
```diff
@@ -122,17 +118,23 @@ Listen on two ports on a particular interface:
...
-    $ roc-recv -vv -s rtp+rs8m::10001 -r rs8m::10002 -s rtp+ldpc::10003 -r ldpc::10004 -s rtp::10005
+    $ roc-recv -vv -s rtp+rs8m:225.1.2.3:10001 -r rs8m:225.1.2.3:10002 --miface 0.0.0.0
 
-Listen on two ports on all IPv4 interfaces and on two ports on all IPv6 interfaces:
+Listen on two ports on a particular multicast interface and join to a multicast group on a particular interface:
 
 .. code::
 
-    $ roc-recv -vv -s rtp+rs8m::10001 -r rs8m::10002 -s rtp+rs8m:[::]:10001 -r rs8m:[::]:10002
+    $ roc-recv -vv -s rtp+rs8m:225.1.2.3:10001 -r rs8m:225.1.2.3:10002 --miface 192.185.12.3
```
gavv: I suggest to use `192.168.0.3` in line with other examples :)

docs/sphinx/manuals/roc_send.rst
```diff
@@ -73,8 +73,6 @@ For example:
 - rtp+rs8m:127.0.0.1:10001
 - rtp+rs8m:[::1]:10001
 
-If FEC is enabled on sender, a pair of a source and repair ports should be used for communication between sender and receiver. If FEC is disabled, a single source port should be used instead.
-
```
gavv: Why remove this?

src/lib/include/roc/address.h
```diff
@@ -85,6 +85,22 @@ typedef struct roc_address {
 ROC_API int
 roc_address_init(roc_address* address, roc_family family, const char* ip, int port);
 
+/** Set multicast interface address.
+ *
+ * @b Parameters
+ *  - @p address should point to a properly initialized address struct
+ *  - @p family should be @c ROC_AF_AUTO, @c ROC_AF_IPv4, or @c ROC_AF_IPv6
+ *  - @p iface should point to a zero-terminated string with a valid IPv4 or
+ *    IPv6 multicast address
```
gavv: `IPv4 or IPv6 multicast address` -> `IPv4 or IPv6 address`

Also suggest to rename `iface` to `iface_ip`.

And also please add some remarks, something like:

```
@remarks
  This function assigns a multicast interface to an address. If the address
  has a multicast interface, roc_receiver_bind() will use it to join the
  multicast group on that interface. Otherwise, roc_receiver_bind() will not
  join the multicast group and the user should do it manually in order to be
  able to receive multicast packets.

  @p iface_ip should be an IP address of the network interface on which to
  join to the multicast group. It may be "0.0.0.0" or "[::]" to join
  the multicast group on all available interfaces.
```

src/lib/include/roc/address.h
```diff
@@ -85,6 +85,22 @@ typedef struct roc_address {
...
+ *
+ * @b Returns
```
gavv: `@b Returns` -> `@returns` ?

src/lib/src/address.cpp
```diff
@@ -67,6 +67,46 @@ int roc_address_init(roc_address* address, roc_family family, const char* ip, in
...
+    packet::Address& pa = get_address(address);
+
+    if (!pa.multicast()) {
+        return -1;
+    }
+
+    if (family == ROC_AF_AUTO || family == ROC_AF_IPv4) {
+        if (pa.set_multicast_iface_v4(iface)) {
+            if (pa.version() == 6) {
+                return -1;
```
gavv: This leaves the address in inconsistent state. If we can't proceed, it's better to keep the address unmodified.

src/modules/roc_packet/target_posix/roc_packet/address.h
```diff
@@ -56,9 +62,15 @@ class Address {
     //! Check whether this is multicast address.
     bool multicast() const;
 
+    //! Check whether multicast interface address was initialized.
+    bool has_multicast_iface() const;
+
     //! Get IP address.
     bool get_ip(char* buf, size_t bufsz) const;
```
gavv: I suggest to rename `get_ip()`, `set_ip...()` etc to `get_host()`, `set_host...()` etc, to avoid confusion (because multicast interface is also an IP).

src/modules/roc_packet/target_posix/roc_packet/address.cpp
```diff
@@ -124,6 +154,29 @@ bool Address::get_ip(char* buf, size_t bufsz) const {
...
+            return false;
+        }
+
+        break;
+
+    default:
+        return false;
+    }
+
+    return true;
```
gavv: false?

src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
```diff
@@ -19,27 +21,36 @@ address_to_str::address_to_str(const Address& addr) {
...
-        if (!addr.get_ip(buffer_, sizeof(buffer_))) {
-            roc_log(LogError, "address to str: can't format ip");
-        }
-
-        const size_t blen = strlen(buffer_);
-        if (snprintf(buffer_ + blen, sizeof(buffer_) - blen, ":%d", addr.port()) < 0) {
-            roc_log(LogError, "address to str: can't format port");
+        if (addr.has_multicast_iface()) {
+            if (snprintf(buffer_, sizeof(buffer_), "%s@%s:%d", ip_to_str(addr).c_str(),
+                         multicast_iface_to_str(addr).c_str(), addr.port())
```
gavv: Since we didn't implement this notation, I suggest to use something more explicit like:

    <ip>:<port> miface <ip>

src/modules/roc_packet/target_stdio/roc_packet/ip_to_str.h
```diff
@@ -0,0 +1,37 @@
...
+#define ROC_PACKET_IP_TO_STR_H_
+
+#include "roc_core/noncopyable.h"
+#include "roc_packet/address.h"
+
+namespace roc {
+namespace packet {
+
+//! Format ip string of network address.
+class ip_to_str : public core::NonCopyable<> {
```
gavv: I think it's an overkill to create a class for every Address field (ip, miface), especially given they're used only in one place. I suggest to remove ip_to_str and multicast_iface_to_str at all and just use get_ip() and get_multicast_iface() directly in UDPReceiver.

src/modules/roc_packet/target_posix/roc_packet/address.h
gavv: I'd also rename `multicast_iface` to `miface` for brevity (but not in the public API).

src/tools/roc_recv/cmdline.ggo
```diff
@@ -12,11 +12,11 @@ section "Options"
 
     option "driver" d "Output driver" typestr="DRIVER" string optional
 
-    option "source" s "Source port triplet (may be used multiple times)"
-        typestr="PORT" string optional multiple
+    option "source" s "Source port triplet" typestr="PORT" string required
```
gavv: This will break `-L`

src/lib/include/roc/address.h
gavv: We should also explain this in the manual page for roc-recv.

src/tools/roc_recv/main.cpp
```diff
@@ -25,6 +26,50 @@
 
 using namespace roc;
 
+namespace {
+
+bool set_multicast_iface(packet::Address& address, const char* miface) {
```
gavv: This should be a method of Address I think.

gavv: Or maybe this should be `address::parse_miface`. Because we probably should expect `[]` brackets for IPv6 addresses and in this case we need some code to parse them.

gavv:
> Could you please test ipv6 multicast?

This works for me:

```
roc-send -vv -s rtp+rs8m:[ff03::1]:10001 -r rs8m:[ff03::1]:10002 -i stash/loituma.wav
roc-recv -vv -s rtp+rs8m:[ff03::1]:10001 -r rs8m:[ff03::1]:10002 --miface ::
```

`ff00::` is probably reserved. I don't have much experience with IPv6.

src/tools/roc_recv/main.cpp
gavv: Because this is a bit confusing:

```
roc-recv -vv -s rtp+rs8m:[ff03::1]:10001 -r rs8m:[ff03::1]:10002 --miface ::
```

I'd rather expect:

```
roc-recv -vv -s rtp+rs8m:[ff03::1]:10001 -r rs8m:[ff03::1]:10002 --miface [::]
```

src/modules/roc_packet/target_posix/roc_packet/address.cpp
dshil: It is the success branch. If something goes wrong we will return false.

docs/sphinx/manuals/roc_recv.rst
dshil: I wanted to keep things the same. We have:

```rst
Listen on two ports on a particular interface:

.. code::

    $ roc-recv -vv -s rtp+rs8m:192.168.0.3:10001 -r rs8m:192.168.0.3:10002
```

I considered that having an example in which we also *Listen on two ports on a particular interface* but this interface corresponds to the multicast address.

docs/sphinx/manuals/roc_recv.rst
dshil: Done

docs/sphinx/manuals/roc_recv.rst
```diff
@@ -122,17 +118,23 @@ Listen on two ports on a particular interface:
 
     $ roc-recv -vv -s rtp+rs8m:192.168.0.3:10001 -r rs8m:192.168.0.3:10002
 
-Listen on two Reed-Solomon ports, two LDPC ports, and one bare RTP port:
```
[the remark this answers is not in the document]
dshil: Fixed

docs/sphinx/manuals/roc_recv.rst
```diff
@@ -80,10 +80,6 @@ For example:
 - rtp+rs8m:127.0.0.1:10001
 - rtp+rs8m:[::1]:10001
 
-If FEC is enabled on sender, a pair of a source and repair ports should be used for communication between sender and receiver. If FEC is disabled, a single source port should be used instead.
-
-Receiver can listen on multiple source and repair ports of different protocols simultaneously. This allows multiple senders which use different protocols and FEC schemes to connect to a single receiver.
```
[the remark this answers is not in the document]
dshil: Correct, fixed.

docs/sphinx/manuals/roc_recv.rst
```diff
@@ -19,8 +19,8 @@ Options
 -v, --verbose             Increase verbosity level (may be used multiple times)
 -o, --output=OUTPUT       Output file or device
 -d, --driver=DRIVER       Output driver
--s, --source=PORT         Source port triplet (may be used multiple times)
--r, --repair=PORT         Repair port triplet (may be used multiple times)
+--s, --source=PORT        Source port triplet
+--r, --repair=PORT        Repair port triplet
```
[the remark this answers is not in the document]
dshil: Fixed

src/tools/roc_recv/cmdline.ggo
dshil: Could you please explain how and why?

docs/sphinx/manuals/roc_send.rst
dshil: Oops, rolled back.

src/tools/roc_recv/cmdline.ggo
gavv: Before:

```
$ roc-recv -L
supported device drivers:
  pulse alsa ossdsp oss

supported file drivers:
  aifc aiffc aiff aif al au snd avr cdda cdr cvsd cvs cvu dat dvms vms f32
  gsrt hcom htk ima la lu maud prc s8 s16 s24 s32 sf ircam sln smp sndr
  sox sph nist 8svx txw u8 sou fssd u16 u24 u32 ul voc vox wav wavpcm amb
  xa flac gsm lpc10 lpc mp3 mp2 sds caf fap mat4 mat mat5 paf pvf sd2 w64
  vorbis ogg
```

After:

```
$ roc-recv -L
roc-recv: '--source' ('-s') option required
```

src/lib/include/roc/address.h
dshil: Look at another `Returns`. I did the same.

src/lib/src/address.cpp
dshil: I did it because we have `ROC_AF_AUTO` and we actually don't know the exact family. We try to set multicast for a ipv4 address and if it fails we will try to set it to the ipv6 address.

From another side, we can remove `family` parameter from the argument list and say the we will set the multicast based on the network address (initialized with roc_address_init). This will simplify the API and the code.

src/lib/include/roc/address.h
gavv: Hm, maybe I did it because it's a list.

src/lib/src/address.cpp
dshil: Discussed to remove roc_address_set_multicast_interface and merge this functionality into roc_address_init.

src/modules/roc_packet/target_stdio/roc_packet/ip_to_str.h
dshil: Well, they are used at least in 2 places:

* udp_receiver_port - for logging
* udp_receiver_port - join/leave multicast group
* address_to_str - again for logging

As address_to_str may be used in many place I think it is worth to keep these helpers, probably we can use templates here to avoid adding helper for each `get*`.

gavv: I don't suggest to remove address_to_str. I suggest to remove ip_to_str and multicast_iface_to_str.

They are used only UDPReceiverPort. And we can easily avoid them by using get_ip() directly:

```
char ip[64];
address_.get_ip(ip, sizeof(ip));

char miface[64];
address_.get_miface(miface, sizeof(miface));

if (int err = uv_udp_set_membership(&handle_, ip,
                                    miface,
                                    UV_JOIN_GROUP)) {
    roc_log(LogError, "udp receiver: uv_udp_set_membership(): [%s] %s",
            uv_err_name(err), uv_strerror(err));
    return false;
}

roc_log(LogDebug, "udp receiver: joined multicast group %s at interface %s",
        ip,
        miface);
```

ip_to_str and multicast_iface_to_str are just convinient wrappers primarily useful for logging. Such wrappers make sense when there are many usages. If there are very few usages, I'd prefer to avoid adding them.

dshil: Done

src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
dshil: Done

src/lib/include/roc/address.h
dshil: Outdated

src/tools/roc_recv/cmdline.ggo
dshil: Fixed

src/tools/roc_recv/main.cpp
dshil: Done

dshil: The remaining thing to discuss. We have the following log record:

```
16:02:29.828 [dbg] roc_netio: udp receiver: joined multicast group ff00:: at interface ::
```

That seems a little bit non-consistent with our entire IPv6 address formatting. I think we can use something similar to:

```
16:02:29.828 [dbg] roc_netio: udp receiver: joined multicast group for port %s
```

and rely on the address_to_str formatting.

Thoughts?

gavv: > and rely on the address_to_str formatting.

Agree.

dshil: Done

src/lib/include/roc/address.h
gavv: > Outdated

Not quite. I think we still need to add an explanation to the manual page.

src/tools/roc_recv/main.cpp
```diff
@@ -263,39 +265,54 @@ int main(int argc, char** argv) {
         return 1;
     }
 
-    if (args.source_given == 0 && args.repair_given == 0) {
-        roc_log(LogError, "at least one --source or --repair port should be specified");
+    if (!args.source_given && !args.repair_given) {
+        roc_log(LogError, "at least --source or --repair port should be specified");
```
gavv: Actually it can be `if (!args.source_given) {`. We can't work without a source port.

src/modules/roc_netio/target_libuv/roc_netio/udp_receiver_port.cpp
```diff
@@ -263,5 +274,42 @@ void UDPReceiverPort::recv_cb_(uv_udp_t* handle,
     self.writer_.write(pp);
 }
 
+bool UDPReceiverPort::join_multicast_group_() {
+    char host[128];
+    address_.get_host(host, sizeof(host));
+
+    char miface[128];
+    address_.get_miface(miface, sizeof(miface));
```
gavv: Lets add and use corresponding constant to packet::Address, something like Address::MaxStrLen.

src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
```diff
@@ -19,29 +19,63 @@ address_to_str::address_to_str(const Address& addr) {
...
         }
 
-        const size_t blen = strlen(buffer_);
+        size_t blen = strlen(buffer_);
         if (snprintf(buffer_ + blen, sizeof(buffer_) - blen, ":%d", addr.port()) < 0) {
             roc_log(LogError, "address to str: can't format port");
         }
 
+        if (addr.has_miface()) {
+            blen = strlen(buffer_);
```
gavv: Here and in other places. I don't think we can call strlen() after snprintf() returned error because I'm not sure that the string is guaranteed to be null-terminated in this case.

In case of error, I suggest to return early and replace the entire buffer with something like `<error>`.

src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_address.cpp
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) 2015 Roc authors
```
gavv: * This code is not stdio-specific, so it should not be under `target_stdio`.
* This code is specific to packet::Address and is not specific to roc_pipeline, so it should be in `roc_packet`.
* s/2015/2019/ here and in `parse_address.h`.

src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
dshil: Done

src/modules/roc_netio/target_libuv/roc_netio/udp_receiver_port.cpp
dshil: Done

src/tools/roc_recv/main.cpp
dshil: Done

src/modules/roc_pipeline/target_stdio/roc_pipeline/parse_address.cpp
dshil: Done

src/lib/include/roc/address.h
dshil: I hope I understood you correctly, see https://github.com/roc-project/roc/blob/81baa7f6f90080b39e7fba604eb134d38575ce76/docs/sphinx/manuals/roc_recv.rst

dshil: It is also need to be added to the help page of `roc-recv`.

docs/sphinx/manuals/roc_recv.rst
```diff
@@ -38,6 +38,7 @@ Options
 -1, --oneshot             Exit when last connected client disconnects (default=off)
 --poisoning               Enable uninitialized memory poisoning (default=off)
 --beeping                 Enable beeping on packet loss  (default=off)
+--miface=IFACE_IP         IP address of the network interface on which to join to the multicast group
```
gavv: It's better to keep the same order of options in .rst and .ggo (and --help).

You can use this command to generate rst table:

    roc-recv --help | ./scripts/help2rst.py

docs/sphinx/manuals/roc_recv.rst
```diff
@@ -65,6 +66,13 @@ Driver
 
 If the driver is omitted, some default driver is selected. If the user did specify the output and it is a file with a known extension, the appropriate file driver is selected. Otherwise, the first device driver available on the system is selected.
 
+Multicast
+---------
```
gavv: Let's move this section after "Port" section.

docs/sphinx/manuals/roc_recv.rst
```diff
@@ -65,6 +66,13 @@ Driver
...
+
+*IFACE_IP* should be an IP address of the network interface on which to join to the multicast group. It may be "0.0.0.0" or "[::]" to join the multicast group on all available interfaces.
+
+If *IFACE_IP* is not specified and multicast address is used, user should manually join to the multicast group.
```
gavv: `to join TO the multicast group` -> `to join the multicast group`

I'd also clarify:

`"0.0.0.0" or "[::]"` -> `"0.0.0.0" (for IPv4) or "[::]" (for IPv6)`

`user should manually join TO the multicast group` -> `THE user is responsible for joining the multicast group manually`

src/tools/roc_recv/cmdline.ggo
```diff
@@ -80,6 +81,9 @@ DRIVER is the type of the output file or device, e.g.:
 PORT is a triplet PROTOCOL:IPADDR:PORTNUM, e.g.:
   rtp+rs8m::10001; rtp+rs8m:127.0.0.1:10001; rtp+rs8m:[::1]:10001;
 
+IFACE_IP is an IP address in one of the following formats:
+  0.0.0.0; 192.168.0.3; [::], [2001:db8::1];
```
gavv: `IFACE_IP is an IP address in one of the following formats:` -> `IFACE_IP is an IPv4 or IPv6 address, e.g.:`

(because the next line is not list of *formats*; it's a list of some examples)

src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
```diff
@@ -19,27 +19,93 @@ address_to_str::address_to_str(const Address& addr) {
 
     switch (addr.version()) {
     case 4: {
-        if (!addr.get_ip(buffer_, sizeof(buffer_))) {
+        if (!addr.get_host(buffer_, sizeof(buffer_))) {
             roc_log(LogError, "address to str: can't format ip");
+            strcpy(buffer_, "error");
```
gavv: Could you please replace `error` and `none` with `<error>` and `<none>`, similar to https://github.com/roc-project/roc/blob/develop/src/modules/roc_address/io_uri_to_str.cpp#L14-L18 ?

docs/sphinx/manuals/roc_recv.rst
dshil: Done

docs/sphinx/manuals/roc_recv.rst
dshil: Done

src/tools/roc_recv/cmdline.ggo
dshil: Agree. Fixed

src/modules/roc_packet/target_stdio/roc_packet/address_to_str.cpp
dshil: Done

docs/sphinx/manuals/roc_recv.rst
gavv: Now it's in after the EXAMPLES sections. And I meant after the Port section: https://github.com/roc-project/roc/blob/2aa69410aa32a782eab61c71f39a9128235cfb4f/docs/sphinx/manuals/roc_recv.rst#port

gavv: a29a82809240bdc12d8996d7112b882ff9fa0e9a

PR #249, opened 2019-09-10
Implement backtrace printing for non-glibc targets

hrishikeshSuresh:
**Problem Summary**
Implement backtrace printing for non-glibc targets (bionic and musl)<br>
**Solution**
In ` ./src/modules/roc_core/target_bionic/roc_core/backtrace.cpp`,
I have implemented captureBacktrace(), to find the size of backtrace stack. In captureBacktrace(), we call _Unwind_Backtrace() to perform stack unwinding through unwind data, using the unwindCallback() function. 
dumpBacktrace() will first check whether backtrace is available or not and then print the following in the format of <br>`#index: address_of_the_cursor function_name`.<br>Here, demangled name is printed if the demangling operation suceeds.
dumpBacktrace_fd() is same as dumpBacktrace() but prints the output to the file parameter passed.<br>
In `./src/modules/roc_core/target_musl/roc_core/backtrace.cpp'`,
backtrace() finds the size of backtrace stack by just iterating till the end of the stack. backtrace_symbols() does the same thing but we print the following in the format of <br>` #index : function_name address_of_cursor offset instruction_pointer`<br> to stderr. backtrace_symbols_fd() does the same but writes the output to the given file descriptor.

**Queries**
1 . I am printing the backtrace in two different ways for each backtrace.cpp files. Which one should I stick to? 
2 . I couldn't add the optional dependency on libunwind.h, so can any hints/ideas be given as to how this has to be done?


gavv:
Thanks! I'll take a look.

One thing I can tell right now:

print_backtrace_emergency() is called from a signal handler (e.g. when we receive SIGSEGV). This means that this function should be signal-safe, as mentioned here:

https://github.com/roc-project/roc/blob/faebd1e0059c16476f6aae8ee5a1cfc66c8fee7f/src/modules/roc_core/backtrace.h#L24-L28

You can read about signal-safety here: http://man7.org/linux/man-pages/man7/signal-safety.7.html

In particular, signal-safe function is allowed to call only other signal-safe functions. malloc and fprintf, and all other functions that work with `FILE*` or perform allocations, are not signal-safe, so we can't use them in print_backtrace_emergency().

Thus, print_backtrace_emergency() is intended to be a simplified version of print_backtrace() that can have limited functionality but can be called from a signal handler. For example, our implementation of print_backtrace_emergency() for glibc does not perform demangling and formatting at all, it just calls backtrace_symbols_fd() which is signal-safe.

libunwind documentation: (https://www.nongnu.org/libunwind/man/libunwind(3).html#section_5) says:

> The manual page for each libunwind routine identifies whether or not it is signal-safe, but as a general rule, any routine that may be needed for local unwinding is signal-safe (e.g., unw_step() for local unwinding is signal-safe). For remote-unwinding, none of the libunwind routines are guaranteed to be signal-safe.

So for libunwind, we should:

* check manual page for every libunwind function and ensure that we only perform local unwinding and only call its signal-safe functions

* in our print_backtrace_emergency() implementation we also should not call any functions from standard library that are not signal-safe, e.g. fprintf() and malloc()

On bionic, we also should find a signal-safe way to print backtrace, or we can leave print_backtrace_emergency() unimplemented for now.

gavv:
>  I couldn't add the optional dependency on libunwind.h, so can any hints/ideas be given as to how this has to be done?

This should be done in `SConstruct` file. Basically you can just duplicate what we're doing for sox and adjust it for libunwind, specifically:
  
* add --disable-libunwind (like --disable-sox)

* automatically append target_libunwind to ROC_TARGETS if we use musl and if --disable-libunwind is not specified

* check whether libunwind is available on system (like we do it in `if 'target_sox' in system_dependecies:` branch for sox)

* download and build libunwind if it's present in --build-3rdparty option (like we do it in `if 'target_sox' in download_dependencies:` branch for sox)

* add libunwind branch to our `3rdparty.py` script

If you will have any trouble, feel free to ask here or in the IRC chat, I'll help you with this.

gavv:
If you will not find a good way to print backtrace in a signal-safe manner on Bionic, there is another option. If libunwind works good on Anrdoid (could you check it please?) we can use it on Bionic too.

In this case we can have three backtrace implementations:

* target_glibc - selected if we're using glibc; prints backtrace using glibc functions

* target_libunwind - selected if we're using musl or bionic and libunwind was not disabled by user; prints backtrace using libunwind functions

* target_nobacktrace - selected if we're using musl or bionic and libunwind was disabled by user; no-op implementation

gavv:
> I am printing the backtrace in two different ways for each backtrace.cpp files. Which one should I stick to?

So the answer to this question is that we need both versions: print_backtrace() is full-featured version that can perform allocations, demangling, use fprintf, etc; and print_backtrace_emergency() is a limited version that can be called from a signal handler.

Since print_backtrace_emergency() can't use fprintf(), it should write to the stderr file descriptor directly, but it doesn't make sense to use fdopen() because it's not signal-safe.

hrishikeshSuresh:
I made print_backtrace_emergency() signal-safe by making use of write() instead of fprintf() in `target_musl/roc_core/backtrace.cpp` . Demangling operation for print_backtrace() was missed in the earlier commit for musl, so that has been added now.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
...
+/* Function to convert unsigned int to decimal/hex string.
+ * This function is signal safe.
+ * Parameters : 
+ * number = the number to be converted.
+ * base = 10 for decimal to char* and 16 for hex to char* conversion.
+ * output_string = the decimal/hex string
+ */
+void uint_to_string(unsigned int number, int base, char *output_string)
+{
+	int i = 12;
```
gavv: When the function is called with a 10-byte buffer:

```
		char index_buffer[10];
		uint_to_string(index, 10, index_buffer);
```

we're smashing the stack here (making a write of of the index_buffer bounds):

```
output_string[i] = "0123456789abcdef"[number % base];
```

Why you set it to 12 here? For 10 base, 4-byte unsigned integer can contain at most 10 digits, so the buffer size should be 11 (10 digits and `\0`) and `i` should be initialized to 10, isn't it?

I suggest to fix the function as follows:

* add `MaxDigits` constant and use it both inside the function and when calling it
* change `unsigned int` to `uint32_t` so that `MaxDigits` will be platform-independent
* also rename the function to `uint32_to_string`

```
enum { MaxDigits = 10 };

void uint32_to_string(uint32_t number, int base, char *output_string)
{
	int i = MaxDigits;
	int j = 0;
	do {
		output_string[i] = "0123456789abcdef"[number % base];
		i--;
		number = number/base;
	} while(number > 0);

	while(++i < MaxDigits + 1)
		output_string[j++] = output_string[i];
	output_string[j] = 0;
}

...

char index_buffer[MaxDigits+1];
uint32_to_string(somenum, 10, index_buffer);
```

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
...
+	/* To point to current frame.
+	 * unw_init_local() is signal safe. 
+	 */
+	unw_init_local(&cursor, &context);
+
+	/* Moving to previously called frames & going through each of them.
+	 * unw_step() is signal safe.
+	 * buffer = final output that will be written to stderr. 
+	 */
+	u_int index = 0;
```
gavv: AFAIK `u_int` is not standardized, so please use `unsigned int` and `uint32_t` here and below.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
...
+		(void) unw_get_proc_name(&cursor, b, sizeof(b), &offset);
+		
+		/* Perform demangling
+		 */
+		char *demangled;
+		int status = -1;
+		if(offset > 0) {
+			char mangled[128];
+			memcpy(mangled, b, 128);
+			demangled = abi::__cxa_demangle(mangled, 0, 0, &status);
```
gavv: abi::__cxa_demangle() is specific to libstdc++ (GNU Standard C++ Library). libunwind supports a variety of platforms, so it is possible that this code will be used without libstdc++. So we can not use this function here. Can libunwind perform demangling for us? If it can't, I think we can just don't perform demangling at all in the libunwind version.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
...
+		strcat(buffer, offset_buffer);
+		strcat(buffer, ") [");
+
+		char address_buffer[16];
+		uint_to_string((u_int)ip, 16, address_buffer);
+		strcat(buffer, address_buffer);
+		strcat(buffer, "]\n");
+
+		write(fd, buffer, 200);
+		// fprintf(fd, " #%d : %s %p 0x%x 0x%x\n", ++i,  b, cursor, (u_int)offset, (u_int)ip);
```
gavv: Please don't commit commented out code.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
...
+
+	while(++i < 13)
+		output_string[j++] = output_string[i];
+	output_string[j] = 0;
+}
+
+/* Return size of backtrace stack by unwinding it.
+ * This must be signal safe.
+ */
+ssize_t backtrace()
```
gavv: As far as I can see, we need this function and performing the unwind twice only to check if the backtrace is available and to print "No backtrace available" if it's not. I think it would be an overkill. In the rare case when there's no backtrace it's acceptable to print just `Backtrace:` and then nothing. So I suggest to get rid of the `backtrace()` function.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+#ifndef LIBUNWIND_HH
+#define LIBUNWIND_HH
```
gavv: Could you please explain why do we need this ifdef?

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
...
+		 * unw_get_proc_name() is signal safe.
+		 */
+		(void) unw_get_proc_name(&cursor, function_name, 40, &offset);
+		/* Printing index, function name, cursor, offset, ip
+		 * using write(). 
+		 * write() & strcat() are signal safe.
+		 */
+		char index_buffer[10];
+		uint_to_string(index, 10, index_buffer);
+		strcat(buffer, index_buffer);
```
gavv: I suggest to check the buffer boundaries here and below and ensure we will not get a buffer overflow.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
...
+		unw_word_t offset, ip;
+		function_name[0] = '\0';
+		/* Get value stored in instruction register
+		 * unw_get_reg() is signal safe.
+		 */
+		unw_get_reg(&cursor, UNW_REG_IP, &ip);
+		/* Get function name & offset
+		 * unw_get_proc_name() is signal safe.
+		 */
+		(void) unw_get_proc_name(&cursor, function_name, 40, &offset);
```
gavv: Why don't we check the return value?

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
gavv: BTW, if we will decide to remove demangling here, backtrace_symbols() and backtrace_symbols_fd() will probably become the same. In this case, we can drop backtrace_symbols() at all and always use backtrace_symbols_fd().

gavv:
Thanks, I've reviewed the libunwind version. Except the comments above, it's looking good.

What are your plans for the bionic version?

BTW, if you want, we could split the PR into two parts: one PR for libunwind and another PR for bionic.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
hrishikeshSuresh: Yes, I overlooked that `i=12` part. Will be fixed.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
hrishikeshSuresh: I didn't find anything function in libunwind that can perform demangling. So I think we can drop backtrace_symbols().

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
hrishikeshSuresh: I was thinking of changing `backtrace()` to `is_backtrace_available()` by just checking if at-least 1 entry is available in the backtrace.  It can be implemented to return just a boolean value. 
true - backtrace available and false - no backtrace available.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
hrishikeshSuresh: This will prevent any error if libunwind.h included twice somewhere (for example a main file where libunwind.h is included again). Since it doesn't impact the code a lot and libunwind is less likely to be used elsewhere, I will remove it.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
gavv: OK.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
gavv: If you think we really need this check, `is_backtrace_available()` is a good option. In this case, can we also avoid iterating the complete backtrace in this function?

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
gavv: On my system none of the libunwind headers define `LIBUNWIND_HH`. So even if `libunwind.h` was already included, we will include it here again.

But this should not be a problem because libunwind headers already contain all necessary guards.

The only concern is that we should define `UNW_LOCAL_ONLY` before including `libunwind.h` first time. But since this include is the first in the .cpp file, we can be sure that this is true.

src/modules/roc_core/target_musl/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,218 @@
...
+		uint_to_string((u_int)offset, 16, offset_buffer);
+		strcat(buffer, offset_buffer);
+		strcat(buffer, ") [");
+
+		char address_buffer[16];
+		uint_to_string((u_int)ip, 16, address_buffer);
+		strcat(buffer, address_buffer);
+		strcat(buffer, "]\n");
+
+		write(fd, buffer, 200);
```
gavv: There are two concerns here:

* we should handle partial writes
* we should use something portable, since write() is not available on all platforms supported by libunwind

For this purpose, I've extracted print_emergency_message() function: 9503e5ea9c324dae7d425d693a7d964bbf66106f. Please rebase on fresh `develop` branch and use it here instead of write().

I've also fixed several issues with our glibc backtrace implementation: d7971de2b23e3f597cfd5ddcb19c6131fdc2d41c. Not related to this PR, just FYI.

hrishikeshSuresh:
I was checking if libunwind can be used for android and in this [file](https://android.googlesource.com/platform/external/libunwind_llvm/+/refs/heads/master/include/libunwind.h), it says the android libunwind api is compatible with the [non-gnu libunwind api](https://www.nongnu.org/libunwind/man/libunwind(3).html). So can we use libunwind with bionic? And go ahead with three implementations - target_glibc, target_libunwind, target_nobacktrace like how you had mentioned earlier.

gavv:
Interesting. I've googled about this version of libunwind a bit.

I've found this repo: https://github.com/alexeikh/android-ndk-backtrace-test

Here is a summary.

* We can use libunwind from Android NDK to print backtrace from a signal handler. But the code will be [architecture-specific](https://github.com/alexeikh/android-ndk-backtrace-test/blob/master/jni/main.c#L135-L153).

* We can use _Unwind_Backtrace() from Bionic to print backtrace from a signal handler as well. The code will be [architecture-specific](https://github.com/alexeikh/android-ndk-backtrace-test/blob/master/jni/main.c#L191-L206) too.

* We can use both libunwind and _Unwind_Backtrace() in an architecture-independent manner, but in this case it will not be possible to use it from a signal handler (we will not see the pre-signal stack).

Adding architecture-specific code and maintaining different versions for different CPUs for such a minor feature would be an overkill. So we're throwing away the first two options.

Then, Android NDK ships with pre-compiled libunwind.a (not sure if it's present on all architectures through), but it doesn't provide libunwind headers. So we will need to ship them manually if we want to use libunwind. The problem with this approach is that these headers may be specific to NDK version, so I'd prefer to avoid this.

Thus, it seems that the best option is to implement print_backtrace() using _Unwind_Backtrace() and to leave print_backtrace_emergency() empty.

But then I've found this page: https://source.android.com/devices/tech/debug

It says that if an Android application crashes, you can find its backtrace in the tombstone file. We should check whether the backtrace will be present there in case of roc_panic(). If it will, we can avoid implementing backtrace on Android at all and just rely on this feature. Unfortunately, I can't test it right now because my Android phone is broken :)

So if you can test it, please do it, and we will see if we need backtrace on Android. If we need it, we should use _Unwind_Backtrace(). If we don't need it, we can leave the bionic version empty.

If you can't test it, you can leave the bionic version as is until someone will have time for it.

gavv:
On the other hand, after some thought, it seems that it would be handy to see the backtrace of a panic in stderr (even if it's also present in tombstone), because the rest panic message is printed to stderr too, and the backtrace is actually a part of this message. And since we already have the code to print it (in this PR), I think we can keep it.

gavv:
So to sum up, I think we need to do the following:

* remove dumpBacktrace_fd() and leave print_backtrace_emergency() for 
 bionic empty
* rename target_musl to target_libunwind (because this code is not specific to musl, but is specific to libunwind)
* add target_nobacktrace with a no-op backtrace implementation
* add target_libunwind and target_nobacktrace support to SConstruct as described above

In result, we will have four backtrace implementations:

* target_glibc - used when libc is glibc
* target_bionic - used when libc is bionic
* target_libunwind - used when libc is not glibc or bionic and libunwind is not disabled
* target_nobacktrace - used when libc is not glibc or bionic and libunwind is disabled

gavv:
I've added libunwind to our alpine environment for CI: https://github.com/roc-project/dockerfiles/commit/4394200fb3a9f83b98a8d522d87fe42b7ac45104

(alpine image is the only one that uses musl and so the only one where we should enable libunwind by default)

hrishikeshSuresh:
In this commit, I haven't replaced `write()` with `print_emergency_message()`, which I will do it later, but I think everything else is done for libunwind. Also, I will start working on SConstruct file.

hrishikeshSuresh:
About checking the backtrace for Android, I can't test right now. Maybe I can do it later, and in case, we need to implement some code, we'll a separate PR for that. For now, `print_backtrace_emergency()` is unimplemented for bionic.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+	/* To store the snapshot of the CPU registers.
+	 */
+	unw_context_t context;
+
+	/* To get snapshot of the CPU register.
+	 * unw_getcontext() is signal safe.
+	 */
+	if(unw_getcontext(&context) < 0)
+		return false;
```
gavv: Please install `clang-format` and run `scons fmt`. It will format code according to our code-style.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,189 @@
...
+		 */
+		unw_get_reg(&cursor, UNW_REG_IP, &ip);
+		/* Get function name & offset
+		 * unw_get_proc_name() is signal safe.
+		 * 'status' checks if unw_get_proc_name() is successful or not.
+		 * 'status = 0', successfully executed the function.
+		 * if status is negative, then we'll stop.
+		 * This is optional and might be useful for debugging.
+		 */
+		int32_t status = unw_get_proc_name(&cursor, function_name, 40, &offset);
```
gavv: I guess we should use sizeof(function_name) instead of 40.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,189 @@
...
+		if(status < 0) {
+			if(status == UNW_EUNSPEC) {
+				write(fd, "An unspecified error occured.", 30);
```
gavv: I don't think we need a verbose error reporting here. The output would be strange: backtrace mixed with errors. Instead, if we can't get the function name, I suggest to exclude it from the line but still print the rest fields.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,189 @@
...
+	 * unw_step() is signal safe.
+	 * buffer = final output that will be written to stderr. 
+	 */
+	uint32_t index = 0;
+	while (unw_step(&cursor) > 0) {
+		char buffer[DefaultLen] = "#";
+		index++;
+		char function_name[128];
+		unw_word_t offset, ip;
+		function_name[0] = '\0';
```
gavv: Could you please rearrange a bit this declarations? We usually follow one simple rule: a variable declaration should be as close as possible to its first use. This increases readability a lot.

So this:

```
char buffer[DefaultLen] = "#";
index++;
char function_name[128];
unw_word_t offset, ip;
function_name[0] = '\0';
...
unw_get_reg(&cursor, UNW_REG_IP, &ip);
...
int32_t status = unw_get_proc_name(&cursor, function_name, 40, &offset);
...
char index_buffer[MaxDigits+1];
uint32_to_string(index, MaxDigits, index_buffer);
strcat(buffer, index_buffer);
```

should be changed to something like:

```
index++;
...
unw_word_t ip;
unw_get_reg(&cursor, UNW_REG_IP, &ip);
...
char function_name[128];
function_name[0] = '\0';
unw_word_t offset;
int32_t status = unw_get_proc_name(&cursor, function_name, 40, &offset);
...
char buffer[DefaultLen] = "#";
char index_buffer[MaxDigits+1];
uint32_to_string(index, MaxDigits, index_buffer);
strcat(buffer, index_buffer);
```

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,189 @@
...
+		}
+		else {
+			/* Printing => index, function name, offset, ip
+			 * using write(). 
+			 * write() & strcat() are signal safe.
+			 */
+			char index_buffer[MaxDigits+1];
+			uint32_to_string(index, MaxDigits, index_buffer);
+			strcat(buffer, index_buffer);
+			strcat(buffer, ": (");
```
gavv: We still need to check buffer boundaries in all these strcat calls.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,189 @@
...
+			strcat(buffer, function_name);
+			strcat(buffer, "+0x");
+		
+			char offset_buffer[MaxDigits+1];
+			uint32_to_string((uint32_t)offset, MaxDigits, offset_buffer);
+			strcat(buffer, offset_buffer);
+			strcat(buffer, ") [");
+
+			char address_buffer[MaxDigits+1];
```
gavv: We could re-use a single buffer instead of the three buffers: index_buffer, offset_buffer, address_buffer.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,189 @@
...
+	if(is_backtrace_available()) {
+		backtrace_symbols_fd(STDERR_FILENO);
+	}
+}
+
+} // namespace core
+} // namespace roc
+
+int main(){
+	return 0;
```
gavv: This should be removed.

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,92 @@
...
  */
 
+#include <cxxabi.h>
+#include <iomanip>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <dlfcn.h>
+#include <unwind.h>
```
gavv: Please remove unnecessary includes here (e.g. iomanip).

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,92 @@
...
+
+
 #include "roc_core/backtrace.h"
 
+/* Implementation of backtrace in bionic/android is similar
+ * to the implementation of backtrace in glibc.
+ */
+struct BacktraceState
```
gavv: We should use an anonymous namespace here.

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,92 @@
...
+			Dl_info info;
+			if (dladdr(addr, &info) && info.dli_sname) {
+				symbol = info.dli_sname;
+			}
+			fprintf(stderr, "#%zd: 0x%x %s", idx, addr, symbol);
+			/* perform demangling
+			 */
+			symbol = abi::__cxa_demangle(symbol, 0, 0, &status);
+			if(status == 0)
+				fprintf(stderr, " %s", symbol);
```
gavv: If demangling succeeded, I suggest to omit the original mangled name, to reduce the line size. We do the same in glibc version.

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,92 @@
...
+			int status = -1;
+
+			Dl_info info;
+			if (dladdr(addr, &info) && info.dli_sname) {
+				symbol = info.dli_sname;
+			}
+			fprintf(stderr, "#%zd: 0x%x %s", idx, addr, symbol);
+			/* perform demangling
+			 */
+			symbol = abi::__cxa_demangle(symbol, 0, 0, &status);
```
gavv: There is a leak here: __cxa_demangle() calls realloc() on its second argument (NULL in this case) and returns the result. So we should free() the demangled name.

https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.3/a01696.html

Also, to reduce the number of allocations here, we can reuse the same buffer for demangling, and then free it at the end. You can find an example in the glibc version of backtrace.

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,92 @@
...
+	}
+}
+
 namespace roc {
 namespace core {
 
-void print_backtrace() {
+namespace {
+
+enum { MaxDepth = 128, MaxLen = 128 };
```
gavv: MaxDepth is not used anywhere. I think we should use MaxDepth instead of MaxLen and remove MaxLen.

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,92 @@
...
+    }
+    return _URC_NO_REASON;
+}
+
+size_t captureBacktrace(void** buffer, size_t max)
+{
+    BacktraceState state = {buffer, buffer + max};
+    _Unwind_Backtrace(unwindCallback, &state);
+
+    return state.current - buffer;
```
gavv: I think we should check if `state.current` is NULL.

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
```diff
@@ -6,15 +6,92 @@
...
+/* Implementation of backtrace in bionic/android is similar
+ * to the implementation of backtrace in glibc.
+ */
+struct BacktraceState
+{
+    void** current;
+    void** end;
+};
+
+static _Unwind_Reason_Code unwindCallback(struct _Unwind_Context* context, void* arg)
```
gavv: Please use `snake_case` for function names.

gavv:
Great. One more round :)

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
hrishikeshSuresh: I will just check this once more.

hrishikeshSuresh: By checking buffer, do you mean say that 200 is not enough for `buffer` array size? or do you want me to do some kind of check before calling strcat()?

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
hrishikeshSuresh: In glibc, the function name is copied into `mangled` array and demangled name is stored in `demangled`, where `mangled` is passed as a param to cxa_demangle(). So this part has to be implemented in bionic the same way?

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,170 @@
...
+bool is_backtrace_available() {
+    /* To store the snapshot of the CPU registers.
+     */
+    unw_context_t context;
+
+    /* To get snapshot of the CPU register.
+     * unw_getcontext() is signal safe.
+     */
+    if (unw_getcontext(&context) < 0)
+        return false;
```
gavv: Nitpick: we always use `{}` so this should become:

```
    if (unw_getcontext(&context) < 0) {
        return false;
    }
```

(the same in other places)

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    /* To point to the current frame in the call stack.
+ */
```
gavv: Nitpick: bad comment formatting here and below (if it's caused by clang-format, maybe you're using an outdated version? I'm using 7.1.0.)

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+/* Function to convert unsigned int to decimal/hex string.
+ * This function is signal safe.
+ * Parameters :
+ * number = the number to be converted.
+ * base = 10 for decimal to char* and 16 for hex to char* conversion.
+ * output_string = the decimal/hex string
+ */
+
+namespace {
```
gavv: Nitpick: please move the anonymous namespace inside roc::core and add a trailing "// namespace" comment, like this:

```
namespace roc {
namespace core {

namespace {

...

} // namespace

...
```

(just to be inline with the code-style)

gavv: The same for the bionic version.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
gavv: The second. Since the line length is not known for sure, we should implement boundary checking.

For example, you can implement a helper function like:

```
void safe_strcat(char* buffer, size_t buffer_size, const char* str) {
    ...
}
```

... which will truncate the string if it is too long and ensure that we never go outside the buffer boundaries, and then use it instead of strcat().

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
gavv: No. In the glibc version, the mangled name is copied to form a null-terminated string. Here you already have a null-terminated string (`symbol`) so there is no need to make a copy.

```
char* demangled_name = (char*)malloc(MaxLen * sizeof(char));
demangled_name = abi::__cxa_demangle(mangled_name, 0, 0, &status);
```

There is a leak here, again: you malloc() a buffer and then overwrite it with another buffer returned by __cxa_demangle(); so the original buffer will be never freed.

The correct and optimal (by number of allocations) code would be something like:

```
size_t demangled_size = MaxLen;
char* demangled_name = (char*)malloc(demangled_size);

for (...) {
   ...
    demangled_name =
        abi::__cxa_demangle(symbol, demangled_name, &demangled_size, &status);
    if (status == 0) {
        fprintf(stderr, " %s\n", demangled_name);
    } else {
        fprintf(stderr, " %s\n", symbol);
    }
}

free(demangled_name);
```

So you should pass demangled_name to __cxa_demangle() each time; and __cxa_demangle() will return either the same or reallocated buffer and will also update the demangled_size variable with the size of the buffer. When the loop finishes, you call free().

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
hrishikeshSuresh: Can I use `strncat()` in that case? http://man7.org/linux/man-pages/man3/strncat.3.html I can specify how many characters can be copied to the buffer.

gavv: Take a look at the strncat() limits the size of the `src` but it does not take into account the size of `dest` so it has the same problem.

You could use strlcat() for that but it's non-portable so we can't use it in target_libunwind.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -16,18 +16,38 @@
...
+void safe_strcat(char* buffer, size_t& buffer_size, const char* str) {
+    size_t len = strlen(str);
+	/* Checking if there is enough space in the buffer
+	 */
+    if (buffer_size + len < MaxLen) {
+        strcat(buffer, str);
+        buffer_size += len;
+    } else {
+        size_t length_to_be_truncated = buffer_size + len - MaxLen;
+        for (size_t i = 0; i < len - length_to_be_truncated; i++) {
```
hrishikeshSuresh: truncation happens with the use of this condition, preventing buffer overflow.

src/modules/roc_core/target_bionic/roc_core/backtrace.cpp
```diff
@@ -6,12 +6,90 @@
...
+            Dl_info info;
+            if (dladdr(addr, &info) && info.dli_sname) {
+                symbol = info.dli_sname;
+            }
+            fprintf(stderr, "#%zd: 0x%p", idx, addr);
+
+            /* perform demangling
+             */
+            demangled_name =
+                abi::__cxa_demangle(symbol, demangled_name, &demangled_size, &status);
```
gavv: If dladdr() failed and symbol is empty, we should not try to perform demangling.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,192 @@
...
+    if (buffer_size + len < MaxLen) {
+        strcat(buffer, str);
+        buffer_size += len;
+    } else {
+        size_t length_to_be_truncated = buffer_size + len - MaxLen;
+        for (size_t i = 0; i < len - length_to_be_truncated && buffer_size < MaxLen-1; i++) {
+            buffer[buffer_size] = str[i];
+            buffer_size++;
+        }
+        buffer[MaxLen - 1] = '\n';
```
gavv: I guess we need `\0` instead of `\n`.

But I suggest to simplify the function and remove this piece at all:

```
void safe_strcat(char* buffer, size_t& buffer_size, const char* str) {
    size_t len = strlen(str);
    if (len > MaxLen - buffer_size - 1) {
        len = MaxLen - buffer_size - 1;
    }
    if (len > 0) {
        strncat(buffer, str, len);
        buffer_size += len;
    }
}
```

gavv:
OK, so the remaining issues are:

* the two minor comments above
* use print_emergency_message() instead of write() (and don't include unistd.h)
* update SConstruct

hrishikeshSuresh:
In SConstruct file, I am not sure what exactly to be done in part we are have to build libunwind (` if 'target_libunwind' in download_dependencies`) part, because in `target_sox` we are checking for a lot of dependent libraries and for libunwind, there are no such dependencies.

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,188 @@
...
+ */
+void safe_strcat(char* buffer, size_t& buffer_size, const char* str) {
+    size_t len = strlen(str);
+    /* Checking if there is enough space in the buffer
+     */
+    if (len < MaxLen - buffer_size - 1) {
+        len = MaxLen - buffer_size - 1;
+    }
+    if (len > 0) {
+        strncat(buffer, str, len);
```
gavv: `strncat(buffer, str, len)` is incorrect. It should be `strncat(buffer + buffer_size, str, len)`.

Also, GCC 8 produces a warning on this piece:

```
error: 'char* strncat(char*, const char*, size_t)'
specified bound 3 equals source length [-Werror=stringop-overflow=]
         strncat(buffer + buffer_size, str, len);
```

Here is how we can suppress the warning:

```
        memcpy(buffer + buffer_size, str, len);
        buffer_size += len;
        buffer[buffer_size] = '\0';
```

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,188 @@
...
+void print_backtrace() {
+    if (!is_backtrace_available()) {
+        fprintf(stderr, "No backtrace available\n");
+    } else {
+        fprintf(stderr, "Backtrace:\n");
+        backtrace_symbols();
+    }
+}
+
+void print_backtrace_emergency() {
```
gavv: Did you test printing backtrace from a signal handler, e.g. on a crash?

It does not work for me. The backtrace does not print the pre-signal stack for some reason.

```
ERROR: caught SIGSEGV

#1: (_ZN3roc4core12_GLOBAL__N_114signal_handlerEiP9siginfo_tPv+0x6d) [6662a8fd]
#2: (sigwaitinfo+0x8) [e3de4e21]
#3: (+0x0) [e3d766d9]
```

scripts/3rdparty.py
```diff
@@ -554,6 +554,28 @@ def makeenv(envlist):
     execute('make -j', logfile)
     install_files('src/sox.h', os.path.join(builddir, 'include'))
     install_files('src/.libs/libsox.a', os.path.join(builddir, 'lib'))
+
+elif name == 'libunwind':
+	download(
+		'http://download.savannah.nongnu.org/releases/libunwind/libunwind-%s.%s.%s.tar.gz' %(ver, ver, ver),
```
gavv: This code produces an incorrect url and paths, e.g.:

```
http://download.savannah.nongnu.org/releases/libunwind/libunwind-1.2.1.1.2.1.1.2.1.tar.gz
```

It also has other problems:

* on my system libunwind fails to build with -fvisibility=hidden
* on alpine libunwind fails to build without --disable-coredump
* the installation step does not work for me (make install prefix=src)

Here is a fixed version that works for me:

```
elif name == 'libunwind':
    download(
        'http://download.savannah.nongnu.org/releases/libunwind/libunwind-%s.tar.gz' % ver,
        'libunwind-%s.tar.gz' % ver,
        logfile,
        vendordir)
    extract('libunwind-%s.tar.gz' % ver,
            'libunwind-%s' % ver)
    os.chdir('src/libunwind-%s' % ver)
    execute('./configure --host=%s %s %s %s' % (
        toolchain,
        makeenv(envlist),
        makeflags(workdir, toolchain, deplist, variant=variant),
        ' '.join([
            '--enable-static',
            '--disable-shared',
            '--disable-coredump',
            '--disable-ptrace',
            '--disable-setjmp',
           ])), logfile)
    execute('make -j', logfile)
    install_files('include/*.h', os.path.join(builddir, 'include'))
    install_files('src/.libs/libunwind.a', os.path.join(builddir, 'lib'))
```

BTW, please use spaces instead of tabs and remove spurious empty lines.

gavv:
Please re-target pull request to the `develop` branch (see https://roc-project.github.io/roc/docs/development/version_control.html#pull-requests)

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
hrishikeshSuresh: I think I figured it out. Is it because I'm always travelling through the stack and when a crash occurs, `unw_get_reg(&cursor, UNW_REG_IP, &ip)` will fail and just quit? If that is what's happening, then I should handle the case where it fails and print the rest of backtrace.

gavv: Quick googling showed that we should not use `unw_getcontext()` for signal handler; instead, it seems that we should use `ucontext` passed to the handler.

I've tried this:

* added a `void* context` argument to `print_backtrace_emergency()`

* forwarded it from `signal_handler()` in `crash.cpp`:

      void signal_handler(int sig, siginfo_t* si, void* context) {
          ...
          print_backtrace_emergency(context);

* removed the local `context` declaration and `unw_getcontext()` call and used the provided pointer instead

And after this I saw the full backtrace.

Could you please check the docs that this is the recommended/supported use-case?

Note the we still need to initialize a local context using `unw_getcontext()` when the user calls `print_backtrace()` (regular call) instead of `print_backtrace_emergency()` (call from signal handler).

hrishikeshSuresh: I didn't find any doc using `ucontext_t` with `unw_context`.
In the unw_getcontext() [doc](https://www.nongnu.org/libunwind/man/unw_getcontext(3).html), it is given that unw_getcontext() is safe to be used from a signal handler. But passing the context as an argument to `print_backtrace_emergency()` does make sense.

gavv:
```
grep -E 'On.*ucontext_t' . -R
./libunwind-x86_64.h:118:/* On x86_64, we can directly use ucontext_t as the unwind context.  */
./libunwind-arm.h:253:/* On ARM, we define our own unw_tdep_context instead of using ucontext_t.
./libunwind-hppa.h:103:/* On PA-RISC, we can directly use ucontext_t as the unwind context.  */
./libunwind-ppc64.h:245:/* On ppc64, we can directly use ucontext_t as the unwind context.  */
./libunwind-ia64.h:150:/* On IA-64, we can directly use ucontext_t as the unwind context.  */
./libunwind-x86.h:164:/* On x86, we can directly use ucontext_t as the unwind context.  */
./libunwind-aarch64.h:172:/* On AArch64, we can directly use ucontext_t as the unwind context.  */
./libunwind-mips.h:130:/* On x86, we can directly use ucontext_t as the unwind context.  FIXME for
./libunwind-ppc32.h:181:/* On ppc, we can directly use ucontext_t as the unwind context.  */
```

Well, on most platforms ucontext_t can be used with libunwind, but still on some platforms (ARM) it can not.

gavv: Hmmm.

I've checked out your branch once more and inserted a crash in a few other places and now I'm seeing the full backtrace without passing context from the signal handler.

Very strange, but I can't reproduce the issue anymore. I'll make a few more experiments, but if I will not reproduce this, and if it also works for you, let's keep it as is (without passing ucontext_t).

gavv: Huh.

The backtrace is full on glibc but is truncated on alpine/musl.

I can reproduce it by inserting a crash somewhere (e.g. division by zero) and running tests in docker:

```
docker run --rm -t --cap-add SYS_PTRACE -u 1000 -v `pwd`:/opt/roc \
  -w /opt/roc rocproject/travis-alpine \
    scons -Q --enable-werror --build-3rdparty=openfec test
```

gavv: Silly me! That's because I did not provide `--enable-debug` flag when building on alpine. I added it and it worked.

So, sorry for the confusion. The current libunwind version is good.

Except that the formatting is broken, I had to fix it:

```
@@ -33,7 +33,7 @@ void safe_strcat(char* buffer, size_t& buffer_size, const char* str) {
     if (len > 0) {
         memcpy(buffer + buffer_size, str, len);
         buffer_size += len;
-        buffer[MaxLen - 1] = '\0';
+        buffer[buffer_size] = '\0';
     }
 }
 
@@ -162,6 +162,7 @@ void backtrace_symbols() {
 
         uint32_to_string((uint32_t)offset, 16, number);
         safe_strcat(buffer, current_buffer_size, number);
+        safe_strcat(buffer, current_buffer_size, "\n");
 
         print_emergency_message(buffer);
     }
```

PR #257, opened 2019-09-16
Added --list-drivers -L support Issue:251

tomd1990:
Made the get_drivers function an abstract function for the iBackend class. Also updated the ggo/ main.cpp files accordingly.

Currently sox can support all of the following formats/ audio device drivers (based on what sox_get_format_fn gives me)

AUDIO FILE FORMATS: aifc aiffc aiff aif al au snd avr cdda cdr cvsd cvs cvu dat dvms vms f4 f32 f8 f64 gsrt hcom htk ima la lu maud prc raw s1 s8 sb s2 s16 sw s3 s24 s4 s32 sl sf ircam sln smp sndr sndt sox sph nist 8svx txw u1 u8 ub sou fssd u2 u16 uw u3 u24 u4 u32 ul voc vox wav wavpcm amb wve xa gsm lpc10 lpc
PLAYLIST FORMATS: m3u pls
AUDIO DEVICE DRIVERS: alsa ossdsp oss pulseaudio

I wasn't sure if there needs to be any separation between Audio FIle formats and device drivers, so I just bundled them together in the output.

### DIFF
--- a/src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
+++ b/src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
@@ -294,11 +294,11 @@ void OFDecoder::reset_session_() {
 
     if (OF_STATUS_OK
         != of_set_callback_functions(
-            of_sess_, source_cb_,
-            // OpenFEC doesn't repair fec-packets in case of Reed-Solomon FEC
-            // and prints curses to the console if we give it the callback for that
-            codec_id_ == OF_CODEC_REED_SOLOMON_GF_2_M_STABLE ? NULL : repair_cb_,
-            (void*)this)) {
+               of_sess_, source_cb_,
+               // OpenFEC doesn't repair fec-packets in case of Reed-Solomon FEC
+               // and prints curses to the console if we give it the callback for that
+               codec_id_ == OF_CODEC_REED_SOLOMON_GF_2_M_STABLE ? NULL : repair_cb_,
+               (void*)this)) {
         roc_panic("of decoder: of_set_callback_functions() failed");
     }
 }
--- a/src/modules/roc_sndio/DriverInfo.cpp
+++ b/src/modules/roc_sndio/DriverInfo.cpp
@@ -1,0 +1,13 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Sink Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_sndio/DriverInfo.h"
+
+namespace roc {
+namespace sndio {} // namespace sndio
+} // namespace roc
--- a/src/modules/roc_sndio/DriverInfo.h
+++ b/src/modules/roc_sndio/DriverInfo.h
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_sndio/DriverInfo.h
+//! @brief Driver info interface.
+
+#ifndef ROC_SNDIO_DRIVERINFO_H_
+#define ROC_SNDIO_DRIVERINFO_H_
+
+namespace roc {
+namespace sndio {
+
+//! Driver info interface.
+class DriverInfo {
+public:
+    DriverInfo();
+    DriverInfo(const char* name)
+        : name(name){};
+    const char* name;
+};
+
+} // namespace sndio
+} // namespace roc
+
+#endif // ROC_SNDIO_DRIVERINFO_H_
--- a/src/modules/roc_sndio/backend_dispatcher.h
+++ b/src/modules/roc_sndio/backend_dispatcher.h
@@ -12,10 +12,12 @@
 #ifndef ROC_SNDIO_BACKEND_DISPATCHER_H_
 #define ROC_SNDIO_BACKEND_DISPATCHER_H_
 
+#include "roc_core/array.h"
 #include "roc_core/iallocator.h"
 #include "roc_core/noncopyable.h"
 #include "roc_core/shared_ptr.h"
 #include "roc_core/singleton.h"
+#include "roc_sndio/DriverInfo.h"
 #include "roc_sndio/ibackend.h"
 #include "roc_sndio/isink.h"
 #include "roc_sndio/isource.h"
@@ -45,6 +47,9 @@ public:
                          const char* driver,
                          const char* input,
                          const Config& config);
+
+    //! Append supported dirvers from all registered backends to Array
+    void get_drivers(core::Array<DriverInfo>& arr);
 
 private:
     friend class core::Singleton<BackendDispatcher>;
--- a/src/modules/roc_sndio/ibackend.h
+++ b/src/modules/roc_sndio/ibackend.h
@@ -12,8 +12,10 @@
 #ifndef ROC_SNDIO_IBACKEND_H_
 #define ROC_SNDIO_IBACKEND_H_
 
+#include "roc_core/array.h"
 #include "roc_core/iallocator.h"
 #include "roc_core/shared_ptr.h"
+#include "roc_sndio/DriverInfo.h"
 #include "roc_sndio/config.h"
 #include "roc_sndio/isink.h"
 #include "roc_sndio/isource.h"
@@ -55,6 +57,9 @@ public:
                                  const char* driver,
                                  const char* input,
                                  const Config& config) = 0;
+
+    //! Append supported dirvers to Array
+    virtual void get_drivers(core::Array<DriverInfo>& arr) = 0;
 };
 
 } // namespace sndio
--- a/src/modules/roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_backend.cpp
+++ b/src/modules/roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_backend.cpp
@@ -57,5 +57,19 @@ ISource* PulseaudioBackend::open_source(core::IAllocator&,
     return NULL;
 }
 
+void PulseaudioBackend::get_drivers(core::Array<DriverInfo>& arr) {
+    const char* format_name = "pulseaudio";
+    size_t n;
+    for (n = 0; n < arr.size(); n++) {
+        if (strcmp(format_name, arr[n].name) == 0) {
+            return;
+        }
+    }
+    if (arr.grow(arr.size() + 1)) {
+        DriverInfo newDriver(format_name);
+        arr.push_back(newDriver);
+    }
+}
+
 } // namespace sndio
 } // namespace roc
--- a/src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
+++ b/src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
@@ -196,5 +196,36 @@ ISource* SoxBackend::open_source(core::IAllocator& allocator,
     return source.release();
 }
 
+void SoxBackend::get_drivers(core::Array<DriverInfo>& arr) {
+    const sox_format_tab_t* formats = sox_get_format_fns();
+    char const* const* format_names;
+    size_t n;
+    for (n = 0; formats[n].fn; n++) {
+        sox_format_handler_t const* handler = formats[n].fn();
+        if (!(handler->flags & SOX_FILE_DEVICE)
+            || ((handler->flags & SOX_FILE_DEVICE)
+                && !(handler->flags & SOX_FILE_PHONY))) {
+            for (format_names = handler->names; *format_names; ++format_names) {
+                if (!strchr(*format_names, '/')) {
+                    size_t x;
+                    for (x = 0; x < arr.size(); x++) {
+                        if (strcmp(*format_names, arr[x].name) == 0) {
+                            break;
+                        }
+                    }
+                    if (x == arr.size()) {
+                        DriverInfo newDriver(*format_names);
+                        if (arr.grow(arr.size() + 1)) {
+                            arr.push_back(newDriver);
+                        }
+                    } else {
+                        continue;
+                    }
+                }
+            }
+        }
+    }
+}
+
 } // namespace sndio
 } // namespace roc
--- a/src/tests/roc_audio/test_freq_estimator.cpp
+++ b/src/tests/roc_audio/test_freq_estimator.cpp
@@ -21,7 +21,7 @@ const double Epsilon = 0.0001;
 
 } // namespace
 
-TEST_GROUP(freq_estimator) {};
+TEST_GROUP(freq_estimator){};
 
 TEST(freq_estimator, initial) {
     FreqEstimator fe(Target);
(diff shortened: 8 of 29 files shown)


gavv:
Great! I'll take a look.

Regarding to formatting. I guess you have run `scons fmt` right? What version of clang-format do you use? I use 7.1.0 and it seems that the formatting is a bit different (your version removed some spaces?). In this case I think we should choose the most recent version of these two.

src/modules/roc_sndio/DriverInfo.h
```diff
@@ -0,0 +1,30 @@
+/*
```
gavv: Please use `snake_case` for filenames (i.e. `driver_info.h`)

src/modules/roc_sndio/backend_dispatcher.h
```diff
@@ -46,6 +48,9 @@ class BackendDispatcher : public core::NonCopyable<> {
                          const char* input,
                          const Config& config);
 
+    //! Append supported dirvers from all registered backends to Array
```
gavv: s/dirvers/drivers/

src/tools/roc_recv/cmdline.ggo
```diff
@@ -3,6 +3,8 @@ usage "roc-recv OPTIONS"
 
 section "Options"
 
+    option "list-drivers" L "Lists all supported audio drivers" optional
```
gavv: s/Lists/List/ (in line with other messages)

src/tools/roc_recv/main.cpp
```diff
@@ -41,6 +43,15 @@ int main(int argc, char** argv) {
 
     pipeline::ReceiverConfig config;
 
+    if (args.list_drivers_given) {
+        core::HeapAllocator allocator;
```
gavv: We already have HeapAllocator declaration in this function. Just move it to the top and use here.

src/tools/roc_recv/main.cpp
```diff
@@ -41,6 +43,15 @@ int main(int argc, char** argv) {
 
     pipeline::ReceiverConfig config;
 
+    if (args.list_drivers_given) {
```
gavv: I'd move this `if` before `pipeline::ReceiverConfig config;` and return from main() at the end:

```
    if (args.list_drivers_given) {
        ...
        return 0;
    }
```

src/tools/roc_recv/main.cpp
```diff
@@ -41,6 +43,15 @@ int main(int argc, char** argv) {
...
+        core::HeapAllocator allocator;
+        core::Array<sndio::DriverInfo> DriverList(allocator);
```
gavv: Please use `snake_case` for variable names.

src/tools/roc_recv/main.cpp
```diff
@@ -41,6 +43,15 @@ int main(int argc, char** argv) {
...
+        sndio::BackendDispatcher::instance().get_drivers(DriverList);
+        for (size_t n = 0; n < DriverList.size(); n++) {
+            printf("%s \n", DriverList[n].name);
```
gavv: Extra space.

src/tools/roc_recv/main.cpp
gavv: The same for roc_send version.

src/tools/roc_recv/main.cpp
gavv: The same for roc_send version.

src/tools/roc_recv/main.cpp
gavv: The same for roc_send version.

src/tools/roc_recv/main.cpp
gavv: The same for roc_send version.

src/tools/roc_recv/cmdline.ggo
gavv: The same for roc_send version.

src/modules/roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_backend.cpp
```diff
@@ -57,5 +57,19 @@ ISource* PulseaudioBackend::open_source(core::IAllocator&,
...
+    const char* format_name = "pulseaudio";
+    size_t n;
+    for (n = 0; n < arr.size(); n++) {
+        if (strcmp(format_name, arr[n].name) == 0) {
+            return;
+        }
+    }
+    if (arr.grow(arr.size() + 1)) {
+        DriverInfo newDriver(format_name);
+        arr.push_back(newDriver);
```
gavv: I suggest to extract this code into a function, something like `add_driver_uniq(array, name)` and re-use it in PulseaudioBackend and SoxBackend.

src/modules/roc_sndio/DriverInfo.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef ROC_SNDIO_DRIVERINFO_H_
+#define ROC_SNDIO_DRIVERINFO_H_
+
+namespace roc {
+namespace sndio {
+
+//! Driver info interface.
+class DriverInfo {
+public:
```
gavv: We usually use `struct` instead of `class` for classes with public fields.

src/modules/roc_sndio/DriverInfo.h
```diff
@@ -0,0 +1,30 @@
...
+    DriverInfo();
+    DriverInfo(const char* name)
+        : name(name){};
+    const char* name;
```
gavv:
```
const char* name;
```

This assumes that the name is a constant string which lifetime is not limited. While this is true for current backends, this looks a bit dangerous because these strings come from external libraries. So I'd prefer to avoid this assumption here and use something like `char name[16]` here.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -196,5 +196,36 @@ ISource* SoxBackend::open_source(core::IAllocator& allocator,
...
+    const sox_format_tab_t* formats = sox_get_format_fns();
+    char const* const* format_names;
+    size_t n;
+    for (n = 0; formats[n].fn; n++) {
+        sox_format_handler_t const* handler = formats[n].fn();
+        if (!(handler->flags & SOX_FILE_DEVICE)
+            || ((handler->flags & SOX_FILE_DEVICE)
+                && !(handler->flags & SOX_FILE_PHONY))) {
+            for (format_names = handler->names; *format_names; ++format_names) {
+                if (!strchr(*format_names, '/')) {
```
gavv: Could you explain why do we need this check?

gavv:
> I wasn't sure if there needs to be any separation between Audio FIle formats and device drivers, so I just bundled them together in the output.

I think it would be convenient to print them separately, like:

```
device drivers:
  alsa
  pulseaudio
  ...

file drivers:
  wav
  ...
```

We already have `IBackend::ProbeFlags`. We can rename it to something like `FilterFlags` and then use them like this:

```
backend_dispatcher.get_drivers(arr, FilterDevice);
// print
backend_dispatcher.get_drivers(arr, FilterFile);
// print
```

gavv:
I've reviewed this, see the comments above. Most of them are cosmetic thought.

src/modules/roc_fec/target_openfec/roc_fec/of_decoder.cpp
```diff
@@ -294,11 +294,11 @@ void OFDecoder::reset_session_() {
 
     if (OF_STATUS_OK
         != of_set_callback_functions(
-            of_sess_, source_cb_,
-            // OpenFEC doesn't repair fec-packets in case of Reed-Solomon FEC
-            // and prints curses to the console if we give it the callback for that
-            codec_id_ == OF_CODEC_REED_SOLOMON_GF_2_M_STABLE ? NULL : repair_cb_,
-            (void*)this)) {
+               of_sess_, source_cb_,
```
dshil: Could you please rebase on the develop branch? It seems that we fixed this formatting recently in https://github.com/roc-project/roc/commit/fa43b484a36da280e0aa7b9b787f7d2010be4421.

src/modules/roc_sndio/DriverInfo.cpp
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * This Sink Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_sndio/DriverInfo.h"
+
+namespace roc {
+namespace sndio {} // namespace sndio
```
dshil: NITPICK: we use the following formatting:

```cpp
namespace roc {
namespace sndio {
} // namespace sndio
} // namespace roc
```

src/modules/roc_sndio/DriverInfo.h
```diff
@@ -0,0 +1,30 @@
...
+#define ROC_SNDIO_DRIVERINFO_H_
+
+namespace roc {
+namespace sndio {
+
+//! Driver info interface.
+class DriverInfo {
+public:
+    DriverInfo();
+    DriverInfo(const char* name)
```
dshil:
```cpp
explicit DriverInfo(const char* name)
```

dshil:
@tomd1990, a small nitpick about the commit message.

You have:

```
 Added --list-drivers -L support Issue:251 
```

It would be nice to have something similar to:

```
Added --list-drivers -L support

Relates: #251 
```

I want to separate the commit title and the issue reference. Also note that we can use `#` before the issue number, Github will link this issue with the current PR.

src/modules/roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_backend.cpp
```diff
@@ -57,5 +57,19 @@ ISource* PulseaudioBackend::open_source(core::IAllocator&,
     return NULL;
 }
 
+void PulseaudioBackend::get_drivers(core::Array<DriverInfo>& arr) {
+    const char* format_name = "pulseaudio";
+    size_t n;
```
dshil: Please, do not use uninitialized variable and put its initialization inside the for-loop.

src/modules/roc_sndio/ibackend.h
```diff
@@ -55,6 +57,9 @@ class IBackend {
                                  const char* driver,
                                  const char* input,
                                  const Config& config) = 0;
+
+    //! Append supported dirvers to Array
+    virtual void get_drivers(core::Array<DriverInfo>& arr) = 0;
```
dshil: @gavv, could we use `core::List` instead of `core::Array` ?

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -196,5 +196,36 @@ ISource* SoxBackend::open_source(core::IAllocator& allocator,
     return source.release();
 }
 
+void SoxBackend::get_drivers(core::Array<DriverInfo>& arr) {
+    const sox_format_tab_t* formats = sox_get_format_fns();
+    char const* const* format_names;
+    size_t n;
```
dshil: Please, do not use uninitialized variable and put its initialization inside the for-loop.

src/tests/roc_audio/test_freq_estimator.cpp
```diff
@@ -21,7 +21,7 @@ const double Epsilon = 0.0001;
 
 } // namespace
 
-TEST_GROUP(freq_estimator) {};
+TEST_GROUP(freq_estimator){};
```
dshil: I think that this change does not relate to the current PR.

Very rarely some files can be pushed to the develop branch without the required formatting. As a result, next time you run `scons fmt -Q` you will see these files to be formatted. In such case I would suggest to reset these unrelated changes and commit only changes related to the current PR.

src/modules/roc_sndio/ibackend.h
gavv: We don't want it. It will require to allocate every DriverInfo object on heap.

src/tests/roc_audio/test_freq_estimator.cpp
gavv: See my comment about formatting above.

src/modules/roc_sndio/ibackend.h
dshil: Well, we about allocations during the array grow?

gavv: With List, there is always one allocation per object. With Array, the allocations can be amortized so we will have log(N) allocations per N objects.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
tomd1990: So I've ran it with and without this check and it makes no difference.  The only reason I put this in is because the sox.c library also includes it, (display_supported_formats I used this function as a guideline to extract the names from sox_format_fns).  I'm not sure why they use it though, none of the names have '/', but I felt that if they included it I should do the same.

gavv: OK, I see, let's keep it :) Could you then add a comment that we replicate the behavior of the display_supported_formats() from sox?

src/tests/roc_audio/test_freq_estimator.cpp
tomd1990: Yeah I was definitely using an older version,  but I shouldn't have included these files in my PR.  I'll limit my scope next time

tomd1990:
> Great! I'll take a look.
> 
> Regarding to formatting. I guess you have run `scons fmt` right? What version of clang-format do you use? I use 7.1.0 and it seems that the formatting is a bit different (your version removed some spaces?). In this case I think we should choose the most recent version of these two.

Sorry for the late response,

I ran scons fmt, but my clang-format is definitely older, version 6.0.0.  This probably explains the difference.  I'll update and run it again.

Thanks for all of the feedback, Ill make these changes soon.

src/tools/roc_recv/main.cpp
tomd1990: Just curious, why have this before the ReceiverConfig config?

gavv: Just to group the code logically and move the config declaration closer to the code that is filling it. So it will look like:

```
if (args.list_drivers_given) {
    // the code unrelated to config
    return 0;
}

pipeline::ReceiverConfig config;

// the code filling the config
```

A general (quite common) recommendation is to declare variables as close as possible to their first use. This is a special case of an even more general recommendation to declare variables as locally as possible.

src/modules/roc_sndio/DriverInfo.h
tomd1990: I'm not sure I follow, while external libraries can store strings as constant char* or char[], wouldnt it be up to each backend's implementation of get_drivers to resolve this?  This is the reason I defined it as an abstract function.  Or is it just more convenient to store it as an array?

gavv: If the external library stores names as a static constant string (which lifetime is not limited), we can go both ways: store the name in `char*` or copy it to `char[N]`.

But if the external library provides a temporary string which lifetime ends after get_drivers() returns, we can't just store the pointer in `char*` because it will become invalid after get_driver() returns. So in this case we *should* copy it to `char[N]`.

So the second approach (`char[N]`) is more universal: it makes no assumptions about the lifetime of the strings provided by backend.

tomd1990:
I'm closing this PR as merging this would make things messy, I'll submit the commit in a new PR

Issue #251, opened 2019-09-11
Print list of available audio drivers


gavv:
## Intro

We have four concepts in `roc_sndio` module:

* audio backend

  Audio backend provides one or multiple audio drivers, and each driver can be used to work with audio sources and audio sinks of some specific type.

  In code, a backend is an implementation of sndio::IBackend interface.

  Examples of backends are: SoxBackend (opens sources and sinks using libsox; it supports numerous drivers, e.g. "wav" or "alsa"), PulseaudioBackend (opens sources and sinks using libpulse; it supports a single driver "pulseaudio").

* audio driver

  Audio driver works with sources and sinks of some specific type.

  In code, a driver is just a string passed to backend implementation when opening a source or a sink. It depends on the specific backend how to interpret this string.

  Examples of drivers are: "wav" (WAV file), "alsa" (ALSA device), "pulseaudio" (PulseAudio device).

* audio source

  Audio source represents a recording audio device or input audio file. We can read sound stream from a source.

  In code, a source is an implementation of sndio::ISource. Sources are opened using IBackend::open_source() method.

* audio sink

  Audio sink represents a playback audio device or output audio file. We can write sound stream to a sink.

  In code, a sink is an implementation of sndio::ISink. Sinks are opened using IBackend::open_sink() method.

Backends are not visible to the user. When the user starts roc-recv or roc-send, they specify the driver and the input name (in roc-send) or an output name (in roc-recv). The name can be a file name or a device name, depending on the driver.

The driver and the name are passed to BackendDispatcher. It looks at the driver name and selects a backend that can handle this driver. Then it passes the driver name and the input or output name to the backend, and the backend creates a source (in roc-send) or a sink (in roc-recv).

## Task

This task is rather simple: we need to be able to print the list of supported drivers.

Something like this:

```
$ roc-recv -L
pulseaudio
alsa
wav
...
```

## Implementation

The following implementation is suggested:

* Add a new structure `sndio::DriverInfo`. It should contain at list one field `const char* name`. We can also add other fields, like backend name, driver type (a file or a device), but this can be done later.

* Add a new method `IBackend::get_drivers(core::Array<DriverInfo>&)` that will add info to the list about every driver supported by the backend. If a driver is already present in the list, it should not be added. It can happen because the same driver can be supported by different backends (e.g. "pulseaudio" driver is supported by PulseaudioBackend and SoxBackend).

* Implement this method for all IBackend implementations.

[...]

PR #262, opened 2019-09-19
Add --list-drivers -L support

tomd1990:
References: #251, #257

### DIFF
--- a/src/modules/roc_sndio/driver_info.cpp
+++ b/src/modules/roc_sndio/driver_info.cpp
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Sink Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <string.h>
+
+#include "roc_sndio/driver_info.h"
+
+namespace roc {
+namespace sndio {
+
+DriverInfo::DriverInfo() {
+    for (size_t x = 0; x < MaxSize; x++) {
+        name[x] = '\0';
+    }
+}
+
+DriverInfo::DriverInfo(const char* driver_name) {
+    size_t length = strlen(driver_name);
+    if (length > MaxSize - 1) {
+        length = MaxSize - 1;
+    }
+    strncpy(name, driver_name, MaxSize);
+    name[length] = '\0';
+}
+
+bool add_driver_uniq(core::Array<DriverInfo>& arr, const char* driver_name) {
+    roc_panic_if(driver_name == NULL);
+    for (size_t n = 0; n < arr.size(); n++) {
+        if (strcmp(driver_name, arr[n].name) == 0) {
+            return false;
+        }
+    }
+    if (arr.grow(arr.size() + 1)) {
+        DriverInfo new_driver(driver_name);
+        arr.push_back(new_driver);
+        return true;
+    } else {
+        return false;
+    }
+}
+
+} // namespace sndio
+} // namespace roc
--- a/src/modules/roc_sndio/driver_info.h
+++ b/src/modules/roc_sndio/driver_info.h
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_sndio/driver_info.h
+//! @brief Driver info interface.
+
+#ifndef ROC_SNDIO_DRIVER_INFO_H_
+#define ROC_SNDIO_DRIVER_INFO_H_
+
+#include "roc_core/array.h"
+
+namespace roc {
+namespace sndio {
+
+//! Driver info interface.
+struct DriverInfo {
+    DriverInfo();
+
+    //! Max size of string
+    enum { MaxSize = 20 };
+
+    //! Parameterized Constructor initializes name, assumes driver_name is terminated with
+    //! null char
+    explicit DriverInfo(const char* driver_name);
+
+    //! Placeholder for the driver name
+    char name[MaxSize];
+};
+
+//! Append driver to array and ensure uniqueness, returns false if unable to add item,
+//! or if unable to allocate space for item
+bool add_driver_uniq(core::Array<DriverInfo>& arr, const char* driver_name);
+
+} // namespace sndio
+} // namespace roc
+
+#endif // ROC_SNDIO_DRIVER_INFO_H_
--- a/src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
+++ b/src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
@@ -135,11 +135,11 @@ bool SoxBackend::probe(const char* driver, const char* inout, int flags) {
     }
 
     if (handler->flags & SOX_FILE_DEVICE) {
-        if ((flags & ProbeDevice) == 0) {
+        if ((flags & FilterDevice) == 0) {
             return false;
         }
     } else {
-        if ((flags & ProbeFile) == 0) {
+        if ((flags & FilterFile) == 0) {
             return false;
         }
     }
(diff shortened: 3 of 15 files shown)


tomd1990:
I'm created a new PR for this issue, I think merging #257 would have made things less nice in the commit history.

A summary of changes/ discussion
- Created new DriverInfo struct, and implemented IBackend/backenddispatcher::get_drivers function
- local variable is now an array for better handling of strings
- Also use the IBackend flags to filter between File Drivers and Device Drivers

src/modules/roc_sndio/driver_info.cpp
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+#include "roc_sndio/driver_info.h"
+#include "string.h"
+
+namespace roc {
+namespace sndio {
+
+DriverInfo::DriverInfo(const char* driver_name) {
+    strcpy(name, driver_name);
```
gavv: We should do boundary checking here.

src/modules/roc_sndio/driver_info.h
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef ROC_SNDIO_DRIVER_INFO_H_
+#define ROC_SNDIO_DRIVER_INFO_H_
+
+namespace roc {
+namespace sndio {
+
+//! Driver info interface.
+struct DriverInfo {
+    DriverInfo();
```
gavv: It seems that `DriverInfo()` is not implemented and is never used.

src/modules/roc_sndio/ibackend.cpp
```diff
@@ -14,5 +14,20 @@ namespace sndio {
 IBackend::~IBackend() {
 }
 
+bool IBackend::add_driver_uniq(core::Array<DriverInfo>& arr, const char* driver_name) {
```
gavv: IBackend is an "interface class", i.e. all its methods should be pure. Actually, this function does not need to be a member of any class, so I suggest to make it a free function and place it in driver_info.h/driver_info.cpp.

src/modules/roc_sndio/driver_info.h
tomd1990: Array requires me to have a constructor declaration like this, it calls it using T().  That's the only reason why I have this, otherwise it won't compile.

src/modules/roc_sndio/ibackend.cpp
```diff
@@ -14,5 +14,20 @@ namespace sndio {
...
+    for (size_t n = 0; n < arr.size(); n++) {
+        if (strcmp(driver_name, arr[n].name) == 0) {
```
gavv: strcmp() will crash if driver_name is NULL. I suggest to handle this case more gracefully, e.g. insert a panic:

```
bool IBackend::add_driver_uniq(core::Array<DriverInfo>& arr, const char* driver_name) {
    roc_panic_if(driver_name == NULL);
    ...
```

or some other way.

src/modules/roc_sndio/driver_info.h
gavv: Ah, I see. Than I suggest to implement it (and zeroise `name` or `name[0]` in it).

src/tools/roc_recv/main.cpp
```diff
@@ -17,13 +18,16 @@
...
 #include "roc_sndio/backend_dispatcher.h"
+#include "roc_sndio/driver_info.h"
 #include "roc_sndio/pump.h"
 
 #include "roc_recv/cmdline.h"
 
 using namespace roc;
 
 int main(int argc, char** argv) {
+    core::HeapAllocator allocator;
```
gavv: My comment "move it to the top" was a bit misleading, sorry. It should be before `if (args.list_drivers_given) {` but after `crash_handler` and after `core::Logger::instance().set_level`. These two things should be before all other stuff.

src/tools/roc_send/main.cpp
```diff
@@ -40,6 +44,16 @@ int main(int argc, char** argv) {
     core::Logger::instance().set_level(
         LogLevel(core::DefaultLogLevel + args.verbose_given));
 
+    if (args.list_drivers_given) {
+        core::Array<sndio::DriverInfo> driver_list(allocator);
+        sndio::BackendDispatcher::instance().get_drivers(driver_list,
+                                                         sndio::IBackend::ProbeDevice);
+        for (size_t n = 0; n < driver_list.size(); n++) {
```
gavv: I guess we need to print device drivers & file drivers here too?

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -196,5 +196,22 @@ ISource* SoxBackend::open_source(core::IAllocator& allocator,
...
 }
 
+void SoxBackend::get_drivers(core::Array<DriverInfo>& arr, ProbeFlags driver_type) {
+    const sox_format_tab_t* formats = sox_get_format_fns();
+    char const* const* format_names;
+    for (size_t n = 0; formats[n].fn; n++) {
+        sox_format_handler_t const* handler = formats[n].fn();
+        if ((!(handler->flags & SOX_FILE_DEVICE)) & (size_t)(ProbeFile & driver_type) >> 2
+            || ((handler->flags & SOX_FILE_DEVICE) && !(handler->flags & SOX_FILE_PHONY))
+                & (size_t)(ProbeDevice & driver_type) >> 3) {
```
gavv: If I understand correctly, this:

```
(size_t)(ProbeFile & driver_type) >> 2
```

and this:

```
(size_t)(ProbeDevice & driver_type) >> 3
```

both are equal to `1`, which is a bit strange code. Also, it will silently break if the enum definition will change.

We can rewrite it like this:

```
if (((handler->flags & SOX_FILE_DEVICE) != 0) == ((ProbeFile & driver_type) != 0)) || ...) {
   ...
}
```

Arguably, we can also make it more readable by extracting a variable, like:

```
bool match;
if (driver_type & ProbeFile) {
    match = (handler->flags & SOX_FILE_DEVICE);
} else if (driver_type & ProbeDevice) {
    match = (handler->flags & SOX_FILE_DEVICE) && !(handler->flags & SOX_FILE_PHONY);
}
if (match) {
   ...
}
```

(or we can extract a function `bool match(handle)`)

Also, I suggest to rename IBackend::ProbeDevice and other flags to "FilterDevice" or "MatchDevice" or something like this. Previously these flags were used only in probe(), but now they're also used here.

gavv:
Great. I leaved a few more comments. BTW, there is no need to close a PR if you don't like the commit history: you can force-push.

tomd1990:
Thanks for the all of the feedback.  I closed it because I thought if I force pushed, I would lose the relevant discussion. I'll keep things in mind next time.

src/modules/roc_sndio/driver_info.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+DriverInfo::DriverInfo() {
+    for (size_t x = 0; x < 20; x++) {
+        name[x] = '\0';
+    }
+}
+
+DriverInfo::DriverInfo(const char* driver_name) {
+    size_t length = strlen(driver_name) + 1;
+    strncpy(name, driver_name, length);
```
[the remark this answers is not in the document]
gavv: Wow, it seems that GCC 8 reported this overflow (in the travis build).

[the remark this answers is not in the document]
tomd1990: Yeah, I think its because the length, for strncpy, should be tied to the destination and not the source.

src/modules/roc_sndio/driver_info.cpp
```diff
@@ -0,0 +1,48 @@
...
+DriverInfo::DriverInfo() {
+    for (size_t x = 0; x < MaxSize; x++) {
+        name[x] = '\0';
+    }
+}
+
+DriverInfo::DriverInfo(const char* driver_name) {
+    size_t length = strlen(driver_name);
+    if (length > 19) {
+        length = 19;
```
gavv: `19` -> `MaxSize-1`

tomd1990:
So I've made those changes, and in addition I've changed strncpy like so:

`strncpy(name, driver_name, MaxSize);`

The reason being is that it shouldn't depend on source string length,  I believe this is the reason travis was throwing an error.

If MaxSize happens to be greater than the source string, it will not cause overflow.  When it encounters the null char it'll just pad the remaining space with null char, if source string is greater will just take MaxSize - 1 char, and append the null char

gavv:
The code looks good, but I tested it, and these commands fail:

```
$ roc-recv -L
roc-recv: '--source' ('-s') option required
$ roc-send -L
roc-send: '--source' ('-s') option required
```

This is because `option "source"` is marked `required` in `cmdline.ggo` for both tools.

Could you please make them `optional`?

roc-send already has necessary port validation so no further action is needed for it. roc-recv doesn't, so we also need to add a check; it would be reasonable to check that at least one port was specified:

```
if (args.source_given == 0 && args.repair_given == 0) {
    roc_log(LogError, "at least one --source or --repair port should be specified");
    return 1;
}
```

gavv:
Thanks!

tomd1990:
Thank you for all of the feedback!

PR #265, opened 2019-09-28
Implement backtrace printing for non-glibc targets

hrishikeshSuresh:
#249 
Still working on this, just referenced this PR.


SConstruct
```diff
@@ -803,6 +814,12 @@ if 'target_sox' in system_dependecies:
 
     tool_env = conf.Finish()
 
+if 'target_libunwind' in system_dependecies:
+    conf = Configure(tool_env, custom_tests=env.CustomTests)
+    if not conf.CheckLibWithHeaderExt('libunwind', 'libunwind.h', 'C', run=not crosscompile):
```
gavv:
```
if 'target_libunwind' in system_dependecies:
    conf = Configure(tool_env, custom_tests=env.CustomTests)
    if not conf.CheckLibWithHeaderExt('libunwind', 'libunwind.h', 'C', run=not crosscompile):
        env.Die("libunwind not found (see 'config.log' for details)")
    tool_env = conf.Finish()
```

A few more issues with this piece:

* we should use `env`, not `tool_env`, because libunwind, unlike sox, is not used only in tools

* the library file is `libunwind.so`, so the compiler option should be `-lunwind`, so the library name here should be `unwind`, not `libunwind`:

      conf.CheckLibWithHeaderExt('unwind', 'libunwind.h', 'C', ...

* we should run pkg-config to gather necessary include and link directories, like: 

      env.ParsePkgConfig('--cflags --libs libunwind')
      if not conf.CheckLibWithHeaderExt(...

hrishikeshSuresh:
The `-ldl & -lunwind` flags must be passed while compiling the backtrace.cpp codes for target_libunwind & target_bionic respectively. Should these go as args for env.ParsePkgConfig() in `SConstruct `?

hrishikeshSuresh:
> The `-ldl & -lunwind` flags must be passed while compiling the backtrace.cpp codes for target_libunwind & target_bionic respectively. Should these go as args for env.ParsePkgConfig() in `SConstruct `?

Okay, just saw your comment. Understood as to how to solve this.

gavv:
I've tested the Android version. It didn't work out of the box, but worked after fixing compiler options a bit: f6e049cf432b67ae717b7636364c36c52329cf14. This is pushed to develop.

hrishikeshSuresh:
I found a few errors (broken link, missing flags etc.) while downloading dependencies. These could be platform specific. Should I fix them in a separate PR?

gavv:
> I found a few errors (broken link, missing flags etc.) while downloading dependencies. These could be platform specific. Should I fix them in a separate PR?

Yep, would be great.

SConstruct
```diff
@@ -803,6 +814,15 @@ if 'target_sox' in system_dependecies:
...
     tool_env = conf.Finish()
 
+if 'target_libunwind' in system_dependecies:
+    conf = Configure(env, custom_tests=env.CustomTests)
+
+    env.ParsePkgConfig('--cflags --libs libunwind')
+
+    if not conf.CheckLibWithHeaderExt('unwind', 'libunwind.h', 'C', run=not crosscompile):
+        env.Die("libunwind not found (see 'config.log' for details)")
+    tool_env = conf.Finish()
```
gavv: s/tool_env/env/

scripts/3rdparty.py
```diff
@@ -554,6 +554,31 @@ def makeenv(envlist):
...
     install_files('src/.libs/libsox.a', os.path.join(builddir, 'lib'))
+
+elif name == 'libunwind':
+	download(
+		'http://download.savannah.nongnu.org/releases/libunwind/libunwind-%s.tar.gz' % ver,
+		'libunwind-%s.tar.gz' % ver,
+		logfile,
+		vendordir)
+	extract('libunwind-%s.tar.gz' % ver,
+			'src/libunwind-%s' % ver)
```
gavv: `src/libunwind-%s` -> `libunwind-%s`

Also please replace tabs with spaces.

scripts/3rdparty.py
```diff
@@ -554,6 +554,31 @@ def makeenv(envlist):
...
+	os.chdir('src/libunwind-%s' % ver)
+	execute('./configure --host=%s %s %s %s' % (
+        toolchain,
+        makeenv(envlist),
+        makeflags(workdir, toolchain, deplist, variant=variant),
```
gavv: On my system, I also need to add `cflags='-fPIC'` here and `'--disable-minidebuginfo'` bellow (but it was not necessary on alpine for some reason).

src/modules/roc_core/target_libunwind/roc_core/backtrace.cpp
```diff
@@ -0,0 +1,187 @@
+/*
```
gavv: https://github.com/roc-project/roc/pull/249#discussion_r329509412

```
@@ -33,7 +33,7 @@ void safe_strcat(char* buffer, size_t& buffer_size, const char* str) {
     if (len > 0) {
         memcpy(buffer + buffer_size, str, len);
         buffer_size += len;
-        buffer[MaxLen - 1] = '\0';
+        buffer[buffer_size] = '\0';
     }
 }
 
@@ -162,6 +162,7 @@ void backtrace_symbols() {
 
         uint32_to_string((uint32_t)offset, 16, number);
         safe_strcat(buffer, current_buffer_size, number);
+        safe_strcat(buffer, current_buffer_size, "\n");
 
         print_emergency_message(buffer);
     }
```

gavv:
Looks good!

But the addition of `-pthread` and `-lm` to sndfile and cpputest looks suspicious. These libraries should be able to build themselves. Probably it's something wrong with our build script or maybe with your environment. Probably we need to fix it some other way. I suggest to remove these two pieces from the PR and open a separate issue or PR. It would help if you will try to reproduce the problem outside of roc. If it is reproducing outside, this fix is okay. Otherwise, we should investigate why.

Also please rebase your branch on fresh develop branch and squash commits into one or two (one for libunwind + scons and one for bionic), and then we can merge this.

gavv:
It seems you forgot to pull a fresh develop branch from upstream before rebasing on it. It has 5 new commits and there are conflicts.

gavv:
OK, ping me when it will be ready for merge.

hrishikeshSuresh:
I pulled the fresh develop branch but I can't squash the commits properly because I was earlier using another branch `backtrace_print` and if I checkout to an old commit and try to squash commits, I can't merge the detached head with develop because of merge conflicts. Is it very important to squash the commits? I tried doing this via git kraken.

gavv:
Yes, it's quite important for us because our goal is to keep the git history linear and human-readable.

Here is how you can do it in your case:

```
# get a fresh copy of your fork
$ git clone https://github.com/hrishikeshSuresh/roc.git
$ cd roc

# checkout develop branch from your fork
$ git checkout develop

# add upstream repo (call it "upstream") and fetch its branches
$ git remote add upstream https://github.com/roc-project/roc.git
$ git fetch upstream
```

Here is your history now (you can see it using "gitk" command):

image

Then:

```
# find best common ancestor between upstream/develop and your develop
$ git merge-base upstream/develop develop
c7cb95937947a8a02ebe10c7f752d1cd979a2c47

# rebase your develop on that commit
$ git rebase c7cb95937947a8a02ebe10c7f752d1cd979a2c47
```

Here is your history after rebase. As you see, it became linear:

image

Then:

```
# choose your text editor here
$ export GIT_EDITOR=vim

# run an interactive rebase
# we want to rebase develop branch from your fork on the develop branch from upstream
$ git rebase -i upstream/develop
```

The last command will spawn an editor where you can edit the history. You'll need to change "pick" to "squash" in every line except the very first one:

image

And then save your changes and exit the editor. After that the editor will be spawned once more and you'll have to edit the commit message, like this:

image

Then, again, you save your changes and exit the editor.

After that the rebase finishes. Here is your history now:

image

As you can see, it is a single commit on top of upstream/develop.

Now you can push it to your fork:

```
git push -f origin develop
```

gavv:
BTW, personally I'd recommend to avoid using GUI tools to manipulate git history. In my experience, they usually only give an illusion of simplicity and can't really exempt from understanding how git works in any non-trivial case.

gavv:
Also, don't forget to remove `-lpthread` and `-lm` from `3rdparty.py`, as I wrote above. You can do it after rebase and then use `git commit --ammend`.

hrishikeshSuresh:
It worked now. Earlier, I did not make the history linear and I was just squashing all commits. Now I understand how to work with such issues. Thank you!

gavv:
You're welcome. LGTM. Can we merge this?

hrishikeshSuresh:
Yes. Everything is done.

gavv:
Thanks for finishing this!

gavv:
Follow-up commit (formatting): 856ecf6eacbfb748eef4015d1c732950e4ea58e2

Issue #258, opened 2019-09-18
URIs and endpoints


gavv:
### Overview

Currently we have a "port" concept, which has the following notation:

```
protocol:ip:port
```

Now we're adding RTSP support which will operate with URIs like:

```
protocol://host:port/path
```

To keep things unified we're migrating from the concept of "port" to a concept of "endpoint".

Endpoint is something, identified by an URI, to which we can connect and then retrieve a stream from it and/or send a stream to it.

Both senders and receivers will be able to act as servers and clients. The server binds to an endpoint; the client connects to it; this forms a unidirectional or bidirectional stream (depending on the protocol and configuration). Then they start transmitting traffic over this stream.

RTP and FECFRAME endpoints will be used to transmit the traffic directly. RTSP endpoint will be used to establish an RTSP session and internally allocate additional RTP and FECFRAME endpoints for actual data transfer.

The user may either create two RTP and FECFRAME endpoints and connect to them manually, or create a single RTSP endpoint and connect to it, allowing Roc to do the rest.

Examples of new syntax for RTP and FECFRAME:

```
roc-recv -s rtp+rs8m://0.0.0.0:10001 -r rs8m://0.0.0.0:10002
roc-recv -s rtp+rs8m://[::]:10001 -r rs8m://[::]:10002
```

To make things even more unified, we will also use URIs for audio devices, so that this:

```
-d alsa -i card0
```

will become this:

```
-i alsa://card0
```

### Implementation

- [x] Create a new module `roc_address`.

- [x] Move `packet::Address` to `address::SocketAddr` - a class representing a parsed socket address, which consists of: 

  - resolved ip
  - tcp or udp port
  - optional resolved multicast address

- [x] Move protocol definitions from `roc_pipeline` to `roc_address`.

- [x] Add `address::EndpointURI` - a class representing a parsed endpoint URI, which consists of: 

  - protocol (scheme)
  - unresolved hostname
  - port number
  - optional path

- [x] Add `address::IoURI` - a class representing a parsed file/device URI, which consists of: 2

  - driver name
  - device of file name

- [x] Add `address::parse_endpoint_uri()` and `address::format_endpoint_uri()`. The former will replace `pipeline::parse_port()`. In contrast to `parse_port()`, it will not allow empty ip/host. This function does not perform any resolving.

- [x] Add `address::parse_io_uri()` and `address::format_io_uri()`.

- [x] Migrate from `packet::Address` to `address::SocketAddr`.

[...]

PR #268, opened 2019-10-03
Use URIs for devices and files

gavv:
This PR switches roc_sndio and tools from using driver name plus input/output name to using a single URI, both for devices and files. The concept of drivers becomes completely hidden from user.

See #258.


gavv:
Moved to #270.

PR #269, opened 2019-10-05
[wip] add golang bindings

### LINKED ISSUE #248: Golang bindings
Create Go bindings for libroc (our public C API).

The bindings should fit well into the Go type system: e.g. an audio frame is a raw pointer in the C API, but it likely should be a []byte slice in Go bindings.

I think it makes sense to provide a set of conventional wrappers for slices of different types, instead of a single method, e.g.:

```
func (s *Sender) WriteAudioBytes(data []byte) int
func (s *Sender) WriteAudioFloats(data []float) int
```

The bindings can be created manually using cgo or maybe we can partially generate the low-level private part using some bindings generator and then writer high-level public part on top of it manually.

We should also add some unit-tests for the Go bindings (written in Go). We will add them to our continuous integration.

The API can be found in `src/lib/include`. The bindings should be placed in `src/bindings/go` or something like this.

Asalle:
Fixes #248.

This is a work in progress, please add your suggestions and comments! go code was generated using [c-for-go](github.com/xlab/c-for-go/) and will be partially rewritten by hand. 

For now I hardcoded absolute link path for libroc (`roc.go`), I am yet to figure out how to pass a relative path there.

No tests has been added yet.

If you run `go build roc` in the `go` dir, it builds with no errors :)


gavv:
Cool! I'll take a look tomorrow evening or on Monday.

gavv:
c-for-go looks interesting!

Here are my comments on the PR.

### Repo

Initially I though that it would be a good idea to keep the library and the bindings in the same git repo so that they will be versioned together. However, after some thought, this idea appeared not so good:

* The user likely will not use the same git checkout or release tarball for the library and the bindings. The library will be typically installed via OS package manager or manually by the user. The bindings will be typically installed via go get. So it does not matter whether the library and the bindings are in the same repo: the user will have two separate checkouts anyway.

* On the other hand, keeping Go bindings in the separate repo makes sense: they will have they own build system and tests, they will have their own documentation (godoc), and it will be nice to keep the size of the repo downloaded by go get smaller.

So it seems that there are no benefits of adding Go bindings to the same repo and there are some benefits of adding them to a separate repo.

Would you mind if I create a new repo in the roc-project organization (roc-project/roc-go) and and you will create a new PR in that repo?

### Wrappers

The generated bindings are quite low-level. They correspond well to the C API but look foreign in the Go world. So we need to create Go-friendly wrappers around the generated code.

> go code was generated using c-for-go and will be partially rewritten by hand.

Just to be sure: do you mean **re**writing the generated code or adding wrappers on top of it? We should not try to rewrite the generated code because we will loose our changes after re-generating it. We should add wrappers instead.

What we need to change in the wrappers:

* For `Address`, `Context`, `Recevier`, and `Sender` replace free functions with methods, so that this:

      func SenderBind(sender []Sender, address []Address) int32

  will change to this:

      func (s *Sender) Bind(address []Address) int32

* For argument that are single values replace slices with passing by value or passing by pointer (depending on whether the function modifies the argument):

      func (s *Sender) Bind(address *Address) int32

  (BTW, maybe we can configure c-for-go to do it for us?)

* Replace int32 return codes with Go `error`:

      func (s *Sender) Bind(address *Address) error

* Create convenient wrappers for frames of different types, as described in the issue:

      func (s *Sender) WriteAudioBytes(data []byte) int
      func (s *Sender) WriteAudioFloats(data []float) int

* Replace `LogHandler` function with a `Logger` interface, so that the user can pass to us `log.Logger` instance from the standard library or any other compatible logger:

      interface Logger { Print(v ...interface{}) }

### Allocations

Here is a piece of the generated code:

```
func SenderWrite(sender []Sender, frame []Frame) int32 {
	csender, _ := (*C.roc_sender)(unsafe.Pointer((*sliceHeader)(unsafe.Pointer(&sender)).Data)), cgoAllocsUnknown
	cframe, _ := unpackArgSFrame(frame)
	__ret := C.roc_sender_write(csender, cframe)
	packSFrame(frame, cframe)
	__v := (int32)(__ret)
	return __v
}

func unpackArgSFrame(x []Frame) (unpacked *C.roc_frame, allocs *cgoAllocMap) {
	if x == nil {
		return nil, nil
	}
	allocs = new(cgoAllocMap)
	defer runtime.SetFinalizer(&unpacked, func(**C.roc_frame) {
		go allocs.Free()
	})

	len0 := len(x)
	mem0 := allocFrameMemory(len0)
	allocs.Add(mem0)
	h0 := &sliceHeader{
		Data: mem0,
		Cap:  len0,
		Len:  len0,
	}
	v0 := *(*[]C.roc_frame)(unsafe.Pointer(h0))
	for i0 := range x {
		allocs0 := new(cgoAllocMap)
		v0[i0], allocs0 = x[i0].PassValue()
		allocs.Borrow(allocs0)
	}
	h := (*sliceHeader)(unsafe.Pointer(&v0))
	unpacked = (*C.roc_frame)(h.Data)
	return
}

func allocFrameMemory(n int) unsafe.Pointer {
	mem, err := C.calloc(C.size_t(n), (C.size_t)(sizeOfFrameValue))
	if err != nil {
		panic("memory alloc error: " + err.Error())
	}
	return mem
}
```

It seems that for every frame, it will perform at least one allocation on the C heap and a few allocations on the Go heap.

Writing and reading frames (roc_sender_write and roc_receiver_read) are hot paths. Ideally we should avoid allocations in these functions.

Both functions don't use the frame memory after the function returns. So although the frame memory is managed by Go GC, it should be safe to pass the raw pointer to that memory to these functions. So theoretically doing it without allocations is quite feasible.

If we can configure c-for-go to do this for us, it would be great. Otherwise, we can configure c-for-go to disable generating bindings for those two functions and write them by hand.

Asalle:
> Would you mind if I create a new repo in the roc-project organization (roc-project/roc-go) and and you will create a new PR in that repo?
Please go ahead, I'll issue a pr there in the next few days.

> We should add wrappers instead.
I read your arguments and I agree with you. We should add wrappers.

> Wrappers:
:+1: Will be done

> Allocations:
I'll see what can be done.

Thank you!

gavv:
I've created a repo: https://github.com/roc-project/roc-go

I've added libroc headers to the repo. This will allow the user to run `go get` without having libroc headers on the system. Only libroc shared library will be required.

(Currently we have only `scons install` which installs both headers and the shared library. But in future roc may be packaged by distros and its headers may be packaged separately in a `-dev` or `-devel` package and I think we should not require it to be installed.)

I've also added a makefile with a target to download these headers (`make include`).

gavv:
About paths:

* Since we have a copy libroc headers, the include search path should be "./include".

* The library search path (for libroc.so) should be determined by pkg-config, ideally. However so far roc doesn't provide a `.pc` file.

  For now I suggest to assume that the user installed roc using `scons install` and libroc.so is available in the default library path (e.g. in `/usr/lib`) so we don't need to list it explicitly. Later we will add pkg-config support.

Asalle:
:+1: 
I'll take a look on the weekend. Should I close this pr?

gavv:
> I'll take a look on the weekend.

Great!

> Should I close this pr? And open a new one in the separate repo?

Yes, please.

Asalle:
Awesome!

Asalle:
@gavv here is the first pr for the go bindings https://github.com/roc-project/roc-go/pull/1
Sorry, I didn't have any other way to contact you. Can I somehow add you to reviews in the future?

gavv:
Excellent, I'll review it & answer the questions tomorrow.

> Sorry, I didn't have any other way to contact you. Can I somehow add you to reviews in the future?

It's okay, I'm subscribed to that repo and get notifications on new PRs so they won't get lost.

gavv:
@Asalle I guess we can close this one?

Asalle:
Yeah, I have a lot things to finish, but it can be done in the next pr.

gavv:
Moved to https://github.com/roc-project/roc-go/pull/1

Issue #272, opened 2019-10-07
Colored logs


gavv:
It would be nice to add colors to our log when we're running in a TTY.

I'd add colors only to error and informational messages and only to the "level" and "message" parts of the line. I'd use bold red for errors and bold blue for info messages.

Something like this:

![1111](https://user-images.githubusercontent.com/8436629/66334183-330ac280-e941-11e9-9ff5-a9b22b932d3e.png)

Steps:

* Add `colors.h` and `colors.cpp` to `src/modules/roc_core/target_posix/roc_core`. We will place the POSIX-specific part of the color-formatting code here.

* Add two new functions to that files:

      bool colors_available();

      bool format_colored(LogLevel level, const char* str,
                          char* buf, size_t bufsz);

  The first should check whether stderr is a TTY. The second should wrap `str` with the ANSI color sequences and write result to `buf`. The color should depend on the `level` argument. It should also check for buffer overflows and return an error if the buffer is too small.

* Use these functions in `core::Logger::print()`. If colors are available, use `format_colored()` for `level` and `message` parts of the line.

* Add `core::Logger::set_colors()` method accepting a enum defining three coloring modes: `ColorsAuto` (use colors if stderr is a TTY), `ColorAlways` (always use colors), `ColorsNever` (never use colors).

  `core::Logger::print()` should take this setting into account.

* Add `--color=auto|always|never` command-line option to all tools (roc-recv, roc-send, roc-conv). Pass the configured mode to `core::Logger::set_colors()`. The default mode should be "auto".

secondW4ve:
Hi, can you tell where I can watch your solution for errors and informational messages? I am not completely understand how it should work, so I would like to watch how it have to be done.

gavv:
Hi,

I have no solution. If you're talking about the screenshot above, I just hardcoded color codes into the logger and a few messages to see how it will look like.

duynhan39:
Hi,

Is there any quick way to run the toolkit and see error message?

gavv:
@duynhan39 For example, you can pass a nonexistent file to roc-send, just like in the example above. See details here: https://roc-project.github.io/roc/docs/manuals/roc_send.html

dshil:
Assigned to @kcy1019

dshil:
See #273

PR #273, opened 2019-10-09
Colored logs (#272)

kcy1019:
screenshot

This PR adds colors to `stderr` logging while attached to a tty by adding:
- `--color=auto|never|always` options to roc tools (`roc_conv`, `roc_recv`, `roc_send`)
- `roc::core::colors_available()`, `roc::core::format_colored()`, `roc::core::Logger::set_colors()` functions
See #272 for more details.

Thanks in advance for advices :)

### DIFF
--- a/src/modules/roc_core/target_posix/roc_core/colors.cpp
+++ b/src/modules/roc_core/target_posix/roc_core/colors.cpp
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "roc_core/colors.h"
+
+//! ANSI Color Codes.
+#define COLOR_NONE ""
+#define COLOR_START "\033["
+#define COLOR_END "m"
+#define COLOR_RESET COLOR_START "0" COLOR_END
+#define COLOR_SEPARATOR ";"
+#define COLOR_BOLD "1"
+#define COLOR_RED "31"
+#define COLOR_BLUE "34"
+
+namespace roc {
+namespace core {
+
+namespace {
+char const* colors_to_levels(LogLevel level) {
+    char const* color = COLOR_NONE;
+    switch (level) {
+    case LogNone: //!< No color for Loglevel None.
+        break;
+    case LogError: //!< Bold Red for Loglevel Error.
+        color = COLOR_START COLOR_BOLD COLOR_SEPARATOR COLOR_RED COLOR_END;
+        break;
+    case LogInfo: //!< Bold Blue for Loglevel Info.
+        color = COLOR_START COLOR_BOLD COLOR_SEPARATOR COLOR_BLUE COLOR_END;
+        break;
+    case LogDebug: //!< No color for Loglevel Debug.
+        break;
+    case LogTrace: //!< No color for Loglevel Trace.
+        break;
+    }
+    return color;
+}
+} // namespace
+
+bool colors_available() {
+    char* term = getenv("TERM");
+    return isatty(STDERR_FILENO) && term && strncmp("dumb", term, 4) != 0;
+}
+
+bool format_colored(LogLevel level, const char* str, char* buf, size_t bufsz) {
+    roc_panic_if_not(str);
+    roc_panic_if_not(buf);
+    int printed =
+        snprintf(buf, bufsz, "%s%s%s", colors_to_levels(level), str, COLOR_RESET);
+    return printed > 0 && (size_t)printed < bufsz;
+}
+
+} // namespace core
+} // namespace roc
--- a/src/modules/roc_core/target_stdio/roc_core/log.cpp
+++ b/src/modules/roc_core/target_stdio/roc_core/log.cpp
@@ -9,6 +9,7 @@
 #include <stdarg.h>
 #include <stdio.h>
 
+#include "roc_core/colors.h"
 #include "roc_core/format_time.h"
 #include "roc_core/log.h"
 
@@ -17,7 +18,8 @@ namespace core {
 
 Logger::Logger()
     : level_(DefaultLogLevel)
-    , handler_(NULL) {
+    , handler_(NULL)
+    , colors_(DefaultColorsMode) {
 }
 
 LogLevel Logger::level() {
@@ -44,6 +46,12 @@ void Logger::set_handler(LogHandler handler) {
     Mutex::Lock lock(mutex_);
 
     handler_ = handler;
+}
+
+void Logger::set_colors(ColorsMode colors) {
+    Mutex::Lock lock(mutex_);
+
+    colors_ = colors;
 }
 
 void Logger::print(const char* module, LogLevel level, const char* format, ...) {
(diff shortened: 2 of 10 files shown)


gavv:
Great! I'll take a look tomorrow.

dshil:
Hi @kcy1019, thanks for the PR,

Could you please run `scons -Q fmt` to use the existing formatting style?

dshil:
@kcy1019, I checked out to your branch, built roc and try to see colored logs but:

image

src/modules/roc_core/target_posix/roc_core/colors.cpp
```diff
@@ -0,0 +1,66 @@
...
+#define COLOR_START      "\033["
+#define COLOR_END        "m"
+#define COLOR_RESET      COLOR_START COLOR_END
+#define COLOR_SEPARATOR  ";"
+#define COLOR_BOLD       "1"
+#define COLOR_RED        "31"
+#define COLOR_BLUE       "34"
+
+
+static char const* colors_to_levels(LogLevel level) {
```
dshil: You use anonymous namespace instead of static here.

src/modules/roc_core/target_posix/roc_core/colors.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "roc_core/colors.h"
+
+namespace roc {
+namespace core {
+
+//! ANSI Color Codes.
+#define COLOR_NONE       ""
+#define COLOR_START      "\033["
+#define COLOR_END        "m"
+#define COLOR_RESET      COLOR_START COLOR_END
```
dshil: It seems that a [valid sequence](https://en.wikipedia.org/wiki/ANSI_escape_code) for resetting all attributes to their defaults is `\033[0m`.

src/modules/roc_core/target_posix/roc_core/colors.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+bool colors_available() {
+    char * term = getenv("TERM");
+    return isatty(STDERR_FILENO) && \
+           term && strncmp("dumb", term, 4) != 0;
+}
+
+bool format_colored(LogLevel level, const char* str,
+                    char* buf, size_t bufsz) {
```
dshil: Lets add some constraints to the function:

```cpp
roc_panic_if_not(str);
roc_panic_if_not(buf);
```

to ensure we won't receive UB.

src/modules/roc_core/target_posix/roc_core/colors.cpp
```diff
@@ -0,0 +1,66 @@
...
+    int printed = snprintf(buf, bufsz, "%s%s%s",
+                           colors_to_levels(level), str,
```
dshil: Probably we can use less general approach here:

```cpp
bool format_colored(LogLevel level, const char* str, char* buf, size_t bufsz) {
    int ret = 0;

    if (level == LogError) {
        ret = snprintf(buf, bufsz, "%s%s%s", "\033[1;31m", str, "\033[0m");
    } else if (level == LogInfo) {
        ret = snprintf(buf, bufsz, "%s%s%s", "\033[1;34m", str, "\033[0m");
    }

   // other stuff
}
```

It is more cryptic but more visual (@gavv ?)

src/modules/roc_core/target_posix/roc_core/colors.cpp
dshil: Hm, it seems that `\033[m` also works for me.

src/modules/roc_core/target_stdio/roc_core/log.cpp
```diff
@@ -87,6 +95,22 @@ void Logger::print(const char* module, LogLevel level, const char* format, ...)
             break;
         }
 
+        bool colors_enabled = colors_ != ColorsNever &&
+                              (colors_ == ColorsAlways || colors_available());
+        if (colors_enabled) {
+            char colored_level_str[16] = {};
+            char colored_message[256] = {};
+
+            if (format_colored(level_, level_str, colored_level_str,
```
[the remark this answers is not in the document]
kcy1019: Thanks, I could reproduce the issue by increasing log verbosity :)

src/modules/roc_core/target_stdio/roc_core/log.h
```diff
@@ -36,11 +36,21 @@ enum LogLevel {
     LogTrace  //!< Debug message (extra verbosity).
 };
 
+//! Colors mode.
+enum ColorsMode {
```
gavv: Could you please move this inside the `core` namespace? We usually don't add things to the root `roc` namespace, LogLevel is one of the few exceptions.

src/modules/roc_core/target_posix/roc_core/colors.cpp
```diff
@@ -0,0 +1,63 @@
...
+    case LogTrace: //!< No color for Loglevel Trace.
+        break;
+    }
+    return color;
+}
+} // namespace
+
+bool colors_available() {
+    char* term = getenv("TERM");
+    return isatty(STDERR_FILENO) && term && strncmp("dumb", term, 4) != 0;
```
gavv: Could you explain why do we need to check `TERM`?

There are two concerns here:

* We're performing a search in environment for every log message. Not very nice.

* What is worse, our library is now calling getenv() from its threads (when it logs something). If the user code outside of our library will call setenv() from another thread, we'll get a race and probably a crash.

I'd prefer to avoid calling getenv() if it's not necessary. If it's necessary, I'd prefer to call it only in tools and only once. E.g. we could add a function colors_init() and call it from main().

src/modules/roc_core/target_posix/roc_core/colors.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+//! ANSI Color Codes.
+#define COLOR_NONE ""
+#define COLOR_START "\033["
+#define COLOR_END "m"
+#define COLOR_RESET COLOR_START "0" COLOR_END
+#define COLOR_SEPARATOR ";"
+#define COLOR_BOLD "1"
+#define COLOR_RED "31"
+#define COLOR_BLUE "34"
```
gavv: Nitpick: could you please move these defines upper, between the include block and the namespace block? (according to our code-style; defines are global so it's a bit strange to have them in the namespace section).

src/modules/roc_core/target_posix/roc_core/colors.cpp
kcy1019: Agree with your points :)
The intention behind $TERM check is actually about convention – many POSIX command line tools supports no-colorized outputs when $TERM=dumb; e.g. coreutils and git.
I firstly thought about adding static variables to the `colors_available()`, but as you suggested, calling it only from tools makes more sense. Thanks!

gavv: Ah, I see. If it's a common convention, it would not be very good to ignore it. Maybe for now it would be better to disable coloring at all when roc is used as a library.

Otherwise we should provide a C API to forcibly enable and disable colors and probably to process the environment in a thread-safe manner, e.g. some global initialization function that is allowed to call getenv(). This looks like an overkill for this feature.

BTW, current API allows to provide a custom log handler, so that the user can implement colored output on top of roc, if needed.

I think we can replace ColorsAlways/ColorsNever/ColorsAuto with just two modes ColorsEnabled/ColorsDisabled (or with a boolean) and by default use ColorsDisabled. Command-line will continue to use always|never|auto and will set ColorsEnabled/ColorsDisabled depending on the command-line option and colors_available() function.

src/tools/roc_conv/main.cpp
```diff
@@ -37,6 +38,23 @@ int main(int argc, char** argv) {
     core::Logger::instance().set_level(
         LogLevel(core::DefaultLogLevel + args.verbose_given));
 
+    switch ((unsigned)args.color_arg) {
+    case color_arg_auto:
+        if (core::colors_available())
+            core::Logger::instance().set_colors(core::ColorsMode(core::ColorsEnabled));
```
gavv: Nitpick: we always use curly braces `{}` for conditions and loops.

src/tools/roc_conv/main.cpp
```diff
@@ -37,6 +38,23 @@ int main(int argc, char** argv) {
...
+        break;
+
+    case color_arg_always:
+        core::Logger::instance().set_colors(core::ColorsMode(core::ColorsEnabled));
+        break;
+
+    case color_arg_never:
+        break;
```
gavv: Nitpick: I suggest to explicitly set the color mode to Disabled here to avoid relying on the specific default value.

gavv:
After fixing the last issues, could you please squash your commits? It may be a single commit or e.g. 3 commits (color functions, logger, tools).

gavv:
LGTM!

dshil:
@kcy1019, Thanks!

kcy1019:
@dshil @gavv Thank you for reviewing the patch. I really appreciate your thoughtful comments :)

Issue #164, opened 2019-02-27
Configure internal frame size in nanoseconds


gavv:
All durations available via public API are now configured in nanoseconds instead of number of samples (packet length, latency, timeouts, etc.)

The only duration parameter that is still configured in number of samples is the `internal_frame_size` (it appears three times in `roc_pipeline/config.h`).

This is not critical since it's not visible in the public API for now. But it would be nice to configure it in nanoseconds too because:

* it will make this parameter independent of the sample rate and channel set
* this parameter affects latency and it is easier to understand the overall latency if all its components are calculated in the same encoding-independent units (nanoseconds in our case)
* this parameter may be added to public API later, and all such parameters in the API should be in nanoseconds

So what we need to do:

* Change type of `internal_frame_size` from `size_t` to `core::nanoseconds_t`.

* Find all components to which this parameter is passed from roc_pipeline. Add sample rate and channel mask parameters to constructors of all these components. Then, inside those constructors, calculate the actual number of samples using `timestamp_from_ns()` function, something like: `packet::timestamp_from_ns(internal_frame_size, sample_rate) * packet::num_channels(channels)`.

* In roc-recv, roc-send, and roc-conv command-line tools, use the string type for the `--frame-size` option and use `parse_duration()` function to parse this string into a nanoseconds value.

hrishikeshSuresh:
If no one else is working on this, can I pick this issue?

gavv:
Sure!

gavv:
PR is merged. So we have finally finished migration from samples to nanoseconds, which is cool :) Thanks @hrishikeshSuresh!

PR #274, opened 2019-10-15
configure internal frame size in nanoseconds

hrishikeshSuresh:
Referring issue #164 
I ran 
`$ roc-recv -vv -s rtp+rs8m::10001 -r rs8m::10002 --frame-size="100"`
and 
`$ roc-send -vv -s rtp+rs8m:127.0.0.1:10001 -r rs8m:127.0.0.1:10002 -i ./file.wav --frame-size="100"`
The audio was streamed successfully and did not encounter any issues.


gavv:
BTW, the build is failing.

src/modules/roc_pipeline/config.h
```diff
@@ -83,7 +83,7 @@ struct SenderConfig {
     packet::channel_mask_t input_channels;
 
     //! Number of samples for internal frames.
-    size_t internal_frame_size;
+    core::nanoseconds_t internal_frame_size;
```
gavv: Here and below, the comments should be updated accordingly, e.g.:

```
//! Duration of the internal frames, in nanoseconds.
```

src/tools/roc_recv/cmdline.ggo
```diff
@@ -43,7 +43,7 @@ section "Options"
         int optional
 
     option "frame-size" - "Internal frame size, number of samples"
-        int optional
+        string optional
```
gavv: We should also provide typestr:

```
typestr="TIME"
```

The `TIME` format is described below in the help text:

```
TIME is an integer number with a suffix, e.g.:
  123ns; 123us; 123ms; 123s; 123m; 123h;
```

BTW it's not present in roc-conv help yet, we should add it too.

src/tools/roc_recv/main.cpp
```diff
@@ -85,7 +85,7 @@ int main(int argc, char** argv) {
             roc_log(LogError, "invalid --frame-size: should be > 0");
             return 1;
         }
-        config.common.internal_frame_size = (size_t)args.frame_size_arg;
+        roc::core::parse_duration(args.frame_size_arg, config.common.internal_frame_size);
```
gavv: We should check for errors here.

src/modules/roc_pipeline/receiver.cpp
```diff
@@ -35,8 +35,11 @@ Receiver::Receiver(const ReceiverConfig& config,
     , timestamp_(0)
     , num_channels_(packet::num_channels(config.common.output_channels))
     , active_cond_(control_mutex_) {
-    mixer_.reset(new (allocator_)
-                     audio::Mixer(sample_buffer_pool, config.common.internal_frame_size),
+    mixer_.reset(new (allocator_) audio::Mixer(
+                     sample_buffer_pool,
+                     packet::timestamp_from_ns(config.common.internal_frame_size,
+                                               config.common.output_sample_rate)
+                         * packet::num_channels(config.common.output_channels)),
```
gavv: We're trying to keep roc_pipeline code very high-level. Usually it only constructs the pipeline but doesn't perform any calculations, low-level manipulations with parameters, etc.

So, as I mentioned in the issue, could you move the calculations inside the components constructors? Here and in all other places. So that, in particular, you'll pass config.common.internal_frame_size to Mixer and it will call timestamp_from_ns and num_channels internally.

Also, we should perform validation of the result. If the calculations result is zero, the corresponding component should log an error and switch to an an fail to construct. See, for example, audio::Watchdog.

This is how we deal with validation:

```
Mixer::Mixer(..., size_t internal_frame_size, ...)
    : valid_(false)
    , frame_size_(timestamp_from_ns(...) * num_channels(...))
{
    if (frame_size_ == 0) {
        roc_log(LogError, ...);
        return;
    }

    ...

    valid_ = true;
}
```

Then pipeline::Receiver will call mixer.valid() and see that mixer was failed to construct.

src/tools/roc_recv/main.cpp
```diff
@@ -249,9 +251,12 @@ int main(int argc, char** argv) {
...
 
-    sndio::Pump pump(
-        sample_buffer_pool, receiver, *sink, config.common.internal_frame_size,
-        args.oneshot_flag ? sndio::Pump::ModeOneshot : sndio::Pump::ModePermanent);
+    sndio::Pump pump(sample_buffer_pool, receiver, *sink,
+                     packet::timestamp_from_ns(config.common.internal_frame_size,
+                                               config.common.output_sample_rate)
+                         * packet::num_channels(config.common.output_channels),
+                     args.oneshot_flag ? sndio::Pump::ModeOneshot
+                                       : sndio::Pump::ModePermanent);
```
gavv: As with roc_pipeline, could you move these calculations inside sndio::Pump constructor? And add validation as well.

src/modules/roc_pipeline/receiver.cpp
hrishikeshSuresh: So I'm thinking of modifying `Mixer` constructor by making it like 
```suggestion
Mixer(core::BufferPool<sample_t>& pool, 
          core::nanoseconds_t frame_size, 
          size_t sample_rate, 
          packet::channel_mask_t number_of_channels); 
```
But won't this also mean that I have to modify all calls made to this constructor `Mixer`, for example, in `src/modules/roc_audio/ text_mixer.cpp`, which just calls 
`Mixer mixer(buffer_pool, MaxSz);`
But this can be solved by making` size_t sample_rate, packet::channel_mask_t number_of_channels` optional by assigning them all to zero, making them default arguments.
So the final constructor would look like
```suggestion
Mixer(core::BufferPool<sample_t>& pool, 
         core::nanoseconds_t frame_size, 
         size_t sample_rate = 0, 
         packet::channel_mask_t number_of_channels = 0);
```
Can I do it like this?

gavv: > packet::channel_mask_t number_of_channels

Note: it should be a channel mask, not number of channels. Then you can convert it to the number of channels using packet::num_channels().

> But won't this also mean that I have to modify all calls made to this constructor Mixer

Yep.

> But this can be solved by making size_t sample_rate, packet::channel_mask_t number_of_channels optional by assigning them all to zero, making them default arguments.

Formerly we used this approach, but then rejected it. The problem is that it was quite error-prone: it was easy to forget to pass optional arguments and create an inconsistency between pipeline elements (e.g. create two elements with different channel masks). So now we don't use default arguments for constructors and always explicitly provide all arguments.

src/tools/roc_recv/cmdline.ggo
hrishikeshSuresh: done

src/modules/roc_pipeline/config.h
hrishikeshSuresh: done.

src/tools/roc_recv/main.cpp
hrishikeshSuresh: done.

hrishikeshSuresh: done.

hrishikeshSuresh:
I just noticed that I didn't update src/modules/mixer/test_mixer.cpp. I will work on it now.

src/modules/roc_pipeline/receiver.cpp
hrishikeshSuresh: done.

hrishikeshSuresh:
I think I might have made a slight mistake here. `frame_size` is a parameter for `ResamplerWriter`. Should it be of type `core::nanoseconds_t` or `size_t` ? Because in the above commits, I thought it should have been changed to `core::nanoseconds_t` and changed `ResamplerWriter`.

src/tools/roc_recv/cmdline.ggo
hrishikeshSuresh: done.

gavv:
I guess it should be nanoseconds_t. BTW I'd change the name from frame_size to frame_length to make it clear that it is something other than just the number of samples.

hrishikeshSuresh:
I will put up a message here once I fix all these errors.

src/modules/roc_audio/resampler_reader.h
```diff
@@ -69,7 +71,7 @@ class ResamplerReader : public IReader, public core::NonCopyable<> {
     IReader& reader_;
 
     core::Slice<sample_t> frames_[3];
-    const size_t frame_size_;
+    core::nanoseconds_t frame_length_;
```
gavv: > Should it be of type core::nanoseconds_t or size_t ? Because in the above commits, I thought it should have been changed to core::nanoseconds_t and changed ResamplerReader.

Hm, I thought you were asking about constructor parameter, but if you were asking about the class field, sorry, I probably misled you.

The constructor parameter should be `nanoseconds_t frame_length` because it accepts the frame duration in time units:

```
    ResamplerReader(...
                    packet::channel_mask_t channel_mask,
                    core::nanoseconds_t frame_length,
                    size_t sample_rate);
```

However the class field likely should be `size_t frame_size_` because it hold the frame size in samples.

hrishikeshSuresh: Yeah that's alright. I will fix them.

hrishikeshSuresh:
You can review all the changes made. I think I fixed almost all errors I had made earlier.

src/tools/roc_recv/cmdline.ggo
```diff
@@ -42,8 +42,8 @@ section "Options"
     option "packet-limit" - "Maximum packet size, in bytes"
         int optional
 
-    option "frame-size" - "Internal frame size, number of samples"
-        int optional
+    option "frame-size" - "Duration of the internal frame size, in nanoseconds"
+        typestr="TIME" string optional
```
gavv: `Duration of the internal frame size` -> `Duration of the internal frames`

src/modules/roc_audio/mixer.cpp
```diff
@@ -28,21 +28,31 @@ sample_t clamp(const sample_t x) {
...
-Mixer::Mixer(core::BufferPool<sample_t>& pool, size_t frame_size)
-    : valid_(false) {
-    roc_log(LogDebug, "mixer: initializing: frame_size=%lu", (unsigned long)frame_size);
+Mixer::Mixer(core::BufferPool<sample_t>& pool, 
+             core::nanoseconds_t frame_length, 
+             size_t sample_rate, 
+             packet::channel_mask_t channel_mask)
+    : valid_(false)
+    , frame_size_((unsigned long)packet::timestamp_from_ns(frame_length, sample_rate) 
+                  * packet::num_channels(channel_mask)) {
```
gavv: Shouldn't it be `size_t` instead of `unsigned long`? Here and in other places.

src/modules/roc_audio/resampler_reader.cpp
```diff
@@ -32,6 +34,9 @@ ResamplerReader::ResamplerReader(IReader& reader,
     if (!init_frames_(buffer_pool)) {
         return;
     }
+    if(frame_size_ == 0) {
+        return;
+    }
```
gavv: It's better to check frame_size_ before calling init_frames_, because init_frames_ uses frame_size_.

src/modules/roc_audio/resampler_writer.cpp
```diff
@@ -71,7 +73,7 @@ void ResamplerWriter::write(Frame& input) {
             frame2_data[frame_pos_ - frame_size_ * 2] = input_data[input_pos];
         }
 
-        // All three slices are full, resampling frame_size_ samples.
+        // All three slices are full, resampling (unsigned long) frame_size_ samples.
```
gavv: This is an accidental change I guess.

src/modules/roc_audio/resampler_writer.cpp
```diff
@@ -20,11 +20,13 @@ ResamplerWriter::ResamplerWriter(IWriter& writer,
                                  core::IAllocator& allocator,
                                  const ResamplerConfig& config,
                                  packet::channel_mask_t channels,
-                                 size_t frame_size)
-    : resampler_(allocator, config, channels, frame_size)
+                                 core::nanoseconds_t frame_length,
+                                 size_t sample_rate)
+    : frame_size_((unsigned long)packet::timestamp_from_ns(frame_length, sample_rate) 
+                  * packet::num_channels(channels))
```
gavv: It seems we forgot to validate frame_size_ in ResamplerWriter ctor?

src/modules/roc_audio/mixer.h
```diff
@@ -67,6 +74,7 @@ class Mixer : public IReader, public core::NonCopyable<> {
     core::Slice<sample_t> temp_buf_;
 
     bool valid_;
+    const size_t frame_size_;
```
gavv: It seems that we need frame_size_ only in ctor. Thus we can make it a local variable in ctor and don't add a new field. Probably the same is true for some other classes too.

src/modules/roc_audio/resampler_reader.h
```diff
@@ -64,12 +66,12 @@ class ResamplerReader : public IReader, public core::NonCopyable<> {
...
     bool init_frames_(core::BufferPool<sample_t>&);
     void renew_frames_();
+    const size_t frame_size_;
 
     Resampler resampler_;
     IReader& reader_;
 
     core::Slice<sample_t> frames_[3];
-    const size_t frame_size_;
     bool frames_empty_;
```
gavv: This frame* fields were grouped together, I suggest to keep this as is. The same is true for some other classes.

src/tools/roc_conv/main.cpp
```diff
@@ -69,22 +70,26 @@ int main(int argc, char** argv) {
...
 
     if (args.frame_size_given) {
-        if (args.frame_size_arg <= 0) {
-            roc_log(LogError, "invalid --frame-size: should be > 0");
+        if (sizeof(args.frame_size_arg) <= 0) {
+            roc_log(LogError, "invalid --frame-size: frame size unspecified");
             return 1;
         }
-        config.internal_frame_size = (size_t)args.frame_size_arg;
+        roc::core::parse_duration(args.frame_size_arg, config.internal_frame_size);
```
gavv: We should check the return value of parse_duration(). Also, `roc::` is not necessary here, we have `using namespace roc` directive above.

src/tools/roc_conv/main.cpp
```diff
@@ -69,22 +70,26 @@ int main(int argc, char** argv) {
     pipeline::ConverterConfig config;
 
     if (args.frame_size_given) {
-        if (args.frame_size_arg <= 0) {
-            roc_log(LogError, "invalid --frame-size: should be > 0");
+        if (sizeof(args.frame_size_arg) <= 0) {
+            roc_log(LogError, "invalid --frame-size: frame size unspecified");
             return 1;
         }
```
gavv: `sizeof(args.frame_size_arg)` will return `sizeof(char*)` i.e. size of the pointer, e.g. 8 on 64-bit machines. If you want to check that the string is non-empty, you should use strlen().

However there is no need to check it here because parse_duration() will do all necessary checks. We just need to check its return value.

src/tools/roc_conv/main.cpp
```diff
@@ -69,22 +70,26 @@ int main(int argc, char** argv) {
...
-        config.internal_frame_size = (size_t)args.frame_size_arg;
+        roc::core::parse_duration(args.frame_size_arg, config.internal_frame_size);
+        if (config.internal_frame_size <= 0) {
+            roc_log(LogError, "invalid --frame-size: should be > 0");
+            return 1; 
+        }
     }
 
-    sndio::BackendDispatcher::instance().set_frame_size(config.internal_frame_size);
+    sndio::BackendDispatcher::instance().set_frame_size((unsigned long) config.internal_frame_size);
```
hrishikeshSuresh: Here, `config.internal_frame_size` is of `core::nanoseconds_t` type. So shouldn't I convert it into frame size by using sample rate & number of channels instead explicit conversion using `size_t`?

hrishikeshSuresh: and again, I think `internal_frame_size` should be changed to `internal_frame_length`. so probably we can add another attribute for frame size to this class or we can use a local variable for this.

gavv: Good catch. Yes, we can't just cast it. And yes, we should rename it to internal frame length. I think it's better to rename both config field and command-line options.

We could change the method signature to set_frame_size(frame_length, sample_rate, channel_mask) and perform the conversion in its implementation.

src/tools/roc_conv/main.cpp
```diff
@@ -69,22 +70,26 @@ int main(int argc, char** argv) {
...
 
-    core::BufferPool<audio::sample_t> pool(allocator, config.internal_frame_size,
+    core::BufferPool<audio::sample_t> pool(allocator, (unsigned long) config.internal_frame_size,
                                            args.poisoning_flag);
 
     sndio::Config source_config;
     source_config.channels = config.input_channels;
     source_config.sample_rate = 0;
-    source_config.frame_size = config.internal_frame_size;
+    source_config.frame_size = (unsigned long) config.internal_frame_size;
```
hrishikeshSuresh: There are few more places like this where this conversion from frame length to frame size. I think its better if we get a function for this operation. How do we go about this?

src/tools/roc_conv/main.cpp
hrishikeshSuresh: In `src/modules/roc_pipeline/config.h`, it seems that we have to define a default frame length variable? I calculated it by  `( defaultFrameSize * defaultSampleRate ) / defaultChannelMask` and the value seems to be 9,408,000. Is this value correct for default frame length?

gavv: Since frame length is measured in nanoseconds now, we need something like:

```
//! Default frame length.
const core::nanoseconds_t DefaultFrameLength = 7 * core::Millisecond;
```

gavv: (it will be 7'000'000)

src/tools/roc_recv/cmdline.ggo
hrishikeshSuresh: done.

src/modules/roc_audio/mixer.cpp
hrishikeshSuresh: done.

src/modules/roc_audio/resampler_reader.cpp
hrishikeshSuresh: done.

src/modules/roc_audio/resampler_writer.cpp
hrishikeshSuresh: removed.

src/modules/roc_audio/mixer.h
hrishikeshSuresh: done.

src/modules/roc_audio/resampler_reader.h
hrishikeshSuresh: done.

src/tools/roc_conv/main.cpp
hrishikeshSuresh: done.

src/tools/roc_conv/main.cpp
hrishikeshSuresh: done.

src/tools/roc_conv/main.cpp
hrishikeshSuresh: done.

src/tools/roc_conv/main.cpp
hrishikeshSuresh: added ns_to_size().

gavv:
A bit busy currently, will take a look in a few days. Could you please fix tests? (see travis build).

gavv:
Also, please rebase on fresh develop.

hrishikeshSuresh:
> A bit busy currently, will take a look in a few days. Could you please fix tests? (see travis build).

Yes, I'm trying to fix it. It looks like CHECK(mixer.valid()) is failing. Is it because mixer.valid() is returning false?

gavv:
> Is it because mixer.valid() is returning false?

Yep.

gavv:
Try to run `roc-test-audio -v` or `roc-test-audio -v -g mixer -n <testname>`.

src/modules/roc_sndio/pump.cpp
```diff
@@ -15,16 +15,24 @@ namespace sndio {
...
+    size_t frame_size = packet::ns_to_size(frame_length, sample_rate, ch_mask);
+    if (frame_size == 0) {
+        roc_log(LogError, "pump: frame size cannot be 0");
+        return;
+    }
+
     if (buffer_pool.buffer_size() < frame_size) {
         roc_log(LogError, "pump: buffer size is too small: required=%lu actual=%lu",
-                (unsigned long)frame_size, (unsigned long)buffer_pool.buffer_size());
+                frame_size, (unsigned long)buffer_pool.buffer_size());
```
gavv: `(unsigned long)` is required in roc_log() due of `%lu`.

src/modules/roc_audio/mixer.cpp
```diff
@@ -28,9 +28,18 @@ sample_t clamp(const sample_t x) {
...
 
-Mixer::Mixer(core::BufferPool<sample_t>& pool, size_t frame_size)
+Mixer::Mixer(core::BufferPool<sample_t>& pool, 
+             core::nanoseconds_t frame_length, 
+             size_t sample_rate, 
+             packet::channel_mask_t channel_mask)
     : valid_(false) {
-    roc_log(LogDebug, "mixer: initializing: frame_size=%lu", (unsigned long)frame_size);
+    size_t frame_size = packet::ns_to_size(frame_length, sample_rate, channel_mask);
+    roc_log(LogDebug, "mixer: initializing: frame_size=%lu", frame_size);
```
gavv: `(unsigned long)` is required in roc_log() due of `%lu`.

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -32,7 +32,8 @@ enum {
     FrameSize = 512,
 
     OutSamples = FrameSize * 100 + 1,
-    InSamples = OutSamples + (FrameSize * 3)
+    InSamples = OutSamples + (FrameSize * 3),
+    SampleRate = 44100
```
gavv: It seems that SampleRate is unused.

src/modules/roc_pipeline/config.h
```diff
@@ -82,12 +85,15 @@ struct SenderConfig {
...
-    //! Number of samples for internal frames.
-    size_t internal_frame_size;
+    //! Duration of the internal frames, in nanoseconds.
+    core::nanoseconds_t internal_frame_length;
 
     //! Packet length, in nanoseconds.
     core::nanoseconds_t packet_length;
 
+    //! Size of the internal frames.
+    size_t internal_frame_size;
```
gavv: This looks a bit messy. Now we have `internal_frame_length` AND `internal_frame_size` and should keep them consistent.

Some components use `internal_frame_length` (Mixer), and some use `internal_frame_size` (ResamplerReader, ResamplerWriter).

This approach is error-prone. It will break if we will update one field but forget to update another. It will also break if we will have components in the pipeline with different channel masks.

I'd prefer to have only one field - `internal_frame_length` - and pass it to all components (including BackendDispatcher). Then every component will use `ns_to_size()`.

Thus I suggest to remove `internal_frame_size` (here and below in this file) and `DefaultInternalFrameSize`.

hrishikeshSuresh:
> A bit busy currently, will take a look in a few days. Could you please fix tests? (see travis build).

I looked into it and the error occurs in `src/modules/roc_audio/mixer.cpp`, where we are checking whether frame_size is greater than temp_buf size (line 45). When I printed the values while running the tests, the frame size = 618 and temp_buf size = 500. So do we have to change the values of frame length or buffer size in the tests?

src/modules/roc_sndio/pump.cpp
hrishikeshSuresh: done.

src/modules/roc_audio/mixer.cpp
hrishikeshSuresh: done.

src/tests/roc_audio/test_resampler.cpp
hrishikeshSuresh: removed SampleRate.

src/modules/roc_pipeline/config.h
hrishikeshSuresh: done in most places. I need to test the code and go through it again because in a recent merge, it seems a few function names were changed. so I'm working on this part again.

src/tests/roc_audio/test_resampler.cpp
gavv: Now we have unused DefaultFrameLength :)

src/modules/roc_audio/resampler_builtin.cpp
```diff
@@ -125,9 +125,12 @@ bool BuiltinResampler::set_scaling(float new_scaling) {
     // Window's size changes according to scaling. If new window size
     // doesn't fit to the frames size -- deny changes.
     if (window_size_ * new_scaling >= frame_size_ch_) {
+        roc_log(LogError,
+                "bool expression %lu >= %lu",
+                (unsigned long)(window_size_ * new_scaling), (unsigned long)frame_size_ch_);
         roc_log(LogError,
                 "resampler: scaling does not fit frame size:"
-                " window_size=%lu frame_size=%lu scaling=%.5f",
+                "window_size=%lu frame_size=%lu scaling=%.5f",
```
gavv: We need that space :)

src/modules/roc_pipeline/receiver_session.cpp
```diff
@@ -145,6 +145,16 @@ ReceiverSession::ReceiverSession(const ReceiverSessionConfig& session_config,
             }
             areader = resampler_poisoner_.get();
         }
+<<<<<<< HEAD
```
gavv: Unresolved conflict.

src/tools/roc_conv/main.cpp
```diff
@@ -68,23 +69,33 @@ int main(int argc, char** argv) {
 
     pipeline::ConverterConfig config;
 
-    if (args.frame_size_given) {
-        if (args.frame_size_arg <= 0) {
-            roc_log(LogError, "invalid --frame-size: should be > 0");
+    if (args.frame_length_given) {
+        if (!core::parse_duration(args.frame_length_arg, config.internal_frame_length)) {
+            roc_log(LogError, "invalid --frame-length: frame size unspecified");
```
gavv: `frame size unspecified` -> `invalid format`

src/tools/roc_conv/main.cpp
```diff
@@ -68,23 +69,33 @@ int main(int argc, char** argv) {
...
+    core::BufferPool<audio::sample_t> pool(
+        allocator, packet::ns_to_size(config.internal_frame_length,
+                                      config.input_sample_rate, config.input_channels),
+        args.poisoning_flag);
 
     sndio::Config source_config;
     source_config.channels = config.input_channels;
     source_config.sample_rate = 0;
-    source_config.frame_size = config.internal_frame_size;
+    source_config.frame_size = packet::ns_to_size(config.internal_frame_length,
```
gavv: We should also replace sndio::Config::frame_size (in bytes) with sndio::Config::frame_length (in nanoseconds) and perform the conversion inside sndio.

src/tools/roc_conv/main.cpp
gavv: The same in other tools.

src/tests/roc_audio/test_mixer.cpp
```diff
@@ -20,11 +20,11 @@ namespace audio {
...
 namespace {
 
-enum { BufSz = 100, MaxSz = 500 };
+enum { BufSz = 100, FrameLength = (core::nanoseconds_t) 7000000, SampleRate = 44100, ChannelMask = 0x1 };
 
 core::HeapAllocator allocator;
-core::BufferPool<sample_t> buffer_pool(allocator, MaxSz, true);
-core::BufferPool<sample_t> large_buffer_pool(allocator, MaxSz * 10, true);
+core::BufferPool<sample_t> buffer_pool(allocator, FrameLength, true);
+core::BufferPool<sample_t> large_buffer_pool(allocator, FrameLength * 10, true);
```
gavv: Now we're passing FrameLength both to Mixer and BufferPool. This is also wrong: BufferPool accepts size in bytes, and Mixer accepts duration in nanoseconds.

I guess we should pass FrameLength to Mixer and pass FrameLength converted to bytes to BufferPool.

gavv: We should also fix other tests. FrameLength and sizes passed to pools should correspond to each other.

gavv:
Could you please rebase on fresh develop and make the history linear?

hrishikeshSuresh:
I have made the history linear right now. Will start making the changes that were requested in the previous commit.

src/tests/roc_audio/test_mixer.cpp
hrishikeshSuresh: I replaced 'MaxSz' with 'FrameLength'. Should I replace 'BufSz' with it's corresponding frame length value and convert it to bytes before using it?

gavv: Look, you can't just replace MaxSz with FrameLength.

Before this PR, we used number of samples everywhere. Now we're using number of samples in some places and number of nanoseconds in other places.

In particular:

* Pool constructor accepts number of samples, as before
* Mixer constructor now accepts number of nanoseconds
* reader.add() accepts number of samples, as before
* expect_output() accepts number of samples, as before

In these tests I'd suggest to use three constants:

* `BufSz` - "regular" buffer size, number of samples
* `MaxBufSz` - maximum buffer size, number of samples
* `MaxBufDuration` - duration corresponding to the maximum buffer size, number of nanoseconds

And use them as follows:

* pass MaxBufDuration to Mixer constructor
* pass MaxBufSz to Pool constructor (this part remains unchanged)
* pass either BufSz or MaxBufSz to reader.add() and expect_output(), depending on the test (this part remains unchanged)

We can define MaxBufDuration to be calculated from MaxBufSz, or vice versa. I'd suggest to leave MaxBufSz equal to 500, as before, and to calculate MaxBufDuration from it.

gavv: And I think we should use the same approach in other tests as well. I.e. define maximum size and calculate corresponding duration from it.

gavv: And it'd better to use some similar names, e.g. `MaxBufSize` + `MaxBufDuration`, as suggested above, to make it clear that these two constants are related.

src/modules/roc_audio/resampler_reader.cpp
```diff
@@ -18,10 +18,13 @@ namespace audio {
...
                                  core::BufferPool<sample_t>& buffer_pool,
-                                 size_t frame_size)
+                                 core::nanoseconds_t frame_length,
+                                 size_t sample_rate,
+                                 roc::packet::channel_mask_t ch_mask)
     : resampler_(resampler)
     , reader_(reader)
-    , frame_size_(frame_size)
+    , frame_size_(packet::ns_to_size(frame_length, sample_rate,
+                  ch_mask))
```
gavv: We need to check if frame_size_ is zero, as we do it in ResamplerWriter.

src/modules/roc_sndio/backend_dispatcher.cpp
```diff
@@ -70,7 +70,10 @@ BackendDispatcher::BackendDispatcher()
 #endif // ROC_TARGET_SOX
 }
 
-void BackendDispatcher::set_frame_size(size_t frame_size) {
+void BackendDispatcher::set_frame_size(core::nanoseconds_t frame_length,
+                                       size_t sample_rate,
+                                       packet::channel_mask_t channels) {
+    size_t frame_size = packet::ns_to_size(frame_length, sample_rate, channels);
```
gavv: Let's move this calculation even deeper, inside SoxBackend::set_frame_size(). BackendDispatcher is a high-level component and it'd be nice to hide these details from it.

gavv:
Ping me when this will be ready for review. Don't forget to remove debugging prints, run `scons fmt`, pull fresh develop and rebase on it, and squash all commits into one or several :)

hrishikeshSuresh:
It looks like the build is failing due to `src/tests/roc_audio/test_mock_reader.h` (which I did not edit), showing `error: iteration 65536 invokes undefined behavior [-Werror=aggressive-loop-optimizations]`. Any idea as to why this is happening here, as this file should be unaffected by the changes I have made in the other files?

gavv:
That code did not change, but the code using it did, and gcc detected that the new usage is incorrect. Actually it found a bug in tests that I already mentioned: you incorrectly pass frame length in nanoseconds, instead of frame size in bytes, to mock reader, which expects number of samples. The value in nanoseconds is very high compared to number of samples, and exceeds MaxSz in mock reader, causing a buffer overflow. Gcc detected that.

src/tools/roc_conv/main.cpp
hrishikeshSuresh: done.

src/tests/roc_audio/test_mixer.cpp
hrishikeshSuresh: done.

src/modules/roc_audio/resampler_reader.cpp
hrishikeshSuresh: done.

src/modules/roc_sndio/backend_dispatcher.cpp
hrishikeshSuresh: done.

gavv:
It seems that the file mode of lots of files was changed, unintentionally I guess:

image

Could you revert the mode change?

hrishikeshSuresh:
> It seems that the file mode of lots of files was changed, unintentionally I guess:
> 
> image
> 
> Could you revert the mode change?

Oh, I will fix it right now.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -187,7 +187,10 @@ SoxBackend::SoxBackend()
     sox_get_globals()->output_message_handler = log_handler;
 }
 
-void SoxBackend::set_frame_size(size_t size) {
+void SoxBackend::set_frame_size(core::nanoseconds_t frame_length,
+                                size_t sample_rate,
+                                packet::channel_mask_t channels) {
+    size_t size = packet::ns_to_size(frame_length, sample_rate, channels);
     core::Mutex::Lock lock(mutex_);
```
gavv: Could you please reorder these 2 lines so that the lock will be the first? Unless required for performance reasons, we usually try to keep such locks in the first line of every method for readability.

Why: When you read code and see that some lines are placed before the lock, you have to check every time whether they access shared resources or not to ensure that's not a bug. And if the lock is always the first, you just don't need to think about it and so it's easier to read the code.

src/modules/roc_audio/latency_monitor.cpp
```diff
@@ -213,8 +213,9 @@ bool LatencyMonitor::update_resampler_(packet::timestamp_t pos,
 
 void LatencyMonitor::report_latency_(packet::timestamp_diff_t latency) {
     if (rate_limiter_.allow()) {
-        roc_log(LogDebug, "latency monitor: latency=%ld target=%lu",
-                (long)latency, (unsigned long)target_latency_);;
+        roc_log(LogDebug, "latency monitor: latency=%ld target=%lu", (long)latency,
+                (unsigned long)target_latency_);
+        ;
```
gavv: Nitpick: extra semicolon.

src/tests/roc_audio/test_mixer.cpp
```diff
@@ -20,11 +20,12 @@ namespace audio {
 
 namespace {
 
-enum { BufSz = 100, MaxSz = 500 };
+enum { BufSz = 100, MaxBufDuration = (core::nanoseconds_t) 11345000, SampleRate = 44100,
+       ChannelMask = 0x1, MaxBufSz = 500 };
```
gavv: OK, so `11345000` was computed as `500 * 1000000000 / 44100`, i.e. `MaxBufSz * core::Second / SampleRate`, right? Can we code it explicitly? Let's compiler to do the job for us. I mean something like:

```
const core::nanoseconds_t MaxBufDuration = MaxBufSz * core::Second / SampleRate;
```

It would be both more readable and easier to maintain (since there will be no need to recompute it if MaxBufSz or SampleRate is changed).

The same for other tests...

src/tests/roc_pipeline/test_converter.cpp
```diff
@@ -46,8 +46,6 @@ TEST_GROUP(converter) {
         config.input_sample_rate = SampleRate;
         config.output_sample_rate = SampleRate;
 
-        config.internal_frame_size = MaxBufSize;
```
gavv: Don't we need to compute config.internal_frame_size from MaxBufSize, as in other tests? I'd avoid relying on the default value here.

src/tests/roc_sndio/target_sox/test_sox_sink.cpp
```diff
@@ -52,7 +52,7 @@ TEST(sox_sink, has_clock) {
 }
 
 TEST(sox_sink, sample_rate_auto) {
-    sink_config.sample_rate = 0;
+    sink_config.sample_rate = SampleRate;
```
gavv: Unintentional change? This test check the specific case when rate is zero.

src/tests/roc_sndio/target_sox/test_sox_source.cpp
```diff
@@ -86,12 +92,13 @@ TEST(sox_source, sample_rate_auto) {
...
 
-        Pump pump(buffer_pool, mock_source, sox_sink, FrameSize, Pump::ModeOneshot);
+        Pump pump(buffer_pool, mock_source, sox_sink, FrameLength, SampleRate, ChMask,
+                  Pump::ModeOneshot);
         CHECK(pump.valid());
         CHECK(pump.run());
     }
 
-    source_config.sample_rate = 0;
+    source_config.sample_rate = SampleRate;
```
gavv: Unintentional change?

src/tools/roc_conv/main.cpp
```diff
@@ -68,23 +69,32 @@ int main(int argc, char** argv) {
...
     pipeline::ConverterConfig config;
 
-    if (args.frame_size_given) {
-        if (args.frame_size_arg <= 0) {
-            roc_log(LogError, "invalid --frame-size: should be > 0");
+    if (args.frame_length_given) {
+        if (!core::parse_duration(args.frame_length_arg, config.internal_frame_length)) {
+            roc_log(LogError, "invalid --frame-length: frame length unspecified");
+            return 1;
+        }
```
gavv: `frame length unspecified` is not correct. if parse_duration() returned error, it means that frame length *was* specified, but in an invalid format. So it should be something like: `invalid --frame-length: bad format`. Same for other tools.

gavv:
Serious work! A few more minor comments... I think we're at the finish line though.

PS. BTW the file mode change was not fully reverted, see https://github.com/roc-project/roc/pull/274/files

src/tests/roc_sndio/target_sox/test_sox_source.cpp
hrishikeshSuresh: I changed this because was checking for build failures in other places and this one was always failing, and later forgot to change it back to 0. I think the issue is, when sample rate is 0, frame length will become infinite. So I think I should check for this case in SoxSink and SoxSource constructors.

hrishikeshSuresh:
```
src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp:63: error: roc_
panic()

ERROR: roc_sndio: !valid_
```
This is the error I get.

src/tests/roc_pipeline/test_converter.cpp
hrishikeshSuresh: We removed 'internal_frame_size' member from `ConverterConfig` and all other structures. Currently, we have only internal frame length, sample rate and channel mask. In line 49, I have already made an update for internal_frame_length.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
hrishikeshSuresh: done.

src/modules/roc_audio/latency_monitor.cpp
hrishikeshSuresh: done.

src/tests/roc_audio/test_mixer.cpp
hrishikeshSuresh: done.

src/tools/roc_conv/main.cpp
hrishikeshSuresh: done.

src/tests/roc_sndio/target_sox/test_sox_source.cpp
gavv: If sample rate is zero, the preferred device rate is gathered from sox or pulseaudio. This is a special mode. We should keep it working. So I guess we should calculate frame size from frame length only after acquiring actual rate.

src/tests/roc_pipeline/test_converter.cpp
gavv: Ah, I see now.

gavv:
Please also revert file mode change for `src/tests/roc_rtp/test_packets/generate_headers.py`.

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -35,7 +35,8 @@ enum {
     FrameSize = 512,
 
     OutSamples = FrameSize * 100 + 1,
-    InSamples = OutSamples + (FrameSize * 3)
+    InSamples = OutSamples + (FrameSize * 3),
+	FrameLength = (core::nanoseconds_t) ( FrameSize * core::Second / InSamples )
```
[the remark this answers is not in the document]
hrishikeshSuresh: done.

src/tests/roc_pipeline/test_sender.cpp
```diff
@@ -42,6 +42,8 @@ enum {
     ManyFrames = FramesPerPacket * 20
 };
 
+const core::nanoseconds_t FrameLength = MaxBufSize * core::Second / (SampleRate * packet::num_channels(ChMask));
```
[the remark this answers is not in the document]
hrishikeshSuresh: done. Wherever `FrameLength` doesn't correspond to MaxBufSize, I've replaced it with `FrameDuration`.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -28,7 +29,7 @@ SoxSink::SoxSink(core::IAllocator& allocator, const Config& config)
         return;
     }
 
-    if (config.frame_size == 0) {
+    if (buffer_size_ == 0 && config.sample_rate != 0) {
```
hrishikeshSuresh: This alone won't work because I have to re-calculate the frame size with device rate from sox or pulseaudio. Will add that with another commit.

src/tests/roc_pipeline/test_sender.cpp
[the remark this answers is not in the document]
gavv: We should also add packet::num_channels to test_mixer.cpp and test_converter.cpp.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
gavv: Yeah, I guess we should delay `buffer_size_` calculation to `prepare()` call.

gavv:
@hrishikeshSuresh this PR is steadily accumulating conflicts, which will make rebasing harder and harder... Do you have some time to finish it? If you're out of time, I could try rebase it and fix the remaining issue with sample rate. IIRC it's the only unresolved issue in this PR.

hrishikeshSuresh:
I am working on it. I think I'll push the changes within 2 days.

gavv:
OK, thanks!

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
hrishikeshSuresh: So prepare() should be called in the constructor, yes?

gavv: We pass frame_length in nanoseconds to constructor. We need to calculate frame_size in bytes from frame_length and sample_rate. However, we don't know sample_rate until open() is called.

open() calls prepare_() and open_(). open_() is the point from which we know sample_rate.

We can reorder prepare_() and open_() calls and calculate frame_size from frame_length in prepare_().

Something like that.

hrishikeshSuresh:
I'll start rebasing it now.

hrishikeshSuresh:
I think you can review the PR now! Only changes I had to make was in `roc_sndio/target_sox/sox_source.cpp`  and `sox_sink.cpp` and updated the tests in roc_sndio, to get device rate from sox/pulseaudio.

gavv:
Great. But could you rebase on fresh develop? The PR have 30 commits currently, and I guess it should contain only one commit (the last one). I think the easiest way to do it is to pull fresh develop, switch to it, and cherry-pick your commit there.

Also, there are some unintentional changes in this PR. E.g.: `roc_recv.rst` (unintentional new line removal?), `src/lib/src/private.h` (it was deleted, PR restores it), `src/lib/src/sender.cpp` (it was moved, PR restores it), `src/modules/roc_address/parse_socket_addr.cpp~7f5219a563768d18bf2a4834935915fd4b68ecfa` (what is this), and so on. After rebasing, please go through your diff and cleanup it.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -28,13 +29,17 @@ SoxSink::SoxSink(core::IAllocator& allocator, const Config& config)
...
     }
 
-    if (config.latency != 0) {
-        roc_log(LogError, "sox sink: setting io latency not supported by sox backend");
+    channels_ = config.channels;
+    frame_length_ = config.frame_length;
+
+    if (frame_length_ == 0) {
+        roc_log(LogError, "sox sink: frame length is zero");
+        out_signal_.rate = sample_rate();
```
gavv: Why do we have this line?

```
out_signal_.rate = sample_rate();
```

I think if it will be executed, it will panic (since `valid_` is false here).

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -134,6 +139,11 @@ void SoxSink::write(audio::Frame& frame) {
 }
 
 bool SoxSink::prepare_() {
+    if (buffer_size_ == 0) {
```
gavv: I think this check is incorrect:

```
if (buffer_size_ == 0) {
```

With this check, we update `buffer_size_` only if it was zero, i.e. only if `config.sample_rate` was zero in constructor. However, even if `sample_rate` was NOT zero in constructor, it may change after opening device (depending on what is supported by device).

So I think we should recompute `buffer_size_` unconditionally.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -134,6 +139,11 @@ void SoxSink::write(audio::Frame& frame) {
...
+        out_signal_.rate = sample_rate();
```
gavv: I guess this line is not needed:

```
out_signal_.rate = sample_rate();
```

We should only update buffer size. `out_signal_.rate` should be already correct here.

src/modules/roc_sndio/target_sox/roc_sndio/sox_source.cpp
```diff
@@ -324,6 +328,11 @@ bool SoxSource::open_() {
         return false;
     }
 
+    if (buffer_size_ == 0) {
```
gavv: Like in SoxSink, the check is incorrect.

Also, currently SoxSource calls prepare_() and then open_(), so we will use incorrect sample rate here. We should first open device and determine its rate, and only then calculate buffer size and resize buffer.

But SoxSource::prepare_() also contains some code that should be executed *before* open (storing driver and input names). I suggest to split prepare into two functions: setup_names_() and setup_buffer_(). And call them in the following sequence: setup_names_(), open_(), setup_buffer_().

For consistency, let's also rename SoxSink::prepare_() to setup_buffer_().

src/modules/roc_sndio/target_sox/roc_sndio/sox_source.cpp
```diff
@@ -324,6 +328,11 @@ bool SoxSource::open_() {
...
+        in_signal_.rate = sample_rate();
```
gavv: I guess this line is not needed:

```
in_signal_.rate = sample_rate();
```

src/modules/roc_pipeline/receiver_session.cpp
```diff
@@ -158,7 +159,9 @@ ReceiverSession::ReceiverSession(const ReceiverSessionConfig& session_config,
 
         resampler_reader.reset(new (allocator) audio::ResamplerReader(
                                    *areader, *resampler_, sample_buffer_pool,
-                                   common_config.internal_frame_size),
+                                   common_config.internal_frame_length,
+                                   common_config.output_sample_rate,
+                                   common_config.output_channels),
```
gavv: ResamplerReader uses frame_length to split its *input* stream into frames. So we should pass to it the sample rate of the input stream, not of the output stream. So we should replace `common_config.output_sample_rate` with `format->sample_rate` here and above when calling `new_resampler()`.

src/modules/roc_pipeline/converter_sink.cpp
```diff
@@ -38,16 +38,18 @@ ConverterSink::ConverterSink(const ConverterConfig& config,
...
             return;
         }
 
-        resampler_writer_.reset(new (allocator)
-                                    audio::ResamplerWriter(*awriter, *resampler_, pool,
-                                                           config.internal_frame_size),
+        resampler_writer_.reset(new (allocator) audio::ResamplerWriter(
+                                    *awriter, *resampler_, pool,
+                                    config.internal_frame_length,
+                                    config.output_sample_rate, config.output_channels),
```
gavv: ResamplerWriter uses frame_length to split its input stream (samples written into it) into frames, so we should use `input_sample_rate` instead of `output_sample_rate` here and in `new_resampler()`.

src/tests/roc_sndio/target_sox/test_pump.cpp
```diff
@@ -24,7 +24,10 @@ namespace sndio {
 
 namespace {
 
-enum { MaxBufSize = 8192, FrameSize = 512, SampleRate = 44100, ChMask = 0x3 };
+enum { MaxBufSize = 8192, SampleRate = 44100, ChMask = 0x3 };
+
+const core::nanoseconds_t MaxBufDuration =
+    MaxBufSize * core::Second / (SampleRate * packet::num_channels(ChMask));
```
gavv: I suggest to rename MaxBufSize/MaxBufDuration to BufSize/BufDuration and reduce size from 8192 to 512 to keep tests a bit more lightweight

src/tests/roc_sndio/target_sox/test_sox_source.cpp
```diff
@@ -31,6 +31,9 @@ enum {
     NumChans = 2
 };
 
+const core::nanoseconds_t MaxBufDuration =
+    MaxBufSize * core::Second / (SampleRate * packet::num_channels(ChMask));
```
gavv: I guess in this test we need FrameDuration (calculated from FrameSize) instead of MaxBufDuration (calculated from MaxBufSize).

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
hrishikeshSuresh: I think we should have this but it must be moved after `valid_` is true.

hrishikeshSuresh: About why we need this, on line 47, we are assigning out_signal_.rate to config.sample_rate. So I think we should update it accordingly.

gavv: Why do you think so?

In ctor we set `out_signal_.rate` to the rate from config:

```
out_signal_.rate = config.sample_rate;
```

Then `open_()` passes `out_signal_` to `sox_open_write()`, which tries to use `out_signal_.rate` if it is 1) non-zero 2) supported by device / file.

Then `sox_open_write()` selects some rate (maybe the requested one, and maybe not) and stores the selected rate into `output_->signal.rate`.

`sample_rate()` checks that the device is already opened and returns `output_->signal.rate`.

So, in other words:

* `out_signal_.rate` is the rate which we're requesting from sox
* `output_->signal.rate` and `sample_rate()` is the rate actually selected by sox

So why we may need this assignment?

The naming is not obvious though. Maybe we want to rename some of these fields.

gavv: > About why we need this, on line 47, we are assigning out_signal_.rate to config.sample_rate. So I think we should update it accordingly.

Didn't see this message. I see. Well, we don't use `out_signal_` in any way except passing it to `sox_open_write()`, so it seems there is no need to update it. Maybe rename it to `requested_signal_params_` or something, to make it clear that it contains *requested* parameters, not actual ones.

src/modules/roc_pipeline/converter_sink.cpp
hrishikeshSuresh: and `output_channels` should be changed to `input_channels` for ResamplerWriter?

src/modules/roc_pipeline/converter_source.cpp
```diff
@@ -35,16 +35,18 @@ ConverterSource::ConverterSource(const ConverterConfig& config,
...
             return;
         }
 
-        resampler_reader_.reset(new (allocator)
-                                    audio::ResamplerReader(*areader, *resampler_, pool,
-                                                           config.internal_frame_size),
+        resampler_reader_.reset(new (allocator) audio::ResamplerReader(
+                                    *areader, *resampler_, pool,
+                                    config.internal_frame_length,
+                                    config.input_sample_rate, config.input_channels),
```
hrishikeshSuresh: I think ResamplerReader should use output_sample_rate & output_channels instead of input_sample_rate & input_channels as we splitting the output stream?

hrishikeshSuresh: > ResamplerReader uses frame_length to split its input stream into frames. So we should pass to it the sample rate of the input stream, not of the output stream. 
Saw this comment only now. So I think I will do a similar thing here.

src/modules/roc_pipeline/converter_sink.cpp
gavv: Yes!

src/modules/roc_pipeline/converter_source.cpp
gavv: > So I think ResamplerReader should use input_sample_rate.

Yes.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -133,7 +137,10 @@ void SoxSink::write(audio::Frame& frame) {
     write_(buffer_data, buffer_pos);
 }
 
-bool SoxSink::prepare_() {
+bool SoxSink::setup_buffer_() {
+    size_t requested_device_rate = sample_rate();
```
gavv: *requested rate* (i.e. the rate we're asking to be used for device) is stored in `out_signal_.rate`. And `sample_rate()` returns `output_->signal.rate`, which is *actual rate* selected for device. And we're using actual rate to calculate buffer size.

So we're correctly using `sample_rate()` here, but the variable is named incorrectly.

I suggest to rename the variable to `rate` or `actual_rate` or something like that, or just remove it at all:

```
buffer_size_ = packet::ns_to_size(frame_length_, sample_rate(), channels_);
```

hrishikeshSuresh: done.

src/tests/roc_sndio/target_sox/test_sox_source.cpp
hrishikeshSuresh: done.

src/tests/roc_sndio/target_sox/test_pump.cpp
hrishikeshSuresh: done.

src/modules/roc_pipeline/converter_sink.cpp
hrishikeshSuresh: done.

src/modules/roc_pipeline/receiver_session.cpp
hrishikeshSuresh: done.

src/modules/roc_sndio/target_sox/roc_sndio/sox_source.cpp
hrishikeshSuresh: done.

src/modules/roc_sndio/target_sox/roc_sndio/sox_source.cpp
hrishikeshSuresh: done.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
hrishikeshSuresh: done.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
hrishikeshSuresh: done.

src/modules/roc_pipeline/receiver_session.cpp
```diff
@@ -158,7 +159,8 @@ ReceiverSession::ReceiverSession(const ReceiverSessionConfig& session_config,
 
         resampler_reader.reset(new (allocator) audio::ResamplerReader(
                                    *areader, *resampler_, sample_buffer_pool,
-                                   common_config.internal_frame_size),
+                                   common_config.internal_frame_length,
+                                   format->sample_rate, common_config.output_channels),
```
gavv: One more problem: it should be `session_config.channels` (# of channels in input stream from network) instead of `common_config.output_channels` (# of channels in output stream to user).

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -133,7 +137,8 @@ void SoxSink::write(audio::Frame& frame) {
     write_(buffer_data, buffer_pos);
 }
 
-bool SoxSink::prepare_() {
+bool SoxSink::setup_buffer_() {
+    buffer_size_ = packet::ns_to_size(frame_length_, sample_rate(), channels_);
```
gavv: Here and in SoxSource, we should also check if the calculated `buffer_size_` is zero and return false in this case.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -17,7 +17,8 @@ namespace sndio {
 SoxSink::SoxSink(core::IAllocator& allocator, const Config& config)
     : output_(NULL)
     , buffer_(allocator)
-    , buffer_size_(config.frame_size)
+    , buffer_size_(
+          packet::ns_to_size(config.frame_length, config.sample_rate, config.channels))
```
gavv: Nitpick: here and in SoxSource, I suggest to set `buffer_size_` to zero in constructor to make it clear that it can't be used until `setup_buffer_()` is called.

src/tests/roc_sndio/target_sox/test_sox_source.cpp
```diff
@@ -214,15 +223,17 @@ TEST(sox_source, pause_restart) {
 
 TEST(sox_source, eof_restart) {
     core::TempFile file("test.wav");
-
+    const core::nanoseconds_t BufDuration =
+        FrameSize * core::Second / (SampleRate * packet::num_channels(ChMask));
```
gavv: It seems this constant is not needed and we can just use FrameDuration instead?

src/modules/roc_pipeline/receiver_session.cpp
hrishikeshSuresh: done.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
hrishikeshSuresh: done.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
hrishikeshSuresh: done.

src/tests/roc_sndio/target_sox/test_sox_source.cpp
hrishikeshSuresh: done.

gavv:
Thanks!

Issue #223, opened 2019-06-16
roc_lib test fails under valgrind


gavv:
So far it happened only once, on Travis. Probably a race in test (that test is real-time).

```
python2 scripts/wrappers/timeout.py 300 valgrind --max-stackframe=10475520 --error-exitcode=1 --exit-on-first-error=yes bin/x86_64-pc-linux-gnu/roc-test-lib
==1783== Memcheck, a memory error detector
==1783== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==1783== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==1783== Command: bin/x86_64-pc-linux-gnu/roc-test-lib
==1783== 
src/tests/roc_lib/test_sender_receiver.cpp:215: error: roc_panic()
ERROR: roc_test: failed comparing sample #1100
frame_num: 112
original: -0.966400,	received: 0.000000
Backtrace:
# roc::core::print_backtrace()
# roc::core::crash(char const*)
# bin/x86_64-pc-linux-gnu/roc-test-lib() [0x42dfea]
# bin/x86_64-pc-linux-gnu/roc-test-lib() [0x422619]
# roc::TEST_sender_receiver_losses_Test::testBody()
# bin/x86_64-pc-linux-gnu/roc-test-lib() [0x459534]
# bin/x86_64-pc-linux-gnu/roc-test-lib
# Utest::run()
# UtestShell::runOneTestInCurrentProcess(TestPlugin*, TestResult&)
# bin/x86_64-pc-linux-gnu/roc-test-lib() [0x4576bc]
# bin/x86_64-pc-linux-gnu/roc-test-lib
# UtestShell::runOneTest(TestPlugin*, TestResult&)
# TestRegistry::runAllTests(TestResult&)
# CommandLineTestRunner::runAllTests()
# CommandLineTestRunner::runAllTestsMain()
# CommandLineTestRunner::RunAllTests(int, char const**)
# bin/x86_64-pc-linux-gnu/roc-test-lib
# /lib/x86_64-linux-gnu/libc.so.6
# bin/x86_64-pc-linux-gnu/roc-test-lib
==1783== 
==1783== Process terminating with default action of signal 6 (SIGABRT)
==1783==    at 0x4E32ED7: raise (raise.c:51)
==1783==    by 0x4E14534: abort (abort.c:79)
==1783==    by 0x42E43E: roc::core::crash(char const*) (crash.cpp:102)
[...]

PR #275, opened 2019-10-16
Fixing roc_lib test fails under valgrind

alexandremgo:
Fix issue #223 

I've relaxed the test requirements to have a certain percentage (for now 90%) of identical middle stream samples.

Should it raises a panic if a sample is not identical and is different from zero ?

### DIFF
--- a/src/tests/roc_lib/test_sender_receiver.cpp
+++ b/src/tests/roc_lib/test_sender_receiver.cpp
@@ -186,6 +186,9 @@ public:
         size_t leading_zeros = 0;
         size_t sample_num = 0;
         size_t frame_num = 0;
+        
+        size_t middle_stream_total = 0;
+        size_t middle_stream_identical = 0;
 
         bool seek_first = true;
         bool finish = false;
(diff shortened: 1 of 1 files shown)


gavv:
The code looks good!

> Should it raises a panic if a sample is not identical and is different from zero ?

Yep. It would be a more correct check.

gavv:
Also, I've done some testing and opened a new issue: #277.

src/tests/roc_lib/test_sender_receiver.cpp
```diff
@@ -212,29 +215,40 @@ class Receiver {
...
                         roc_panic("%s", sbuff);
-                    } else {
-                        sample_num++;
                     }
                 }
             }
         }
+
+        double identical = (double)middle_stream_identical / middle_stream_total;          
+        if(identical < .9) {
```
dshil: @alexandremgo could you please run `scons -Q fmt` to use the existing formatting style?

alexandremgo: yes sorry

gavv:
Thanks!

PR #280, opened 2019-10-24
Extract and reuse string helpers

gavv:
For endpoint URIs (#258), I'll need the same string helpers as we already have re-implemented in a few places. I've extracted these functions so that hey can be reused.

Note that since these functions are used in backtrace printing, they should be signal-safe and can not panic.


gavv:
After some thought I decided that they can panic if we can be sure that these panics will never happen in the backtrace code. This is quite likely because there is not much such code, so I added the panics.

gavv:
@hrishikeshSuresh FYI.

PR #281, opened 2019-10-24
IoURI fixes

gavv:
I'm working on EndpointURI (#258) and also prepared several fixes for IoURI to make the two URL types consistent.

Changes:

* Don't parse query and fragment as a part of a path. Instead, fail the whole parsing if they're present. (IoURI will disallow them. EndpointURI will allow them).

* Don't allow formatting invalid URIs (e.g. without a scheme or without a path).

* Allocate memory on heap for large URI components. Don't limit components lengths. (EndpointURI will contain more components and using fixed-size arrays for them would lead to too large objects, something like a few kilobytes per URI).

### DIFF
--- a/src/modules/roc_address/io_uri.h
+++ b/src/modules/roc_address/io_uri.h
@@ -12,18 +12,20 @@
 #ifndef ROC_ADDRESS_IO_URI_H_
 #define ROC_ADDRESS_IO_URI_H_
 
+#include "roc_core/noncopyable.h"
 #include "roc_core/stddefs.h"
 
 namespace roc {
 namespace address {
 
 //! Audio file or device URI.
-struct IoURI {
+class IoURI : public core::NonCopyable<> {
+public:
     //! Initialize empty URI.
     IoURI();
 
-    //! Returns true if the URI is empty.
-    bool is_empty() const;
+    //! Returns true if the URI has all required components.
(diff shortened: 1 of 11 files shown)


src/modules/roc_address/io_uri.h
```diff
@@ -12,37 +12,56 @@
...
 namespace roc {
 namespace address {
 
 //! Audio file or device URI.
-struct IoURI {
+class IoURI : public core::NonCopyable<> {
+public:
     //! Initialize empty URI.
-    IoURI();
+    IoURI(core::IAllocator&);
```
dshil: explicit?

src/modules/roc_address/io_uri.h
```diff
@@ -12,37 +12,56 @@
...
 
-    //! Returns true if the URI is empty.
-    bool is_empty() const;
+    //! Returns true if the URI has all required components.
```
dshil: Could you specify what these components are?

src/modules/roc_address/io_uri.h
gavv: Done.

src/modules/roc_address/io_uri.h
gavv: Done.

PR #284, opened 2019-11-03
Freebsd

strfry:
These are some necessary changes to make it work on FreeBSD.
I added an additional target that is explicitly considered at several points. Just hope it won"t get too messy with more target OS's.

Note that despite the unit tests finishing without errors, this is little tested besides a quick try with roc-send.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -7,6 +7,7 @@ import SCons.Script
 # supported platform names
 supported_platforms = [
     'linux',
+    'freebsd',
     'darwin',
     'android',
 ]
@@ -511,6 +512,8 @@ if not platform:
         platform = 'android'
     elif 'linux' in host:
         platform = 'linux'
+    elif 'freebsd' in host:
+        platform = 'freebsd'
     elif 'darwin' in host:
         platform = 'darwin'
 
@@ -598,7 +601,7 @@ else:
                  "provide either known '--platform' or '--override-targets' option"),
                     host, ', '.join(supported_platforms))
 
-    if platform in ['linux', 'android', 'darwin']:
+    if platform in ['linux', 'android', 'darwin', 'freebsd']:
         env.Append(ROC_TARGETS=[
             'target_posix',
             'target_stdio',
@@ -606,12 +609,12 @@ else:
             'target_libuv',
         ])
 
-    if platform in ['linux', 'android']:
+    if platform in ['linux', 'android', 'freebsd']:
         env.Append(ROC_TARGETS=[
             'target_posixtime',
         ])
 
-    if platform in ['linux']:
+    if platform in ['linux', 'freebsd']:
         if not GetOption('disable_libunwind'):
             env.Append(ROC_TARGETS=[
                 'target_libunwind',
@@ -1027,7 +1030,7 @@ if GetOption('enable_pulseaudio_modules'):
 
 env = conf.Finish()
 
-if 'target_posix' in env['ROC_TARGETS'] and platform not in ['darwin']:
+if 'target_posix' in env['ROC_TARGETS'] and platform not in ['darwin', 'freebsd']:
     env.Append(CPPDEFINES=[('_POSIX_C_SOURCE', '200809')])
 
 for t in env['ROC_TARGETS']:
--- a/SConstruct
+++ b/SConstruct
@@ -60,13 +60,13 @@ env.SConsignFile(os.path.join(env.Dir('#').abspath, '.sconsign.dblite'))
 # libraries are no different
 env['STATIC_AND_SHARED_OBJECTS_ARE_THE_SAME'] = 1
 
-if platform.system() == 'Darwin':
-    default_prefix = '/usr/local'
-else:
+if platform.system() == 'Linux':
     # it would be better to use /usr/local on Linux too, but PulseAudio
     # is usually installed in /usr and does no search /usr/local for
     # dynamic libraries; so by default we also use /usr for consistency
     default_prefix = '/usr'
+else:
+    default_prefix = '/usr/local'
 
 AddOption('--prefix',
           dest='prefix',
--- a/scripts/3rdparty.py
+++ b/scripts/3rdparty.py
@@ -320,7 +320,7 @@ if name == 'libuv':
             '--with-pic',
             '--enable-static',
         ])), logfile)
-    execute('make -j', logfile)
+    execute('make', logfile)
     install_tree('include', os.path.join(builddir, 'include'))
     install_files('.libs/libuv.a', os.path.join(builddir, 'lib'))
 elif name == 'openfec':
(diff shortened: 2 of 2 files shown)


SConstruct
```diff
@@ -7,6 +7,7 @@ import SCons.Script
 # supported platform names
 supported_platforms = [
     'linux',
+    'freebsd',
```
gavv: What do you think about other BSDs? If in future we will also add support, say, for OpenBSD, will we need a different platform for it? Or maybe we can have a single `bsd` platform for all of them, instead of `freebsd`, `openbsd`, etc? If you don't known, we can keep `freebsd` for now.

gavv:
Cool!

> I added an additional target that is explicitly considered at several points. Just hope it won"t get too messy with more target OS's.

I think this is OK.

scripts/3rdparty.py
```diff
@@ -320,7 +320,7 @@ def makeenv(envlist):
             '--with-pic',
             '--enable-static',
         ])), logfile)
-    execute('make -j', logfile)
+    execute('make', logfile)
```
gavv: `make -j` is significantly faster for some of the dependencies, at least on on my machine. I'd prefer to keep it at least on Linux.

Can we do one of the following?

1. Detect whether `make` is GNU make and use `-j` if it is?

2. Or detect whether `gmake` is available on the system and use `gmake -j` if it is and `make` if it's not?

3. Or just always require `gmake` to be available (it would be a simpler approach but I'm not sure if gmake is available everywhere by default, e.g. on Alpine and FreeBSD. If it's not, this approach is not preferred).

gavv:
There are also other places in SConstruct where we could add `freebsd`.

```
        if platform in ['linux'] and not GetOption('disable_pulseaudio'):
            env.Append(ROC_TARGETS=[
                'target_pulseaudio',
            ])
```

```
if ((not GetOption('disable_tools') \
        or not GetOption('disable_examples')) \
    and not GetOption('disable_pulseaudio')) \
  or GetOption('enable_pulseaudio_modules'):
    if platform in ['linux', 'android']:
        all_dependencies.add('alsa')
        all_dependencies.add('pulseaudio')
```

Since, AFAIK, pulseaudio is available for FreeBSD, I think we should support it. But since it's rarely used there, I guess, we should disable pulseaudio by default on FreeBSD (but not on Linux). We could add `--enable-pulseaudio` option to enable it.

```
    if platform in ['linux', 'darwin']:
        env.Append(LIBS=[
            'pthread',
        ])
```

This adds `-pthread` (without `-l`). Don't we need it on FreeBSD as well?

```
    if platform in ['linux', 'android']:
        test_env['RPATH'] = test_env.Literal('\\$$ORIGIN')

        lib_env['SHLIBSUFFIX'] = '%s.%s' % (lib_env['SHLIBSUFFIX'], abi_version)
        lib_env.Append(LINKFLAGS=[
            '-Wl,-soname,libroc%s' % lib_env['SHLIBSUFFIX'],
        ])

        if variant == 'release':
            lib_env.Append(LINKFLAGS=[
                '-Wl,--version-script=%s' % env.File('#src/lib/roc.version').path
            ])
```

This adds soname and hides some symbols from the shared library. I think we can enable this on FreeBSD too?

```
    if platform in ['linux', 'android']:
        if compiler_ver[:2] >= (6, 0):
            for var in ['CXXFLAGS', 'CFLAGS']:
                env.Append(**{var: [
                    '-Wno-redundant-parens',
                ]})

    if platform in ['linux']:
        if compiler_ver[:2] >= (8, 0):
            for var in ['CXXFLAGS', 'CFLAGS']:
                env.Append(**{var: [
                    '-Wno-extra-semi-stmt',
                    '-Wno-atomic-implicit-seq-cst',
                ]})
```

```
    if platform in ['linux', 'android']:
        env.Append(LINKFLAGS=[
            '-Wl,--no-undefined',
        ])
```

```
if platform in ['linux']:
    tool_env.Append(LINKFLAGS=[
        '-Wl,-rpath-link,%s' % env.Dir('#3rdparty/%s/%s/rpath' % (
            host, thirdparty_compiler_spec)).abspath,
    ])
```

As well as this?

gavv:
BTW did you try to build Roc with different options? Debug/release, enable/disable sanitizers, enable -Werror, enable/disable different dependencies? I'm not saying we need to test all these combinations right now, but this would be a long-term goal to add FreeBSD to the list of officially supported platforms.

Also, we'll need (again, we can do it later) to add FreeBSD to CI. We could use qemu in travis or maybe use [Cirrus CI](https://cirrus-ci.org/guide/FreeBSD/) (I don't know much about them, needs research). It would be preferable to use a single CI for all build though. Are you aware of any other options?

gavv:
Finally, could you please re-target the PR to `develop` branch? See here: https://roc-project.github.io/roc/docs/development/version_control.html

(If github don't allow you to do this, I can do it for you. I have the button, but since it can change commit history and create conflicts, I'm not doing this without your confirmation).

SConstruct
strfry: I don't know what kinds of problems are awaiting on those other problems.
Sure, having just a single `bsd` platform sounds nice, but until this is shown to actually work I'd stick with the more specific ones.

scripts/3rdparty.py
strfry: 4. Determine sensible amount of threads to use (# of CPU cores) and pass it to -j. BSD make supports this option, just not without an argument. 

Just requiring `gmake` would also be fine by me, but unfortunately many Linux distros don't ship it (as in a symlink to `make`)

strfry:
You're probably right about those further instances of 'linux' tests also applying to FreeBSD. I'll look into them later.

The pulseaudio option isn't much of a problem IMHO. The goal is to submit an actual Port Makefile upstream, and dealing with optional dependencies should be part of that. 

I didn't do testing with various configurations yet. Maybe that can serve as a motivation, to move the CI at a later point to a more cross-platform solution ;)

Feel free to do change the PR to `develop`. I actually rebased the thing on `develop` after i noticed it exists, so it must have been a missclick that i can't correct now.

scripts/3rdparty.py
gavv: > Determine sensible amount of threads to use (# of CPU cores) and pass it to -j. BSD make supports this option, just not without an argument.

Great, this is the best option I think.

SConstruct
gavv: OK.

scripts/3rdparty.py
gavv: #310

gavv: Merged into develop.

gavv:
I've cherry-picked cab8dbbc360fc6960cd7cfcdad4ecf3a679841c9 into develop because it seems to be a good idea even without freebsd support.

gavv:
Closing this for now. Feel free to reopen if you decide to continue working on it.

gavv:
I've created an issue: #360.

Issue #235, opened 2019-07-05
Switchable resampler backends


gavv:
Currently we use our own resampler implementation. We have two problems with it:

* Performance. It is quite CPU-consuming.
* Quality. We did not perform a full-fledged research yet.

For the second point, we have #153. However this will need some time.

For now, it would be nice to be able to use alternative resampler implementations. There are at least two good implementations fitting our needs:

* libspeex
* libzita-resampler

It should be quite straightforward to employ them. We currently have ResamplerWriter (sender pipeline element), ResamplerReader (receiver pipeline element), and Resampler (the resampling engine).

So we can do the following:

* add IResampler interface
* make ResamplerWriter and ResamplerReader to use the interface instead of the specific implementation
* make Resampler to implement IResampler and rename it to something like BuiltinResampler
* add SpeexResampler and ZitaResampler implementations

Then we should make the backend configurable:

* add ResamplerBackend enum with an ID for every IResampler implementation
* add ResamplerBackend field to pipeline::SenderConfig and pipeline::ReceiverConfig
* add ResamplerMap, a factory similar to fec::CodecMap, that will create an IResampler implementation according to given ID
* use ResamplerMap in pipeline::Sender and pipeline::ReceiverSession
* add command-line options to all tools to configure ResamplerBackend ID
* also add an enum and a config field to API

gavv:
Assigned to @CristobalM.

gavv:
PR was merged and we are now able to implement new backends. I've created two new issues for that:

* #314
* #315

So I'm closing this one.

@CristobalM thanks! If you'd like to pick up one of those issues please leave a comment there.

PR #287, opened 2019-11-14
Switchable resampler backends (partial)

CristobalM:
Hi. I found this project very interesting and decided to contribute some code.

This PR references https://github.com/roc-project/roc/issues/235.  With https://github.com/roc-project/roc/commit/4350117aef0999431446a989c8d11b6f6fc92ff9 we can add zita and speex resamplers more easily and with https://github.com/roc-project/roc/commit/4350117aef0999431446a989c8d11b6f6fc92ff9 we have speex available, but not implemented yet to do the resampling (gives warnings about unused variables).

### DIFF
--- a/src/modules/roc_audio/resampler_builtin.h
+++ b/src/modules/roc_audio/resampler_builtin.h
@@ -6,14 +6,16 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-//! @file roc_audio/resampler.h
+//! @file roc_audio/resampler_builtin.h
 //! @brief Resampler.
 
-#ifndef ROC_AUDIO_RESAMPLER_H_
-#define ROC_AUDIO_RESAMPLER_H_
+#ifndef ROC_AUDIO_RESAMPLER_BUILTIN_H_
+#define ROC_AUDIO_RESAMPLER_BUILTIN_H_
 
 #include "roc_audio/frame.h"
 #include "roc_audio/ireader.h"
+#include "roc_audio/iresampler.h"
+#include "roc_audio/resampler_config.h"
 #include "roc_audio/units.h"
 #include "roc_core/array.h"
 #include "roc_core/noncopyable.h"
@@ -24,34 +26,14 @@
 namespace roc {
 namespace audio {
 
-//! Resampler parameters.
-struct ResamplerConfig {
-    //! Sinc table precision.
-    //! @remarks
-    //!  Affects sync table size.
-    //!  Lower values give lower quality but rarer cache misses.
-    size_t window_interp;
-
-    //! Resampler internal window length.
-    //! @remarks
-    //!  Affects sync table size and number of CPU cycles.
-    //!  Lower values give lower quality but higher speed and also rarer cache misses.
-    size_t window_size;
-
-    ResamplerConfig()
-        : window_interp(128)
-        , window_size(32) {
-    }
-};
-
 //! Resamples audio stream with non-integer dynamically changing factor.
-class Resampler : public core::NonCopyable<> {
+class BuiltinResampler : public IResampler, public core::NonCopyable<> {
 public:
     //! Initialize.
-    Resampler(core::IAllocator& allocator,
-              const ResamplerConfig& config,
-              packet::channel_mask_t channels,
-              size_t frame_size);
+    BuiltinResampler(core::IAllocator& allocator,
+                     const ResamplerConfig& config,
+                     packet::channel_mask_t channels,
+                     size_t frame_size);
 
     //! Check if object is successfully constructed.
     bool valid() const;
@@ -72,6 +54,8 @@ public:
     void renew_buffers(core::Slice<sample_t>& prev,
                        core::Slice<sample_t>& cur,
                        core::Slice<sample_t>& next);
+
+    ~BuiltinResampler();
 
 private:
     typedef uint32_t fixedpoint_t;
@@ -141,4 +125,4 @@ private:
 } // namespace audio
 } // namespace roc
 
-#endif // ROC_AUDIO_RESAMPLER_H_
+#endif // ROC_AUDIO_RESAMPLER_BUILTIN_H_
--- a/src/modules/roc_audio/resampler_map.cpp
+++ b/src/modules/roc_audio/resampler_map.cpp
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_audio/resampler_map.h"
+#include "roc_audio/resampler_builtin.h"
+#include "roc_audio/resampler_config.h"
+
+namespace roc {
+namespace audio {
+
+IResampler* ResamplerMap::new_resampler(ResamplerBackend resampler_backend,
+                                        core::IAllocator& allocator,
+                                        const ResamplerConfig& config,
+                                        packet::channel_mask_t channels,
+                                        size_t frame_size) {
+    if (resampler_backend != ResamplerBackend_Builtin) {
+        roc_panic(
+            "No valid resampler backend selected.. selected resampler backend is : %d",
+            resampler_backend);
+    }
+
+    core::UniquePtr<IResampler> resampler;
+    switch (resampler_backend) {
+    case ResamplerBackend_Builtin:
+        resampler.reset(new (allocator)
+                            BuiltinResampler(allocator, config, channels, frame_size),
+                        allocator);
+        break;
+    }
+
+    if (!resampler || !resampler->valid()) {
+        return NULL;
+    }
+
+    return resampler.release();
+}
+
+} // namespace audio
+} // namespace roc
--- a/src/modules/roc_pipeline/config.h
+++ b/src/modules/roc_pipeline/config.h
@@ -13,7 +13,7 @@
 #define ROC_PIPELINE_CONFIG_H_
 
 #include "roc_audio/latency_monitor.h"
-#include "roc_audio/resampler.h"
+#include "roc_audio/resampler_config.h"
 #include "roc_audio/watchdog.h"
 #include "roc_core/stddefs.h"
 #include "roc_core/time.h"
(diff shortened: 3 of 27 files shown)


gavv:
Thanks for your interest and for the PR!

Will take a look. I'm a little busy this month, so will take some time.

The build is failing because speex is not installed on our travis docker images. You can read about them here: https://roc-project.github.io/roc/docs/development/continuous_integration.html

Would be cool if you'll add it to the images. Docker images are automatically built from this repo: https://github.com/roc-project/dockerfiles, so you can just make a PR to that repo.

We will also need to add speex to https://github.com/roc-project/vendor. It is used as a cache for --build-3rdparty. It is a submodule of the roc repo.

Also, please install clang-format and run `scons fmt` to format the code according to our config.

src/modules/roc_audio/resampler_map.cpp
```diff
@@ -0,0 +1,42 @@
...
+IResampler* ResamplerMap::new_resampler(ResamplerBackend resampler_backend,
+                                    core::IAllocator& allocator,
+                                    const ResamplerConfig& config,
+                                    packet::channel_mask_t channels,
+                                    size_t frame_size){
+    
+    switch (resampler_backend)
+    {
+    case BUILTIN_RESAMPLER:
+        current_resampler.reset(new (allocator) BuiltinResampler(allocator, config, channels, frame_size), allocator);
```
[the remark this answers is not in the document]
gavv: I also suggest to remove `valid()` from IResampler interface and check it here, before returning the pointer, like:

```
UniquePtr<BuiltinResampler> r(new (allocator) BuiltinResampler());
if (!r || !r->valid()) {
    return NULL;
}
return r.release();
```

This approach will give a guarantee that ResamplerMap may return only successfully constructed objects, which would be nice. We already use such approach in other factories.

src/modules/roc_pipeline/converter.h
```diff
@@ -50,7 +51,9 @@ class Converter : public sndio::ISink, public core::NonCopyable<> {
     audio::NullWriter null_writer_;
 
     core::UniquePtr<audio::PoisonWriter> resampler_poisoner_;
-    core::UniquePtr<audio::ResamplerWriter> resampler_;
+    core::UniquePtr<audio::ResamplerWriter> resampler_writer_;
+    core::UniquePtr<audio::ResamplerMap> resampler_map_;
```
gavv: Other maps (CodecMap, FormatMap) are passed from the upper code instead of creating them in the pipeline. Such design assumes that the user can somehow configure them map and then pass it to the pipeline, which will use the map to create instances of codecs/parsers/etc.

ResamplerMap currently does not have any configuration options, but I suggest to use the same approach for all maps for consistency.

src/modules/roc_pipeline/config.h
```diff
@@ -199,13 +206,20 @@ struct ReceiverConfig {
 
     //! Parameters common for all sessions.
     ReceiverCommonConfig common;
+
+    //! To specify which resampling algorithm will be used.
+    audio::ResamplerBackend resampler_backend;
```
gavv: We already have `resampler_backend` in ReceiverSessionConfig. Instead duplicating it here and then copying to ReceiverSessionConfig, I suggest to remove this field and use `default_session.resampler_backend` instead.

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -101,7 +101,14 @@ TEST(resampler, invalid_scaling) {
...
     MockReader reader;
-    ResamplerReader rr(reader, buffer_pool, allocator, config, ChMask, FrameSize);
+    /*
+    core::IAllocator& allocator,
+              const ResamplerConfig& config,
+              packet::channel_mask_t channels,
+              size_t frame_size
+    */
+    BuiltinResampler resampler(allocator, config, ChMask, FrameSize);
+    ResamplerReader rr(reader, resampler, buffer_pool, FrameSize);
```
gavv: We will need to check all available resampler backends in all tests. We already use this approach in FEC tests:

https://github.com/roc-project/roc/blob/42ed45ed714488f45ff36483e8baaa4f5f3829cc/src/tests/roc_fec/target_openfec/test_writer_reader.cpp#L212-L219

Every test has a for loop that iterates over all FEC schemes. FEC scheme is passed to CodecMap to create appropriate codec.

gavv:
Thanks again for the PR! This feature is quite important for us because the quality and performance of our builtin resampler is in question.

I've leaved some comments. Many of them are stylistic though.

gavv:
BTW, if it would be more convenient, we can split this into a few PRs. E.g. the first PR may add IResampler and ResamplerMap but only a single BuiltinResampler backend. The second PR may add SpeexBackend, etc.

CristobalM:
Sure!, I'll make those changes this weekend and try to make the build work

src/modules/roc_audio/resampler_map.cpp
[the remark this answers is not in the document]
CristobalM: > It's a bit surprising that `new_resampler()` changes the state of ResamplerMap and even destroys the previously created resampler. It is also dangerous given that we return a raw pointer to the user. If the user will store the pointer and then call `new_resampler()` again, the pointer will become invalid.
> 
> We also have `get_current_resampler()`, but it seems it's not used anywhere.
> 
> I suggest to make ResamplerMap stateless and remove `current_resampler` and `get_current_resampler()`. It's common for factory pattern.
> 
> `new_resampler()` can just return a raw pointer to the resampler and the user will be responsible to store it into UniquePtr. It will be allowed to share ResamplerMap instance and call `new_resampler()` to create several resamplers simultaneously.

You are right, I assumed there that every user would instantiate ResamplerMap to then use a concrete resampler and give them the responsibility to use the IResampler pointer well, but it might be confusing to use in that way. I made the change that you suggested.

> I also suggest to remove `valid()` from IResampler interface and check it here, before returning the pointer, like:
> 
> ```
> UniquePtr<BuiltinResampler> r(new (allocator) BuiltinResampler());
> if (!r || !r->valid()) {
>     return NULL;
> }
> return r.release();
> ```
> 
> This approach will give a guarantee that ResamplerMap may return only successfully constructed objects, which would be nice. We already use such approach in other factories.

I think it would be more convenient to keep the valid() method inside IResampler to avoid doing the check with every backend and instead doing it once after the switch statement.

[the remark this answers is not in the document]
gavv: > I think it would be more convenient to keep the valid() method inside IResampler to avoid doing the check with every backend and instead doing it once after the switch statement.

Makes sense. Let's keep it.

src/modules/roc_pipeline/converter.cpp
```diff
@@ -33,18 +33,30 @@ Converter::Converter(const ConverterConfig& config,
...
+
+        resampler_map_.reset(new (allocator) audio::ResamplerMap, allocator);
+        if (!resampler_map_) {
+            return;
+        }
+
+        resampler_.reset(resampler_map_->new_resampler(
+                             config.resampler_backend, allocator, config.resampler,
+                             config.input_channels, config.internal_frame_size),
                          allocator);
```
gavv: We should check for error here:

```
if (!resampler_) {
    return;
}
```

src/modules/roc_pipeline/receiver.cpp
```diff
@@ -224,7 +224,7 @@ bool Receiver::create_session_(const packet::PacketPtr& packet) {
         return false;
     }
 
-    const ReceiverSessionConfig sess_config = make_session_config_(packet);
+    ReceiverSessionConfig sess_config = make_session_config_(packet);
```
gavv: Why do we drop `const` here?

src/modules/roc_pipeline/converter.cpp
gavv: The same is true for other pipelines.

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -101,7 +101,14 @@ TEST(resampler, invalid_scaling) {
     enum { ChMask = 0x1, InvalidScaling = FrameSize };
 
     MockReader reader;
-    ResamplerReader rr(reader, buffer_pool, allocator, config, ChMask, FrameSize);
+    /*
+    core::IAllocator& allocator,
+              const ResamplerConfig& config,
+              packet::channel_mask_t channels,
+              size_t frame_size
+    */
```
gavv: Could you please remove commented code?

gavv:
Thanks! I've marked resolved most of the conversations and added a few more comments.

Regarding to https://github.com/roc-project/roc/pull/287#discussion_r347808406. Do you have plans on it in this PR?

src/modules/roc_audio/resampler_builtin.h
```diff
@@ -24,34 +26,14 @@
...
-
-    ResamplerConfig()
-        : window_interp(128)
-        , window_size(32) {
-    }
-};
-
 //! Resamples audio stream with non-integer dynamically changing factor.
-class Resampler : public core::NonCopyable<> {
+class BuiltinResampler : public IResampler {
```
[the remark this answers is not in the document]
CristobalM: IResampler already inherits from NonCopyable, should I remove that inheritance and add it to BuiltinResampler?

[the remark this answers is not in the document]
gavv: Oh, I didn't notice that. Yes, I think it's better to inherit it in implementations, not in the interface class. This does not really matter, but usually we do it this way.

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -100,147 +104,176 @@ TEST_GROUP(resampler) {
...
-    MockReader reader;
-    ResamplerReader rr(reader, buffer_pool, allocator, config, ChMask, FrameSize);
+    for (size_t n_backend = 0; n_backend < Test_n_resampler_backends; n_backend++) {
+        ResamplerBackend backend = Test_resampler_backends[n_backend];
 
-    CHECK(rr.valid());
+        MockReader reader;
+        core::UniquePtr<IResampler> resampler(
+            resampler_map.new_resampler(backend, allocator, config, ChMask, FrameSize),
+            allocator);
```
gavv: Let's also add a check that resampler was successfully constructed (to all tests):

```
core::UniquePtr<IResampler> resampler(...);
CHECK(resampler);
```

src/modules/roc_pipeline/converter.h
gavv: Or we can at least avoid allocating resampler map on heap and just use a local variable, so this:

```
        resampler_map_.reset(...)
        if (!resampler_map_) {
            return;
        }

        resampler_.reset(resampler_map_->new_resampler(...));
```

would become this:

```
        ResamplerMap resampler_map;

        resampler_.reset(resampler_map.new_resampler(...));
```

gavv:
LGTM! Could you also squash commits?

gavv:
Thanks!

Issue #162, opened 2019-02-20
Improve self-profiling support


gavv:
We have `ProfilingWriter` pipeline element. It implements `IWriter` interfaces and itself acts as a proxy for another `IWriter`. During that proxying, it measure the average `write()` time (i.e. average frame processing time) and periodically reports it via `roc_log()`.

This element is currently used only in the `roc-conv` command-line tool, which uses `pipeline::Converter` pipeline.

We need a few improvements here:

* First, we need an improvement in `ProfilingWriter` itself: we should  use moving average instead of the overall average.

* Second, we should make it optional. Currently it is enabled in `pipeline::Converter` unconditionally. Instead, we should add `--profiling` command-line option and add `ProfilingWriter` to the pipeline only if that option was passed. An example of a conditionally-added pipeline element is `Interleaver` in `pipeline::Sender`.

* Finally, we should also add `ProfilingWriter` element to `pipeline::Sender` and `pipeline::Receiver` and add `--profiling` option to `roc-send` and `roc-recv` command-line tools.

Docs:

* https://roc-project.github.io/roc/docs/internals/code_structure.html
* https://roc-project.github.io/roc/docs/internals/data_flow.html

RohanK99:
What is the moving average with respect to as in what is the rolling window that we want to calculate the average of? Is this based on the variable moving window that the resampler is operating on? If so how can we extrapolate that value for use to get the moving average for ProfileWriter::write

gavv:
Hi! Sorry for late reply.

> Is this based on the variable moving window that the resampler is operating on?

Not quite. ProfilingWriter is the user of Resampler and itself is used by the sound I/O code. Sound I/O code reads (usually) fixed-size frames from ProfilingWriter, and it, in turn, reads these frames from Resampler. Resampler fills the requested fixed-size frames by reading a necessary amount of smaller-size frames from underlying pipeline components and scaling them a bit.

ProfilingWriter just calculates number of samples per second which goes through it. Actually it doesn't bother what the frame size is (variable, fixed, etc.)

Currently it calculates the average speed from the beginning of times, which is not very convenient. And I'm suggesting to replace it with SMA, i.e. calculate the average speed only for the last N samples (e.g. for last `5*sample_rate` samples, i.e. last 5 seconds).

RohanK99:
Should the duration of the window in seconds (eg. 5 second window) be configurable as a command line option? or should I fix it as a constant somewhere and if so what should that constants value be?

[...]

PR #291, opened 2019-12-05
enhanced self profiling

RohanK99:
addressed issue #162


src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,72 @@
...
+    const double speed = (double)frame.size() / num_channels_ / elapsed * core::Second;
+
+    if (running_data_size_ < running_data_.size()) {
+        running_data_[running_data_head_++] = speed;
+        moving_avg_ = (moving_avg_ * running_data_size_ + speed);
+        moving_avg_ /= (++running_data_size_);
+    } else {
+        moving_avg_ +=
+            (speed - running_data_[running_data_head_]) / (running_data_.size());
+        running_data_[running_data_head_++] = speed;
```
gavv: It seems there is a problem with this approach: it doesn't take frame size into account.

We're storing last `running_data_.size()` frames, but we don't know what the frame size is and even can't assume that it is the same for every frame.

So, if I understand correctly, when sample rate is 44100 and frame size is 10 ms, we will calculate SMA for last 441 seconds instead of last 1 second.

The difficulty is that the frame size may come from the user of our library when they call `roc_receiver_read()` or `roc_sender_write()` and it may be anything.

Any ideas how to deal with it?

I'm thinking about the following scheme:

* Virtually split LogInterval into fixed-length chunks. Say, 100 chunks of 10 ms each one.
* Track average speed corresponding to each chunk.
* Calculate SMA for these chunks.

So, `running_data_` size is equal to number of chunks (100), and each `running_data_` element is equal to average speed (arithmetic mean) of all *frames* partially or completely *overlapping* with the *chunk* corresponding to the element.

Here is how it looks like:

```
|     chunk 1     |     chunk 2     |     chunk 3     |     chunk 4     |
| frame 1 | frame 2 | frame 3 |           frame 4             | frame 5 |
```

* `chunk 1` will be average of speed for `frame 1` and `frame 2`
* `chunk 2` will be average of speed for `frame 2`, `frame 3`, and `frame 4`
* `chunk 3` will be speed for `frame 4`
* `chunk 4` will be average of speed for `frame 4` and `frame 5`

To implement this, it should be enough to store just a few values:

* current chunk number
* number of samples already added to current chunk (offset in current chunk of the last frame end)
* number of frames already added to current chunk
* average speed for current chunk (arithmetic mean of all frames already added to current chunk)

When a new frame appears, we update the average speed of current chunk and check whether we should switch to the next one.

src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,72 @@
...
+    , running_data_head_(0)
+    , running_data_size_(0)
+    , moving_avg_(0)
+    , sample_rate_(sample_rate)
+    , num_channels_(packet::num_channels(channels)) {
+    if (num_channels_ == 0) {
+        roc_panic("profiling writer: n_channels is zero");
+    }
+    if (sample_rate_ == 0) {
+        roc_panic("profiling writer: sample_rate is zero");
```
gavv: `profiling writer` -> `profiler`

gavv:
Thanks for the PR! I left a few comments.

src/modules/roc_audio/profiler.cpp
gavv: (Note: when computing average speed of a chunk, we likely should weight frames according to the number of samples they contribute (overlap) to the chunk.)

gavv:
Could you also add unit tests for Profiler? It's algorithm is not trivial and likely will become even more complicated.

src/modules/roc_audio/profiling_reader.cpp
```diff
@@ -0,0 +1,41 @@
...
+void ProfilingReader::read(Frame& frame) {
+    profiler_.begin_frame(frame);
+
+    const core::nanoseconds_t elapsed = read_(frame);
+
+    profiler_.end_frame(frame, elapsed);
+}
+
+core::nanoseconds_t ProfilingReader::read_(Frame& frame) {
+    const core::nanoseconds_t start = core::timestamp();
```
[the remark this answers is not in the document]
gavv: Though, I you'll add unit tests for Profiler (https://github.com/roc-project/roc/pull/291#issuecomment-562073898), it probably makes sense to keep this as is, so that Profiler will not depend on real time and can be tested easily.

src/modules/roc_audio/profiler.cpp
RohanK99: So I have been thinking about this algorithm for a while and have realized that after the first second has elapsed and all the chunks are full, the next averages will be gross approximations. We will be removing the average speed of chunk 1 * (elapsed milliseconds) samples from the frame and adding frame.size() new samples to the chunk and recomputing the average. Assuming that there are more than 1 frame in that chunk then the number of samples removed will not be accurate and the next number of samples removed will also not be accurate as the chunks average speed has changed again. Are we okay with this degree of imprecision. 

The alternative mechanism that will be more memory intensive but will be more accurate would be creating a doubly linked list and each node having stored the number of samples and time elapsed. That way we can accurately remove the number of samples that we need to based on that specific frames average and not the approximate average from the chunks.

It would look something like this:
* note each node is a frame stats
* note we are computing average for 25ms to demonstrate the point.

node 1 (samples: 10, time: 10ms) ->  node 2 (samples: 15, time: 5ms) -> node 3 (samples: 20, time: 10ms).

now when we get a new frame that will be node 4 that will for example be node4(samples: 10, time: 5ms). What will happen is that we will chain node4 to the end. Then modify node 1 such that we remove 5 ms worth of samples from it so that node 1 will now look like node1(samples: 5, time: 5ms). So now the linked list still contains 25ms worth of data in it and the samples are accurate then the moving average can be recomputed. Once node1 has time: 0ms then it will be deleted and the tail ptr will point to node2. The moving average in this case will be the sum of all the samples / LogInterval seconds.

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,51 @@
...
+namespace {
+
+const core::nanoseconds_t interval = 1000;
+const int sample_rate = 300;
+const int num_channels = 1;
+core::HeapAllocator allocator;
+
+} // namespace
+
+TEST_GROUP(profiler) {};
```
RohanK99: Not sure how to initialize test group properly. Any advice would be appreciated.

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+TEST(profiler, test_moving_average) {
```
RohanK99: This test, covers every aspect of the algorithm yet it isn't very readable and looks like I have just picked magic values for the test. Is there any better way to test this and should I write any other tests for this algorithm?

src/modules/roc_audio/profiler.cpp
gavv: > So I have been thinking about this algorithm for a while and have realized that after the first second has elapsed and all the chunks are full, the next averages will be gross approximations.

I don't quite understand the problem. Could you provide an example?

How I see it:

```
|   chunk 1   |   chunk 2   |
| frame 1  | frame 2 part 1 | frame 2 part 2 |
```

Assume that our ring buffer is 2-chunks size and the last frame (frame 2) is larger than the remaining "space" (number of samples) in the ring buffer.

In this case frame 2 is split into two parts. The first part is equal to the "space" available in the ring buffer and the second part is the rest. Size of first part is used to weight the frame speed and to add it to chunk 2. Size of the second (remaining) part and speed of the frame are stored in the profiler state.

When profiler will be removing chunk 1 and adding chunk 3, it will find the stored size of the remaining part of last frame (frame 2 part 2) and its speed. It will add the speed weighted with size to the new chunk (chunk 3). If the remaining size is larger than the chunk size, this procedure should be repeated until the remaining part becomes smaller than chunk.

This way each frame will contribute to a chunk proportionally to the size of the part overlapping with the chunk.

Am I missing something?

src/tests/roc_audio/test_profiler.cpp
gavv: Test group is actually a class. If you need any setup/teardown procedure common to all tests, you can add it here. If you don't need one, you can leave it empty.

https://cpputest.github.io/manual.html#setup_teardown

src/tests/roc_audio/test_profiler.cpp
gavv: I think the approach is OK, but it would be better to replace magic constant and to replace expected values with an expression using those constants. So e.g. you're passing Speed1 and Speed2 and expecting average (Speed1 + Speed2) / 2.

src/modules/roc_audio/profiler.h
```diff
@@ -0,0 +1,88 @@
...
+             packet::channel_mask_t channels,
+             size_t sample_rate);
+
+    //! Init audio frame.
+    void begin_frame(size_t frame_size);
+
+    //! Profile frame speed
+    void end_frame(size_t frame_size,
+                   core::nanoseconds_t elapsed,
+                   const core::nanoseconds_t interval = LogInterval);
```
gavv: Can we make interval the argument of constructor instead of end_frame()? It would be strange to have a different interval for different frames. It would also be strange to have different intervals for log reporting and for calculations.

Also, we usually try to avoid default arguments.

src/modules/roc_audio/profiler.h
```diff
@@ -37,15 +37,14 @@ class Profiler : public core::NonCopyable<> {
     //! Initialization.
     Profiler(core::IAllocator& allocator,
              packet::channel_mask_t channels,
-             size_t sample_rate);
+             size_t sample_rate,
+             core::nanoseconds_t interval = LogInterval);
```
RohanK99: Since it would be a pain to create a test where we are actually using 1000000000 ns, I had to keep the default value so I can override the default when testing.

src/modules/roc_audio/profiler.h
```diff
@@ -37,15 +37,14 @@ class Profiler : public core::NonCopyable<> {
...
     void begin_frame(size_t frame_size);
 
     //! Profile frame speed
-    void end_frame(size_t frame_size,
-                   core::nanoseconds_t elapsed,
-                   const core::nanoseconds_t interval = LogInterval);
+    void end_frame(size_t frame_size, core::nanoseconds_t elapsed);
 
     //! For Testing Only
     double get_moving_avg() {
```
RohanK99: can this be implemented in header or should I move this implementation into the cpp

src/tests/roc_audio/test_profiler.cpp
RohanK99: Is there a way to inherit this class from Profiler like you can in Gtest test classes. This would allow me to better test the private member variables etc. and would make my test significantly cleaner and better.

RohanK99:
I am also confused whats breaking the build. I can see the error but not sure how I fix that. I tried a couple of casts but neither of them work. So any insight would be helpful

src/tests/roc_audio/test_profiler.cpp
gavv: I think no.

Also, in Roc we try to follow the known practice of testing public methods only (blackbox testing) and designing class interfaces with testability in mind. This approach forces you to think more on decomposition and hopefully leads to a better division of responsibilities between classes.

My rule of thumb is: if you can't test the class as a blackbox, likely it currently lies on a wrong abtraction level. You should either decrease it by extracting part of its functionality into a new class which you can test separately, or increase it by moving some functionality from higher levels into the class. There are exceptions, though.

What specific problems do you have with Profiler currently?

src/modules/roc_audio/profiler.h
gavv: I think both is OK.

src/modules/roc_audio/profiler.h
gavv: This problem appears because LogInterval is currently hardcoded in Profiler. We can just move this constant from Profiler to Pipeline (config.h). Pipeline classes will pass it to profilers they create. Tests will pass their own constants. Profiler will use this constant both for rate limiter and for calculations.

gavv:
This page describe how you can run CI builds locally: https://roc-project.github.io/roc/docs/development/continuous_integration.html#run-locally

You can adjust the command according to the failing build.

```
src/tests/roc_audio/test_profiler.cpp:41:66: error: implicit conversion changes
      signedness: 'long' to 'unsigned long' [-Werror,-Wsign-conversion]
  ...LONGS_EQUAL((uint64_t)profiler.get_moving_avg(), samples * core::Second / (uint64_t)time);
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/CppUTest/UtestMacros.h:165:33: note: expanded from macro
      'LONGS_EQUAL'
  LONGS_EQUAL_LOCATION(expected,actual,__FILE__, __LINE__)
                                ^
/usr/include/CppUTest/UtestMacros.h:171:70: note: expanded from macro
      'LONGS_EQUAL_LOCATION'
  { UtestShell::getCurrent()->assertLongsEqual((long)expected, (long)actua...
```

samples is size_t (unsigned). core::Second is nanoseconds_t which is int64_t (signed). Likely you need an explicit cast.

Probably you also need UNSIGNED_LONGS_EQUAL instead of LONGS_EQUAL.

src/tests/roc_audio/test_profiler.cpp
RohanK99: No worries. Would've liked that feature just to ensure that all internal calculations are being properly calculated but I can manage the same outcome using abstraction.

src/modules/roc_audio/profiler.cpp
RohanK99: Ah, I understand now. I believe that would work but you would need to keep track of more data than my method. I would suggest we stick to the method I implemented which I find is actually easier to understand and achieves the same outcome.

gavv: > but you would need to keep track of more data than my method

Well, I wanted to achieve exactly opposite :)

The current code in this PR introduces a linked list of frame nodes. For each read or write operation, we're allocating a node and also probably removing a node. That's a considerable burden to the allocator. We could use a pool (core::Pool) to avoid (re)allocations, but that feels like an overkill for just a frame speed profiler. I'd like to avoid it and use something more lightweight.

On the other hand, I was suggesting the following:

* have a fixed-sized (and rather small) ring buffer, allocated once at session creation
* have a few integer fields for the current profiler state

That's all. No allocations and run-time.

I guess I've failed to to explain the idea clearly, so here is a sample code: https://ideone.com/PEmCwu

It demonstrates how we can fill the ring buffer of fixed-sized chunks from variable-length frames. Then we can compute SMA from that ring buffer, but I omitted that part for brevity.

RohanK99: Actually I am still a little confused. I thought we want to profile for the last `LogInterval` seconds. In this case we are profiling the last 500 samples. `CHUNK_LENGTH` is in `nanoseconds_t` but the comment appears like its 50 samples / chunk which won't necessarily elapse in 50 nanoseconds.

gavv: Ah, sorry, nanoseconds_t is there by mistake. CHUNK_LENGTH should be size_t and it defines number of samples. Operating with samples instead of nanoseconds makes the implementation simpler.

LogInterval should be still defined as nanoseconds because each session can have its own sample rate and thus number of samples will be different. However we can convert from nanoseconds to samples if we know the sample rate and number of channels. These parameters are constant for a session, so we can perform the conversion at construction time.

See also #274, it introduces a helper for such conversion. But until it's merged, we can do it by hand.

RohanK99:
Hey sorry I haven't worked on this recently, I'm currently in school and will finish this ticket this upcoming summer sorry!

gavv:
@RohanK99 No problem, take your time.

RohanK99:
Hey so I started working on this ticket again and I'm just getting up to speed with migrating my code to the new codebase. I seem to be encountering a new issue where I get an undefined reference to all the PulseAudio functions in `module-roc-sink.c`. I am running `scons -Q --enable-pulseaudio-modules --build-3rdparty=openfec,pulseaudio` to build the project. I am running Ubuntu 20.04. It runs fine on the CI, but I'm not sure what the issue is on my end. I have libpulse-dev installed. Just seems like it won't link properly but not entirely sure. Let me know if you know how to fix this.

gavv:
It seems our modules are not building on pulseaudio 13.99.1, which is used on ubuntu 20.04 and is not included to our CI. I'll look at it. For now you can build without `--enable-pulseaudio-module`.

gavv:
I've pushed a fix to `develop`.

RohanK99:
Thanks for the fix! Works like a charm

RohanK99:
Hey @gavv seems like theres another build issue when running tests I get this error "error: openfec not found (see 'config.log' for details)" and opening config.log yields this.
```.sconf_temp/conftest_242836043263364393122001005940534440299.c:3:10: fatal error: of_openfec_api.h: No such file or directory
    3 | #include <of_openfec_api.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
scons: Configure: no
```

RohanK99:
I've modified the moving average calculation to use the discussed method earlier in this thread. Take a look and let me know if any modifications need to be made. Also is the CI down because my commit doesn't seem to be building on the CI.

gavv:
> I get this error "error: openfec not found (see 'config.log' for details)" and opening config.log yields this.

I guess you need either install openfec system-wide or use --build-3rdparty=openfec

https://github.com/roc-project/roc/issues/349#issuecomment-621047157

gavv:
> I've modified the moving average calculation to use the discussed method earlier in this thread. Take a look and let me know if any modifications need to be made.

Thanks! I'll take a look in upcoming days.

> Also is the CI down because my commit doesn't seem to be building on the CI.

Not sure what happened, it seems this PR on github somehow lost its "connection" to travis. Maybe try to reopen the PR.

RohanK99:
> > I get this error "error: openfec not found (see 'config.log' for details)" and opening config.log yields this.
> 
> I guess you need either install openfec system-wide or use --build-3rdparty=openfec
> 
> [#349 (comment)](https://github.com/roc-project/roc/issues/349#issuecomment-621047157)

The issue I have is slightly different. My build is working fine using 3rd party build of openfec and I also have a system wide installation of openfec. The issue I run into where it can't find openfec is specific to running tests.

RohanK99:
hey @gavv I was wondering if you had a chance to look at this PR or have any free time soon to look at this PR. I wanted to start working on another ticket and would prefer not to have 2 open PR's at the same time as it becomes hard for me to maintain.

gavv:
@RohanK99 sorry for delay. I'll take a look at it now.

src/modules/roc_audio/profiler.h
```diff
@@ -0,0 +1,66 @@
...
+#include "roc_core/rate_limiter.h"
+#include "roc_core/refcnt.h"
+#include "roc_core/time.h"
+#include "roc_packet/units.h"
+
+namespace roc {
+namespace audio {
+
+//! Profiler
+class Profiler : public core::NonCopyable<> {
```
gavv, line 29: Could you please briefly describe our algorithm in the class comment?

Important things to note:

* Briefly state the goal: periodically report average processing speed (# of samples processed per time unit) during last N seconds.

* Briefly state the problem: we want to calculate SMA efficiently (with O(1) complexity, without allocations, and as lightweight as possible), but we have variable-size frames and SMA needs fixed-size chunks.

* Briefly describe the ring buffer, what does the chunk represent, and how its value is calculated.

* Briefly describe how we use weighted mean, SMA, and CMA.

src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,92 @@
...
+            last_chunk_samples = 0;
+            chunks_[last_chunk_num] = 0;
+        }
+
+        frame_size -= n_samples;
+    }
+
+    if (rate_limiter_.allow()) {
+        roc_log(LogDebug,
+                "profiler avg for last %llu sec : %lu sample/sec (%.2f sec/sec)",
```
gavv, line 85: Nitpick:

`profiler avg` -> `profiler: avg`
`sec :` -> `sec:`

(to follow our logging conventions)

src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,92 @@
...
+        roc_panic("profiler: sample_rate is zero");
+    }
+    chunks_.resize(num_chunks_);
+}
+
+void Profiler::begin_frame(size_t frame_size) {
+    if (frame_size % num_channels_ != 0) {
+        roc_panic("profiler: unexpected frame size");
+    }
+}
```
gavv, line 42: It seems we don't need this method anymore. Maybe move the check to end_frame() and remove begin_frame()? We can then rename end_frame() to add_frame() or something like this.

src/modules/roc_pipeline/converter_source.cpp
```diff
@@ -68,6 +68,17 @@ ConverterSource::ConverterSource(const ConverterConfig& config,
         areader = pipeline_poisoner_.get();
     }
 
+    if (config.profiling) {
+        profiler_.reset(new (allocator) audio::ProfilingReader(
```
gavv, line 72: Since profiler is now on top of poisoner, we should also move `profiler_` field after `pipeline_poisoner_` field here: https://github.com/roc-project/roc/blob/7a802f6e966e0702a2fba6c6b4b3adef0e5fbf1b/src/modules/roc_pipeline/converter_sink.h

We need it to ensure that poisoner is not destroyed before profiler, because profiler object uses poisoner object (although in current profiler implementation doesn't use nested reader, it may change in future).

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,61 @@
...
+namespace {
+
+struct TestFrame {
+    TestFrame(size_t size, core::nanoseconds_t time)
+        : size_(size)
+        , time_(time) {
+    }
+
+    size_t size_;
+    core::nanoseconds_t time_;
```
gavv, line 26: Nitpick: these fields are public, so they should not have trailing underscore.

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const core::nanoseconds_t millisecond = 1000000L;
```
gavv, line 29: Nitpick: can we use core::Millisecond instead?

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,61 @@
...
+        TestFrame(50, 50 * core::Second),      TestFrame(25, 25 * core::Second),
+        TestFrame(25, 25 * core::Second),      TestFrame(25, 25 * core::Second),
+        TestFrame(25, 25 * core::Second / 2),  TestFrame(40, 40 * core::Second),
+        TestFrame(60, 60 * core::Second / 3),  TestFrame(50, 50 * core::Second),
+        TestFrame(125, 125 * core::Second / 3)
+    };
+
+    // hand calculated values
+    double expected_moving_average[9] = { 1.000, 1.000, 1.000, 1.000, 1.167,
+                                          1.167, 1.580, 1.580, 2.280 };
```
gavv, line 52: With this hard-coded values, it's hard to verify whether the test is correct. Could we replace them with explicit expressions, like `(a + b + c) / 3` (where a, b, c are the numbers)?

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,61 @@
...
+core::HeapAllocator allocator;
+
+} // namespace
+
+TEST_GROUP(profiler) {};
+
+TEST(profiler, test_moving_average) {
+    Profiler profiler(allocator, num_channels, sample_rate, interval);
+
+    TestFrame frames[9] = {
```
gavv, line 42: Nitpick: we can omit `9` here and in expected_moving_average. And we can use ROC_ARRAY_SIZE in the for loop. Also I'd add an assertion that the size of the two arrays is the same.

src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,92 @@
...
+namespace audio {
+
+Profiler::Profiler(core::IAllocator& allocator,
+                   packet::channel_mask_t channels,
+                   size_t sample_rate,
+                   core::nanoseconds_t interval)
+    : rate_limiter_(interval)
+    , interval_(interval)
+    , chunk_length_(sample_rate / 100)
+    , num_chunks_((size_t)(interval / (core::Second / 100)) + 1)
```
gavv, line 24: So we have 10ms chunks. Can we make this 10ms a constant? (local to profiler.cpp I think)

gavv: Why are we always adding 1 to num_chunks? Maybe you need this instead? https://stackoverflow.com/questions/17944/how-to-round-up-the-result-of-integer-division

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,61 @@
...
+        TestFrame(50, 50 * core::Second),      TestFrame(25, 25 * core::Second),
+        TestFrame(25, 25 * core::Second),      TestFrame(25, 25 * core::Second),
+        TestFrame(25, 25 * core::Second / 2),  TestFrame(40, 40 * core::Second),
+        TestFrame(60, 60 * core::Second / 3),  TestFrame(50, 50 * core::Second),
+        TestFrame(125, 125 * core::Second / 3)
```
gavv, line 47: If I understand correctly, our chunks size is 10ms, and all frames in this test are larger than 10ms. So the case when the frame is smaller than the chunk is not covered. We should cover add either in this test or add a separate test.

RohanK99: @gavv this does cover cases where the entire chunk is not covered. The way the algorithm you linked to works is that it is contingent on frame speed and frame samples. In the cases like 25*core::Second/2 this is covering less than one chunk. The time elapsed is only used to calculate the speed for those samples and then according to sample rate the speeds of the frames are populated into each chunk. I have the rudimentary implementation here that expanded on top of your code that you provided earlier. I'll wait till you respond to this to implement the rest of the feedback incase this is not the approach we were intending.  https://ideone.com/h5QPck

gavv: Ah, I see now. Looks good then.

gavv: BTW, please don't resolve conversations by yourself, so that the reviewer could go through unresolved conversations next time. (This convention is often used with github and gitlab). To mark them as done, you can leave a comment or use "reactions".

src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,92 @@
...
+            (frame_speed - last_chunk_speed) / last_chunk_samples * n_samples;
+
+        // last chunk is full
+        if (last_chunk_samples == chunk_length_) {
+            last_chunk_num = (last_chunk_num + 1) % num_chunks_;
+
+            // ring buffer is full
+            if (last_chunk_num == first_chunk_num) {
+                moving_avg_ +=
+                    (last_chunk_speed - chunks_[first_chunk_num]) / (num_chunks_ - 1);
```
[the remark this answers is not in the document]
RohanK99: It is num_chunks_ -1 because when we initialize num_chunks_ we have to add 1 to total size and the reason for this is because the moving average calculation takes the average of the last n samples and sets the current one to 0. This means that theres only num_chunks_ - 1 previous samples that are valid as the newest one is currently being written to. You can check the behaviour by changing the value of num_chunks_ using this: https://ideone.com/h5QPck

src/modules/roc_audio/profiler.cpp
RohanK99: > Why are we always adding 1 to num_chunks? Maybe you need this instead? https://stackoverflow.com/questions/17944/how-to-round-up-the-result-of-integer-division

refer to https://github.com/roc-project/roc/pull/291/files/7a802f6e966e0702a2fba6c6b4b3adef0e5fbf1b#r424274466

RohanK99:
Hey @gavv my tests still don't work and I was wondering if you had any insight. Since I started working on this ticket again my tests have completely stopped working. I run `scons -Q --enable-werror --enable-pulseaudio --build-3rdparty=openfec,pulseaudio test` and not a single test runs it just finishes the command with no output to the console and I am extremely confused why this is happening. Previously I also could just run `scons -Q test` and it was able to run the tests but now running that I get an error that openfec could not be found even though it was built as 3rd party. I have been relying on the travis builds for feedback but that is quite time consuming an inconvenient. I decided to get an 18.04 image going incase that was my issue, I replicated all the steps in the documentation only to encounter the same problem. I was wondering if you had any idea why this might be or how to fix it. Additionally locally running it doesn't print all the warnings a lot of warnings only show up on travis which I find fairly strange.

RohanK99:
> Hey @gavv my tests still don't work and I was wondering if you had any insight. Since I started working on this ticket again my tests have completely stopped working. I run `scons -Q --enable-werror --enable-pulseaudio --build-3rdparty=openfec,pulseaudio test` and not a single test runs it just finishes the command with no output to the console and I am extremely confused why this is happening. Previously I also could just run `scons -Q test` and it was able to run the tests but now running that I get an error that openfec could not be found even though it was built as 3rd party. I have been relying on the travis builds for feedback but that is quite time consuming an inconvenient. I decided to get an 18.04 image going incase that was my issue, I replicated all the steps in the documentation only to encounter the same problem. I also used the same commands as used in the docker files / 18.04.sh and still had the same problem. I was wondering if you had any idea why this might be or how to fix it. Additionally locally running it doesn't print all the warnings a lot of warnings only show up on travis which I find fairly strange.

solved my own problem looks like i had to `--enable-tests`

RohanK99:
I have implemented all the feedback. It is ready for review again. I can squash the commits once its ready to merge.

gavv:
Yeah, you now need `--enable-tests` since tests are now disabled by default (they are not needed for regular users).

I think we should report a warning or an error when using `scons test` without `--enable-tests`. Added to my TODO.

As for the warnings, travis build includes may different compiler versions and target architectures, that's why there are more warnings. I'd suggest to install at least two: recent clang and recent gcc, and check both before pushing. You can also run travis builds locally using docker.

See details here: https://roc-project.github.io/roc/docs/development/continuous_integration.html

Here are uptodate developer instructions (for develop branch): https://github.com/roc-project/roc/blob/develop/docs/sphinx/building/developer_cookbook.rst

gavv:
> I can squash the commits once its ready to merge.

See my comment here: https://github.com/roc-project/roc/pull/350#issuecomment-626325657

src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,112 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_audio/profiler.h"
+#include "roc_core/log.h"
+#include "roc_core/panic.h"
+#include "roc_core/shared_ptr.h"
+
+namespace {
+const size_t chunk_time = 10;
```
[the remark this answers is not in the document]
gavv: Also, I suggest to make both parameters of type nanoseconds_t for convenience (so it'd be easy to compare them with other parameters like frame size when reading code).

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,107 @@
...
+         + (0.5 * frame_speeds[3] + 0.5 * frame_speeds[4])
+         + (0.8 * frame_speeds[5] + 0.2 * frame_speeds[6]) + frame_speeds[6])
+        / 5;
+
+    // 1st chunk overwritten
+    expected_average[7] = ((0.5 * frame_speeds[1] + 0.5 * frame_speeds[2])
+                           + (0.5 * frame_speeds[3] + 0.5 * frame_speeds[4])
+                           + (0.8 * frame_speeds[5] + 0.2 * frame_speeds[6])
+                           + frame_speeds[6] + frame_speeds[7])
+        / 5;
```
gavv, line 93: Maybe set 7th frame to something other than `TestFrame(50, 50 * core::Second)`? Currently it seems that `frame_speeds[0]` are `frame_speeds[7]` equal, so `expected_average[6]` and `expected_average[7]` are also equal and theoretically we wont note a bug here?

src/modules/roc_audio/profiler.cpp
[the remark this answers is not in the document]
gavv: I see now. The new version of tests looks much more clear, thanks!

So we're ignoring the very last (incomplete) chunk, right? And our report is actually always outdated by 10ms.

Well, this is acceptable, but isn't there an easy way to avoid it? E.g. maybe we can just take into account last_chunk_speed and last_chunk_size in get_moving_avg() and "add" them to the returned value? (And also use get_moving_avg() when printing to log).

This way we wont have that 10ms lag in our reports, which would be nice.

src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+void Profiler::add_frame(size_t frame_size, core::nanoseconds_t elapsed) {
+    roc_panic_if(!valid_);
+
+    if (frame_size % num_channels_ != 0) {
+        roc_panic("profiler: unexpected frame size");
+    }
+
+    const double frame_speed =
```
gavv, line 64: Nitpick: maybe it makes sense to extract the calculation (lines 64-101) into a private method, say, update_avg_().

gavv:
https://github.com/roc-project/roc/pull/291#discussion_r425643733

src/tests/roc_audio/test_profiler.cpp
RohanK99: We've tested every case including larger than one chunk, less than one chunk spanning more than 2 chunks etc. The reason that one exists is the first time it is filled the full buffer and it is ensuring that it overwrites the first chunk. So I would argue that it isn't testing the same thing as the first frame.

src/modules/roc_audio/profiling_reader.cpp
```diff
@@ -0,0 +1,41 @@
...
+                                 core::nanoseconds_t interval)
+    : profiler_(allocator, channels, sample_rate, interval)
+    , reader_(reader) {
+}
+
+bool ProfilingReader::read(Frame& frame) {
+    const core::nanoseconds_t elapsed = read_(frame);
+
+    profiler_.add_frame(frame.size(), elapsed);
+    return true;
```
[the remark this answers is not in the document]
RohanK99: added this, but not entirely sure that makes sense the comment says false is when theres nothing left to read so shouldn't add_frame be called when reader_.read returns true. I might be wrong.

RohanK99:
Should be ready for review again :)

RohanK99:
Additionally, build will fail and this is the error: `error: implicit conversion turns
      floating-point number into integer: 'float' to 'const size_t' (aka
      'const unsigned long') [-Werror,-Wfloat-conversion]
  ...* ((float)profiler_config.chunkDuration / (float)core::Second))` I find these implicit conversions the hardest to fix on the Android ARM platform. Let me know if you have any insight on how to fix this, thank you.

src/modules/roc_audio/profiling_reader.cpp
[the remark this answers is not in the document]
gavv: Sorry, a typo.

I meant, "if reader. read() **returned false**, this means there's nothing to read and we shouldn't call add_frame in this case".

src/modules/roc_audio/profiler.h
```diff
@@ -0,0 +1,109 @@
...
+    ProfilerConfig(core::nanoseconds_t interval, core::nanoseconds_t chunk_duration)
+        : profilingInterval(interval)
+        , chunkDuration(chunk_duration) {
+    }
+
+    //! Rolling window duration and reporting interval.
+    core::nanoseconds_t profilingInterval;
+    
+    //! Duration of samples each chunk can hold in the circular buffer.
+    core::nanoseconds_t chunkDuration;
```
gavv, line 47: We use snake_case for fields. CamelCase is used only for classes and constants.

gavv:
Thanks, I'll take a look tomorrow.

src/modules/roc_pipeline/config.h
```diff
@@ -90,6 +91,12 @@ struct SenderConfig {
     //! Fill unitialized data with large values to make them more noticable.
     bool poisoning;
 
+    //! Profile moving average of frames being written.
+    bool profiling;
+
+    //! Profiler configuration.
+    struct audio::ProfilerConfig profiler_config;
```
gavv, line 98: Here and in other places, `struct` keyword is not needed. BTW, `profiler_config()` initializer is redundant too.

src/modules/roc_audio/profiling_reader.cpp
```diff
@@ -0,0 +1,48 @@
...
+    return ret;
+}
+
+core::nanoseconds_t ProfilingReader::read_(Frame& frame) {
+    const core::nanoseconds_t start = core::timestamp();
+
+    reader_.read(frame);
+
+    return core::timestamp() - start;
+}
```
gavv, line 41: This code does something fundamentally wrong :)

What we do here:

* call reader_.read() once, and ignore its return value
* call reader_.read() once more (?) and store its return value
* if the second call failed (?), call add_frame()
* return stored return value from the second call...

What we should do instead:

1. call reader_.read() and store its return value
2. if the return value was true, i.e. if reader did the actual work, call add_frame()
3. return stored return value

The idea of (2) is to skip add_frame() call if read() did not perform any computations, so that its early and likely very fast return wouldn't affect our statistics. The idea of (3) is to forward return value from nested reader to the caller, so that the presence of the profiler in the pipeline would be transparent.

Sorry for not being clear first time.

What is really bad here is that we don't have pipeline integration tests that catch such a serious bug as calling read() twice instead of once(). That's because they don't enable profiler.

Those tests are located at `src/tests/roc_pipeline`. Could you please enabled profiling in them? (and ensure that they will catch this bug BTW).

src/tests/roc_audio/test_profiler.cpp
```diff
@@ -0,0 +1,107 @@
...
+    TestFrame(size_t sz, core::nanoseconds_t tm)
+        : size(sz)
+        , time(tm) {
+    }
+
+    size_t size;
+    core::nanoseconds_t time;
+};
+
+const int sampleRate = 5000; // 50 samples / chunk
```
gavv, line 28: Nitpick: camel case constants and class names starts with capital letter, i.e. `SampleRate`. I think I should document all these little conventions...

gavv:
What do you think about this suggestion? https://github.com/roc-project/roc/pull/291#discussion_r425643733

RohanK99:
> What do you think about this suggestion? [#291 (comment)](https://github.com/roc-project/roc/pull/291#discussion_r425643733)

Forgot to reply to this. Essentially the algorithm isn't actually lagging behind 10ms and we're not ignoring the absolute last chunk. Whats happening is that while a chunk is not completely full aka the next 10ms hasn't elapsed we keep the last N chunks to have the moving average for the last N*10ms. So in this sense we need N+1 chunks. N for the previous data and 1 for the new data. So our algorithm is not lagging behind 10ms because as soon as the new chunk gets populated so does our moving average. We aren't exactly ignoring the last chunk, more just waiting for it to get filled. Its like the buffer of the buffer. Hope that makes sense.

gavv:
> We aren't exactly ignoring the last chunk, more just waiting for it to get filled.

Yeah, that's what I'm talking about.

If you call read() and then call get_moving_avg(), the returned avg wont take into account the last 10ms of samples.

E.g. if frame size = chunk size = 10ms, get_moving_avg() wont take into account the last frame. That frame will be taken into account only when a new frame is written or read.

So by "lagging" I mean that at any point of time, the reported average is behind the *actual* mathematical average by 10ms. Or, in other words, any batch of samples will contribute to the reported average only after next 10ms of samples is processed.

Again, that's not really a big deal, but I feel that technically it's easy enough to fix this small problem (we only need to fix get_moving_avg() and tests, and the main algorithm can be kept as is). If this is so, why not to fix it and make the algorithm results a bit more clear?

Let me know if I'm getting something wrong.

src/modules/roc_audio/profiling_reader.cpp
```diff
@@ -0,0 +1,48 @@
...
+                                 ProfilerConfig profiler_config)
+    : profiler_(allocator, channels, sample_rate, profiler_config)
+    , reader_(reader) {
+}
+
+bool ProfilingReader::read(Frame& frame) {
+    bool ret;
+    const core::nanoseconds_t elapsed = read_(frame, ret);
+
+    if (!ret) {
```
gavv, line 29: `!ret` -> `ret`

src/modules/roc_audio/profiling_reader.h
```diff
@@ -0,0 +1,51 @@
...
+                    core::IAllocator& allocator,
+                    packet::channel_mask_t channels,
+                    size_t sample_rate,
+                    ProfilerConfig profiler_config);
+
+    //! Read audio frame.
+    virtual bool read(Frame& frame);
+
+    //! Check if the profiler was succefully constructed.
+    virtual bool valid() const;
```
gavv, line 39: virtual is not needed here.

src/modules/roc_audio/profiling_writer.h
```diff
@@ -25,24 +26,23 @@ namespace audio {
...
+                    core::IAllocator& allocator,
+                    packet::channel_mask_t channels,
+                    size_t sample_rate,
+                    ProfilerConfig profiler_config);
 
     //! Write audio frame.
     virtual void write(Frame& frame);
 
+    //! Check if the profiler was succefully constructed.
+    virtual bool valid() const;
```
gavv, line 39: virtual is not needed here.

RohanK99:
> > We aren't exactly ignoring the last chunk, more just waiting for it to get filled.
> 
> Yeah, that's what I'm talking about.
> 
> If you call read() and then call get_moving_avg(), the returned avg wont take into account the last 10ms of samples.
> 
> E.g. if frame size = chunk size = 10ms, get_moving_avg() wont take into account the last frame. That frame will be taken into account only when a new frame is written or read.
> 
> So by "lagging" I mean that at any point of time, the reported average is behind the _actual_ mathematical average by 10ms. Or, in other words, any batch of samples will contribute to the reported average only after next 10ms of samples is processed.
> 
> Again, that's not really a big deal, but I feel that technically it's easy enough to fix this small problem (we only need to fix get_moving_avg() and tests, and the main algorithm can be kept as is). If this is so, why not to fix it and make the algorithm results a bit more clear?
> 
> Let me know if I'm getting something wrong.

It will account for the last 10ms of samples "IF" the last chunk is full. If you would like precise readings within that 10ms frame though, it would require reworking a significant amount of change in the algorithm, Currently if you see below we only update the moving average when the chunk is full. So if we were to change this we would have to change how we calculate moving average as well. Right now moving average is updated every time a chunk is filled. 

```
// last chunk is full
        if (last_chunk_samples_ == chunk_length_) {
            last_chunk_num_ = (last_chunk_num_ + 1) % num_chunks_;

            // ring buffer is full
            if (last_chunk_num_ == first_chunk_num_) {
                // Simple Moving Average: https://en.wikipedia.org/wiki/Moving_average
                moving_avg_ +=
                    (last_chunk_speed - chunks_[first_chunk_num_]) / (num_chunks_ - 1);
                first_chunk_num_ = (first_chunk_num_ + 1) % num_chunks_;
            } else {
                // Cumulative Moving Average: https://en.wikipedia.org/wiki/Moving_average
                moving_avg_ = ((moving_avg_ * (last_chunk_num_ - 1) + last_chunk_speed)
                               / last_chunk_num_);
            }

            last_chunk_samples_ = 0;
            chunks_[last_chunk_num_] = 0;
        }
```

src/modules/roc_audio/profiler.cpp
```diff
@@ -0,0 +1,113 @@
...
+namespace audio {
+
+Profiler::Profiler(core::IAllocator& allocator,
+                   packet::channel_mask_t channels,
+                   size_t sample_rate,
+                   ProfilerConfig profiler_config)
+    : rate_limiter_(profiler_config.profiling_interval)
+    , interval_(profiler_config.profiling_interval)
+    , chunk_length_((size_t)(
+          sample_rate * ((float)profiler_config.chunk_duration / (float)core::Second)))
```
gavv, line 24: Let's use `packet::ns_to_size()` here.

gavv:
> If you would like precise readings within that 10ms frame though

Yep.

> it would require reworking a significant amount of change in the algorithm

As I suggested above, can't we just take the last incomplete chunk into account in get_moving_avg()? Something like this:

```
float get_moving_avg() const {

    const size_t num_samples_in_moving_avg = (chunk_length_ * (num_chunks_ - 1 )); 

    return (moving_avg_ * num_samples_in_moving_avg + chunks_[last_chunk_num_])
                               / (num_samples_in_moving_avg + last_chunk_samples_);
}
```

gavv:
(If this simple solution wont work for some reason, let's keep it as is).

gavv:
Sorry for conflicts, was delivering hotfix to master. Rebase should be easy though, you can just --skip conflicting commits.

RohanK99:
done :) ready for review

gavv:
Now, when we take the last chunk into account in get_moving_avg(), shouldn't we remove "+1" from here?

```
num_chunks_((size_t)((unsigned long)profiler_config.profiling_interval
                           / (unsigned long)(profiler_config.chunk_duration))
                  + 1)
```

Is the current num_chunks_ correct?

I tried to remove "+1" and the tests are still passing. Actually they're passing even if I replace "+1" with "-1" or "+2" :) I think we should cover the overflow case with more care, covering corner case in tests is very important.

docs/sphinx/manuals/roc_send.rst
```diff
@@ -37,6 +37,7 @@ Options
 --resampler-window=INT      Number of samples per resampler window
 --interleaving              Enable packet interleaving  (default=off)
 --poisoning                 Enable uninitialized memory poisoning (default=off)
+--poisoning                 Enable uninitialized memory poisoning (default=off)
```
gavv, line 40: A typo: should be --profiling.

BTW, you can generate this piece of RST using this command:

```
roc-send -h | scripts/help2rst.py
```

gavv:
The rest looks good!

RohanK99:
> Now, when we take the last chunk into account in get_moving_avg(), shouldn't we remove "+1" from here?
> 
> ```
> num_chunks_((size_t)((unsigned long)profiler_config.profiling_interval
>                            / (unsigned long)(profiler_config.chunk_duration))
>                   + 1)
> ```
> 
> Is the current num_chunks_ correct?
> 
> I tried to remove "+1" and the tests are still passing. Actually they're passing even if I replace "+1" with "-1" or "+2" :) I think we should cover the overflow case with more care, covering corner case in tests is very important.

Oh good catch the +1 will still be required. Turns out I was using LONGS_EQUALS instead of DOUBLES_EQUAL. Should be good now

gavv:
Great.

Yep, +1 is really required.

But after playing a bit more with the tests, it seems we forgot one more small thing. When we're adding samples from the last chunk in get_moving_avg(), we should also subtract the same number of samples from the first chunk, so that our moving average is always calculated for `profiling_interval` samples.

We can even see this from the tests:

```
    expected_average[8] = ((((0.8 * frame_speeds[5] + 0.2 * frame_speeds[6])
                             + frame_speeds[6] + frame_speeds[7] + frame_speeds[8] * 2)
                            / 5)
                               * samples_in_moving_avg
                           + frame_speeds[8] * (frames[8].size - 100))
        / (samples_in_moving_avg + (frames[8].size - 100));
```

Here we expect the number of samples to be `(samples_in_moving_avg + (frames[8].size - 100))`, i.e. 250 + 25, but actually it always should be 250.

Consider the following fix:

```
float Profiler::get_moving_avg() {
    if (!buffer_full_) {
        const size_t num_samples_in_moving_avg = (chunk_length_ * last_chunk_num_);

        return (moving_avg_ * num_samples_in_moving_avg
                + chunks_[last_chunk_num_] * last_chunk_samples_)
            / (num_samples_in_moving_avg + last_chunk_samples_);
    } else {
        const size_t num_samples_in_moving_avg = (chunk_length_ * (num_chunks_ - 1));

        return (moving_avg_ * num_samples_in_moving_avg
                - chunks_[first_chunk_num_] * last_chunk_samples_
                + chunks_[last_chunk_num_] * last_chunk_samples_)
            / num_samples_in_moving_avg;
    }
}
```

```
    // 2nd and 3rd chunk overwritten 4th partially populated
    expected_average[8] =
        ((((0.8 * frame_speeds[5] + 0.2 * frame_speeds[6]) + frame_speeds[6]
           + frame_speeds[7] + frame_speeds[8] * 2)
          / 5)
             * samples_in_moving_avg
         - (0.8 * frame_speeds[5] + 0.2 * frame_speeds[6]) * (frames[8].size - 100)
         + frame_speeds[8] * (frames[8].size - 100))
        / samples_in_moving_avg;
```

RohanK99:
> expected_average[8] =
>         ((((0.8 * frame_speeds[5] + 0.2 * frame_speeds[6]) + frame_speeds[6]
>            + frame_speeds[7] + frame_speeds[8] * 2)
>           / 5)
>              * samples_in_moving_avg
>          - (0.8 * frame_speeds[5] + 0.2 * frame_speeds[6]) * (frames[8].size - 100)
>          + frame_speeds[8] * (frames[8].size - 100))
>         / samples_in_moving_avg;

I agree, I have added this fix.

gavv:
Rohan, thanks a lot for finishing this! It will help us to troubleshoot issues like #299.

RohanK99:
No problem! Thanks for all the help :)

Issue #290, opened 2019-12-02
Print thread ID in log messages


gavv:
## Problem

It would be convenient to include integral thread identifier in log messages.

Currently messages look like this:

```
13:36:34.803 [dbg] roc_sndio: initializing pulseaudio backend
```

and with thread ID they should look like:

```
13:36:48.977 [7843] [dbg] roc_sndio: initializing pulseaudio backend
```

This thread ID can be used to distinguish threads, and, at least on Linux, to find these threads in ps/top and to send signals to them.

## Solution

We can use `syscall(SYS_gettid)` on Linux and `pthread_getthreadid_np()` on OSX and (in future) on BSD. Both function are non-portable.

Steps:

* Add `target_linux` and `target_bsd` to [SConstruct](https://github.com/roc-project/roc/blob/42ed45ed714488f45ff36483e8baaa4f5f3829cc/SConstruct#L601). `target_linux` should be enabled if `platform` is `linux` or `android`. `target_bsd` should be enabled if `plarform` is `darwin`. 

* Add `core::format_tid()`, similar to `core::format_time()`. It should have two implementations: `src/modules/roc_core/target_linux/roc_core/format_tid.cpp` (using SYS_gettid) and `src/modules/roc_core/target_bsd/roc_core/format_tid.cpp` (using pthread_getthreadid_np).

* Use `core::format_tid()` in `core::Logger::print()`.

dshil:
@gavv, what about using `uv_thread_self`?

dshil:
@gavv, what about using `uv_thread_self`?

gavv:
@dshil it's portable, but, just like pthread_self(), it returns an opaque handle that is not an integral type. It can't be printed, or used to find the thread in ps/top, or used to to send a signal to the thread.

gavv:
Assigned to @RohanK99.

PR #295, opened 2019-12-08
Added Thread-ID to logging

RohanK99:
Addressed issue #290. Tried compiling for mac and it seemed that I couldn't include `#include <pthread_np.h>` in order to use `pthread_getthreadid_np()` so I had to add `#define` flags around it to check for BSD otherwise Linux wouldn't compile either. Might be doing something wrong in regards to the BSD/Mac to obtain the thread id. Also unsure if I should make the thread id's return an `int` or a `pid_t`.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -609,6 +609,7 @@ else:
     if platform in ['linux', 'android']:
         env.Append(ROC_TARGETS=[
             'target_posixtime',
+            'target_linux',
         ])
 
     if platform in ['linux']:
--- a/src/modules/roc_core/format_tid.h
+++ b/src/modules/roc_core/format_tid.h
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_core/format_tid.h
+//! @brief Retrieve the current thread id.
+
+#ifndef ROC_CORE_FORMAT_TID_H_
+#define ROC_CORE_FORMAT_TID_H_
+
+namespace roc {
+namespace core {
+
+//! Retrieve and format current time.
+//
+//!
+//! @returns
+//!  if obtaining thread id was successful
+//!
+//! @note
+//!  This function should not log anything because it is used
+//!  in the logger implementation.
+bool format_tid(char* buf, size_t bufsz);
+
+} // namespace core
+} // namespace roc
+
+#endif // ROC_CORE_FORMAT_TID_H_
--- a/src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
+++ b/src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <pthread.h>
+#include <stdint.h>
+#include <stdio.h>
+
+#include "roc_core/format_tid.h"
+
+namespace roc {
+namespace core {
+
+bool format_tid(char* buf, size_t bufsz) {
+    uint64_t tid = 0;
+    pthread_threadid_np(NULL, &tid);
+    if (snprintf(buf, bufsz, "%llu", tid) < 0)
+        return false;
+    return true;
+}
+
+} // namespace core
+} // namespace roc
--- a/src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
+++ b/src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
@@ -17,10 +17,9 @@ namespace core {
 
 bool format_tid(char* buf, size_t bufsz) {
     uint64_t tid = 0;
-    pthread_threadid_np(NULL, &tid);
-    if (snprintf(buf, bufsz, "%llu", tid) < 0)
-        return false;
-    return true;
+    int ret_tid = pthread_threadid_np(NULL, &tid);
+    int ret_buff = snprintf(buf, bufsz, "%llu", (unsigned long long)tid);
+    return ret_tid == 0 && ret_buff > 0 && (size_t)ret_buff < bufsz;
 }
 
 } // namespace core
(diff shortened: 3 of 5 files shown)


SConstruct
```diff
@@ -604,11 +604,14 @@ else:
             'target_stdio',
             'target_gcc',
             'target_libuv',
+            'target_linux',
+            'target_bsd',
```
gavv: With this two lines we enable target_linux and target_bsd on all three platforms: linux, android, macos. We should not do it. We should enable target_linux only on linux and android, and target_bsd only on macos (darwin). Since we already do it below, these two lines can be just removed.

src/modules/roc_core/format_tid.h
```diff
@@ -0,0 +1,34 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_core/format_tid.h
+//! @brief Retrieve the current thread.
+
+#ifndef ROC_CORE_FORMAT_TID_H_
+#define ROC_CORE_FORMAT_TID_H_
+
+#include <sys/types.h>
```
gavv: `sys/types.h` header is not portable. Since this file (src/modules/roc_core/format_tid.h) does not belong to a target directory, we should include only portable headers here. I guess we can just remove this one.

src/modules/roc_core/format_tid.h
```diff
@@ -0,0 +1,34 @@
...
+//! Retrieve and format current time.
+//
+//!
+//! @returns
+//!  the thread id as an int
+//!
+//! @note
+//!  This function should not log anything because it is used
+//!  in the logger implementation.
+int format_tid();
```
gavv: I suggest to convert the signature to the following:

```
bool format_tid(char* buf, size_t bufsz);
```

(similar to format_time)

This way we'll avoid using platform-dependent types (pid_t/int in this case) in the declaration of platform-independent function. The function will gather thread ID and format it as string, and it will not matter what specific types it uses.

src/modules/roc_core/target_bsd/roc_core/format_tid.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <pthread_np.h>
+
+#include "roc_core/format_tid.h"
+
+namespace roc {
+namespace core {
+
+int format_tid() {
+    return pthread_getthreadid_np();
+}
```
gavv: I was mistaken. MacOS and BSD have different functions: it's `pthread_threadid_np` on MacOS (defined in `pthread.h`) and `pthread_getthreadid_np` on BSD (defined in `pthread_np.h`). So we can't use the same target directory for MacOS and BSD...

We don't have BSD support yet. We will add it in #284. Thus, let's remove target_bsd for now. Let's add only target_linux. We already have target_darwin for MacOS.

format_tid from target_linux will use syscall() and format_tid from target_darwin will use pthread_threadid_np.

gavv:
Thanks for the PR!

RohanK99:
I have added the fixes for Mac OSX but using "pthread_getthreadid_np()" on my local machine still causes the build to fail. Any ideas?

src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2019 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <pthread.h>
+#include <stdint.h>
```
RohanK99: should I include `<stdint.h>` here or include `core/stddefs.h`

src/modules/roc_core/target_linux/roc_core/format_tid.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <unistd.h>
+
+#include "roc_core/format_tid.h"
+
+namespace roc {
+namespace core {
+
+bool format_tid(char* buf, size_t bufsz) {
+    pid_t tid = (pid_t)syscall(SYS_gettid);
+    if (snprintf(buf, bufsz, "%d", tid) < 0)
```
RohanK99: for one line if's should I use curly braces or no?

src/modules/roc_core/target_stdio/roc_core/log.cpp
```diff
@@ -77,6 +78,11 @@ void Logger::print(const char* module, LogLevel level, const char* format, ...)
             timestamp[0] = '\0';
         }
 
+        char pid[20] = {};
```
RohanK99: buffer is size 20 because largest 64 bit number is 20 characters long

src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
gavv: We usually use `core/stddefs.h`, but `stdint.h` is also OK here because we're inside target directory and can be sure that `stdint.h` is available.

src/modules/roc_core/target_linux/roc_core/format_tid.cpp
gavv: Yes, we always use curly braces.

src/modules/roc_core/target_stdio/roc_core/log.cpp
gavv: Then it should be 21 - we need one more byte for terminating zero.

src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace roc {
+namespace core {
+
+bool format_tid(char* buf, size_t bufsz) {
+    uint64_t tid = 0;
+    pthread_threadid_np(NULL, &tid);
+    if (snprintf(buf, bufsz, "%llu", tid) < 0)
+        return false;
+    return true;
```
gavv: I'm not sure if uint64_t is guaranteed to be unsigned long long. I suggest to add an explicit cast:

```
snprintf(buf, bufsz, "%llu", (unsigned long long)tid)
```

Also, the snprintf error checking is not quite correct. It should be something like:

```
int ret = snprintf(buf, bufsz, "%llu", tid);
return ret > 0 && (size_t)ret < bufsz;
```

src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <stdio.h>
+
+#include "roc_core/format_tid.h"
+
+namespace roc {
+namespace core {
+
+bool format_tid(char* buf, size_t bufsz) {
+    uint64_t tid = 0;
+    pthread_threadid_np(NULL, &tid);
```
gavv: Let's check pthread_threadid_np return value too.

src/modules/roc_core/target_darwin/roc_core/format_tid.cpp
gavv: The same for linux version.

gavv:
Thanks!

PR #323, opened 2020-01-15
Fixed broken sox download URL (replaced with more generic version)

garethmarlow:


gavv:
Thanks!

gavv:
Also merged into master.

PR #331, opened 2020-01-26
Fix potential deadlock in async_close()

gavv:
In the previois implementation, async_close() was allowed to invoke
handle_close() in place. async_close() is always called under a
mutex. handle_close() acquires the mutex too, resulting in a deadlock.

The problem didn't appear in practice because the circumstances when
handle_close() is called in place are very rare: uv_udp_init() should
fail to make it happen, or double async_close() call should happen.

In the new implementation, instead of calling handle_close() in place,
async_close() returns false instead, which means that there is no need
to wait for handle_close() call.

### DIFF
--- a/src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
+++ b/src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
@@ -111,14 +111,20 @@ bool UdpSenderPort::open() {
     return true;
 }
 
-void UdpSenderPort::async_close() {
+bool UdpSenderPort::async_close() {
     core::Mutex::Lock lock(mutex_);
 
     stopped_ = true;
 
-    if (pending_ == 0) {
-        close_();
+    if (fully_closed_()) {
+        return false;
     }
+
+    if (pending_ == 0) {
+        start_closing_();
+    }
+
+    return true;
 }
 
 void UdpSenderPort::write(const packet::PacketPtr& pp) {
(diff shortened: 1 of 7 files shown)


src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -111,14 +111,24 @@ bool UdpSenderPort::open() {
     return true;
 }
 
-void UdpSenderPort::async_close() {
+bool UdpSenderPort::async_close() {
     core::Mutex::Lock lock(mutex_);
 
     stopped_ = true;
 
+    if (!handle_initialized_ && !write_sem_initialized_) {
```
dshil: NITPICK: we have the following in several places:

```cpp
 if (!handle_initialized_ && !write_sem_initialized_) 
```

Probably we could inline it in: ```was_opened``` ?

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -252,13 +262,10 @@ packet::PacketPtr UdpSenderPort::read_() {
 
 void UdpSenderPort::close_() {
     if (closed_) {
-        return; // handle_closed() was already called
+        return;
```
dshil: NITPICK: let's rearrange the check like we have in UdpReceiverPort:

```
// check if was successfully opened
// check if was closed
// close
```

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
gavv: I've extracted `fully_closed_()`.

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
gavv: Done.

Issue #102, opened 2017-08-06
Switch to cryptographically strong PRNG


gavv:
## Problem

In RTP, the SSRC value and the initial value for seqnums and timestamps should be random.

Here is what [RFC 3550](https://tools.ietf.org/html/rfc3550) says about seqnums:

> The initial value of the sequence number
      SHOULD be random (unpredictable) to make known-plaintext attacks
      on encryption more difficult, even if the source itself does not
      encrypt according to the method in Section 9.1, because the
      packets may flow through a translator that does.  Techniques for
      choosing unpredictable numbers are discussed in [17].

These requirements have security reasons. The RFC recommends using a cryptographically strong PRNG (CSPRNG), but currently our PRNG is uniform but not cryptographically strong.

## Solution

We already use libuv, which also provides CSPRNG: http://docs.libuv.org/en/v1.x/misc.html#c.uv_random

Here is what we should do:

* Rename current PRNG: `roc_core/target_posix/roc_core/random.h|cpp` to `roc_core/target_posix/roc_core/fast_random.h|cpp` and `core::random()` to `core::fast_random()`.

* Add `roc_core/target_libuv/roc_core/secure_random.h|cpp` with `core::secure_random()`. Implement is using synchronous version of `uv_random()` (set `loop`, `req`, and `cb` to NULL).

* Use `secure_random()` instead of `fast_random()` in `audio::Packetizer` and `fec::Writer`.

gavv:
I think we can address this when we'll start working on SRTP support.

gavv:
Should be done after #229.

fusuiyi123:
Hi @gavv i will send a pr for this, thanks.

fusuiyi123:
Q: I don't quite understand the original `random()` usage here: https://github.com/roc-project/roc/blob/master/src/modules/roc_audio/packetizer.cpp#L38, and `int uv_random(uv_loop_t* loop, uv_random_t* req, void* buf, size_t buflen, unsigned int flags, uv_random_cb cb)` returns succeed or not, which I don't quite understand how to use it here. New to this repo and your docs really impressed me, Would appreciate your help.

gavv:
@fusuiyi123 Hi, you're welcome.

> I don't quite understand the original random() usage here

We generate a random number in range [0; 2^32) and assign it to `source_`. (`packet::source_t` is a 32-bit unsigned integer and so `packet::source_t(-1)` is 2^32-1).

[...]

PR #340, opened 2020-03-29
add secure_random and change random to fast_random

fusuiyi123:
### fixes https://github.com/roc-project/roc/issues/102
* how to add test for secure_random?
* is pthread and `void random_init()` still needed in secure_random?
* I don't understand how to add `ifdef` from your comments, but after changed the version to `1.33.0` it compiles successfully.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,15 +99,6 @@ jobs:
         - scripts/travis/docker.sh rocproject/cross-aarch64-linux-android:api28 scripts/travis/android/aarch64-linux-android.sh
         - cat build.status
 
-    - name: "macOS Xcode 7.3"
-      stage: Build
-      os: osx
-      osx_image: xcode7.3
-      script:
-        - scripts/travis/macos/xcode-7.3.sh
-      after_failure:
-        - cat config.log
-
     - name: "macOS Xcode 9.1"
       stage: Build
       os: osx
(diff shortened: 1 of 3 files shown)


gavv:
Hi,

Please target PRs to develop branch, as stated in [CONTRIBUTING.md](https://github.com/roc-project/roc/blob/master/CONTRIBUTING.md). Note that in develop branch random.h and random.cpp have changed.

> how to add test for secure_random?

I think we can go without tests in this case. This function is trivial and yet hard to test.

> is pthread and void random_init() still needed in secure_random?

Blindly copy-pasting code is not a good idea. If you read carefully what random_init() does, you'll see that it initializes a seed for nrand48(), which you don't use here.

> I don't understand how to add ifdef from your comments

What exactly do you have troubles with? If you don't know how to use preprocessor, you should go to the C manual. If you don't know how to gather libuv version, you should go to the libuv manual.

I really appreciate any help (we need it), but if you ask for advice or a piece of code for every minor issue, this help will become pretty meaningless. On the other hand, feel free to ask anything specific to this project which you can't learn from widely available sources. Also, I'd recommend [this read](http://www.catb.org/esr/faqs/smart-questions.html) (as well as other Eric's reads; they're inspiring).

> but after changed the version to 1.33.0 it compiles successfully.

Actually two build jobs failed:

rocproject/cross-aarch64-linux-gnu:gcc-7.4
https://travis-ci.org/github/roc-project/roc/jobs/668487440?utm_medium=notification&utm_source=github_status

rocproject/cross-aarch64-linux-android:api28
https://travis-ci.org/github/roc-project/roc/jobs/668487443?utm_medium=notification&utm_source=github_status

We should investigate why. Travis don't give enough details, so please try to reproduce them locally. See https://roc-project.github.io/roc/docs/development/continuous_integration.html#run-locally

Yeah, and thanks for taking the time on this project!

fusuiyi123:
thanks for your advice! I will make sure to follow the good question practice.

Issue #328, opened 2020-01-21
Improve integration tests for public API


gavv:
Currently we have 3 [integrations tests](https://github.com/roc-project/roc/blob/d61d81ace3fd443d966c21b519e2cc08861f61bb/src/tests/roc_library/test_sender_receiver.cpp#L429) for our [C API](https://roc-streaming.org/toolkit/docs/api.html):

* without FEC
* with Reed-Solomon FEC and without losses
* with Reed-Solomon FEC and with losses

We need to cover more parameter combinations:

- [x] Use single context for sender and receiver, or use separate contexts.
- [x] Use one sender per receiver, or connect multiple senders to receiver (sequential version).
- [ ] Use one sender per receiver, or connect multiple senders to receiver (parallel version).
- [x] Use one receiver, or create multiple receivers sharing one context, and connect a sender to each receiver.
- [x] Use Reed-Solomon or LDPC-Staircase FEC schemes, with and without losses.
- [ ] Enable or disable packet interleaving.
- [ ] Use different resampler profiles.
- [ ] Use different clock sources (external or internal).
- [ ] Test a few different presets of FEC block size, latency, and breakage detection parameters. For example, one low latency preset and one high latency preset.

Before adding more tests, it would be also nice to extract Context, Sender, Receiver, and Proxy from test_sender_receiver.cpp into separate files in src/tests/roc_library (i.e. test_context.h, test_sender.h, and so on).

This work can be split into multiple PRs, one or a few tests per PR.

CleverShovel:
I would like to try write these tests.

gavv:
@CleverShovel Great, you're welcome!

CleverShovel:
@gavv Is this normal that in roc_library directory there are files like test_context.cpp, test_sender.cpp and others and I will add test_context.h and others but they will not related to these cpp?

gavv:
Good catch. Since this issue was created, the file layout in tests was changed a bit. According to the new layout, these files should be named like `roc_library/test_helpers/context.h`, `roc_library/test_helpers/sender.h`, etc. See, for example, tests for roc_pipeline or roc_audio.

CleverShovel:
@gavv Does each item from the list in issue description mean concrete tests or only a parameter to combine?

gavv:
[...]

PR #344, opened 2020-03-30
Add api integration tests

CleverShovel:
Issue #328
Changed RepairPackets in utils.h to 7 because openfec can't initialize matrix for LDPC.

### DIFF
--- a/src/tests/roc_library/test_sender_receiver.cpp
+++ b/src/tests/roc_library/test_sender_receiver.cpp
@@ -8,428 +8,23 @@
 
 #include <CppUTest/TestHarness.h>
 
-#include <stdio.h>
-#include <unistd.h>
+#include "test_helpers/context.h"
+#include "test_helpers/proxy.h"
+#include "test_helpers/receiver.h"
+#include "test_helpers/sender.h"
 
-#include "roc_address/socket_addr.h"
-#include "roc_core/buffer_pool.h"
-#include "roc_core/heap_allocator.h"
-#include "roc_core/log.h"
-#include "roc_core/panic.h"
-#include "roc_core/random.h"
-#include "roc_core/stddefs.h"
-#include "roc_core/thread.h"
-#include "roc_core/time.h"
 #include "roc_fec/codec_map.h"
-#include "roc_netio/event_loop.h"
-#include "roc_packet/packet_pool.h"
-#include "roc_packet/queue.h"
 
-#include "roc/context.h"
-#include "roc/endpoint.h"
 #include "roc/log.h"
-#include "roc/receiver.h"
-#include "roc/sender.h"
 
 namespace roc {
+namespace library {
 
 namespace {
 
-enum {
-    MaxBufSize = 500,
-
-    SampleRate = 44100,
-    NumChans = 2,
-
-    SourcePackets = 10,
-    RepairPackets = 5,
-
-    PacketSamples = 100,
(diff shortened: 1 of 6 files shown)


src/tests/roc_library/test_sender_receiver.cpp
```diff
@@ -11,534 +11,282 @@
...
-    roc_endpoint* input_repair_endp_;
+    sender.start();
+    receiver.run();
+    sender.stop();
+    sender.join();
+}
 
-    netio::UdpReceiverConfig recv_source_config_;
-    netio::UdpReceiverConfig recv_repair_config_;
+TEST(sender_receiver, multiple_senders_one_receiver) {
```
gavv, line 168: This test sequentially runs one sender and then another. It's a good idea to add such a test, but we also need a test that runs two senders in parallel. It would be a more complicated task because we'll have to adjust receiver logic accordingly.

I suggest to rename this test to `multiple_senders_one_receiver_sequential` and then add a new test `multiple_senders_one_receiver_simultaneous`. I think it's better to add the new test in a new PR because this one is too big already. In this PR we can just rename the test.

src/tests/roc_library/test_sender_receiver.cpp
```diff
@@ -11,534 +11,282 @@
...
+    sender_2.start();
+    receiver.run();
+    sender_2.stop();
+    sender_2.join();
+}
 
-TEST_GROUP(sender_receiver) {
-    roc_sender_config sender_conf;
-    roc_receiver_config receiver_conf;
+TEST(sender_receiver, multiple_receivers_and_senders) {
```
gavv, line 197: The same concern with this test. We need a parallel version. Not sure if we need a sequential version of this test, given that we will already have a sequential version of the previous test. Up to you.

src/tests/roc_library/test_sender_receiver.cpp
```diff
@@ -11,534 +11,282 @@
...
-    sender.join();
+    sender_3.start();
+    receiver_3.run();
+    sender_3.stop();
+    sender_3.join();
 }
 
-TEST(sender_receiver, fec_without_losses) {
-    if (!is_fec_supported()) {
+TEST(sender_receiver, ldpc_without_losses) {
```
gavv, line 241: I suggest to group reed-solomon and LDPC tests together, e.g. move `ldpc_without_losses` and `ldpc_with_losses` right after `fec_with_losses` (which should be renamed to `rs8m_with_losses I guess).

src/tests/roc_library/test_sender_receiver.cpp
```diff
@@ -11,534 +11,282 @@
 #include <stdio.h>
 #include <unistd.h>
```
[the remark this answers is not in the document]
CleverShovel: Without <stdio.h> header build fails because of `snprintf` in file `test_helpers/receiver.h`. Should I add this header directly to the receiver.h file or maybe there is some other way to fix it?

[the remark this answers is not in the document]
gavv: Yeah, better move it to receiver.h because each header should be self-sufficient and should not depend on what headers the user included earlier.

gavv:
LGTM!

The test `multiple_senders_one_receiver_sequential` failed once:

Screenshot_2020-04-04_10-55-30

I restarted the build and it passed. Let's merge this and see how it will go. If it fails more, we'll have to investigate why.

PR #345, opened 2020-03-30
add CSPRNG

### LINKED ISSUE #102: Switch to cryptographically strong PRNG

fusuiyi123:
fixes #102

### DIFF
--- a/src/modules/roc_core/target_posix/roc_core/fast_random.h
+++ b/src/modules/roc_core/target_posix/roc_core/fast_random.h
@@ -6,23 +6,23 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-//! @file roc_core/target_posix/roc_core/random.h
+//! @file roc_core/target_posix/roc_core/fast_random.h
 //! @brief Helpers to work with random numbers.
 
-#ifndef ROC_CORE_RANDOM_H_
-#define ROC_CORE_RANDOM_H_
+#ifndef ROC_CORE_FAST_RANDOM_H_
+#define ROC_CORE_FAST_RANDOM_H_
 
 #include "roc_core/stddefs.h"
 
 namespace roc {
 namespace core {
 
-//! Get random integer.
+//! Get a random integer from a non cryptographically secure, but fast PRNG.
 //! Thread-safe.
 //! @returns random value in range [from; to].
-uint32_t random(uint32_t from, uint32_t to);
+uint32_t fast_random(uint32_t from, uint32_t to);
 
 } // namespace core
 } // namespace roc
 
-#endif // ROC_CORE_RANDOM_H_
+#endif // ROC_CORE_FAST_RANDOM_H_
--- a/src/modules/roc_core/target_posix/roc_core/secure_random.cpp
+++ b/src/modules/roc_core/target_posix/roc_core/secure_random.cpp
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2020 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <uv.h>
+
+#include "roc_core/fast_random.h"
+#include "roc_core/log.h"
+#include "roc_core/panic.h"
+#include "roc_core/secure_random.h"
+
+namespace roc {
+namespace core {
+
+// Based on uv_random from libuv and arc4random_uniform() from OpenBSD.
+#if UV_VERSION_MAJOR > 1 || (UV_VERSION_MAJOR == 1 && UV_VERSION_MINOR >= 33)
+bool secure_random(uint32_t from, uint32_t to, uint32_t& result) {
+    roc_panic_if_not(from <= to);
+
+    const uint64_t upper = (uint64_t)to - from + 1;
+    const uint64_t min = -upper % upper;
+    uint64_t val;
+
+    for (;;) {
+        if (int err = uv_random(NULL, NULL, &val, sizeof(val), 0, NULL)) {
+            roc_log(LogError, "secure random : uv_random(): [%s] %s", uv_err_name(err),
+                    uv_strerror(err));
+            return false;
+        }
+        if (val >= min) {
+            break;
+        }
+    }
+
+    const uint32_t ret = from + (uint32_t)(val % upper);
+
+    roc_panic_if_not(ret >= from);
+    roc_panic_if_not(ret <= to);
+
+    result = ret;
+    return true;
+}
+#else  // UV_VERSION_MAJOR < 1 || (UV_VERSION_MAJOR == 1 && UV_VERSION_MINOR < 33)
+bool secure_random(uint32_t from, uint32_t to, uint32_t& result) {
+    result = fast_random(from, to);
+    return true;
+}
+#endif // UV_VERSION_MAJOR > 1 || (UV_VERSION_MAJOR == 1 && UV_VERSION_MINOR >= 33)
+
+} // namespace core
+} // namespace roc
--- a/src/modules/roc_core/target_posix/roc_core/secure_random.h
+++ b/src/modules/roc_core/target_posix/roc_core/secure_random.h
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2020 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_core/target_posix/roc_core/secure_random.h
+//! @brief Helpers to work with CSPRNG.
+
+#ifndef ROC_CORE_SECURE_RANDOM_H_
+#define ROC_CORE_SECURE_RANDOM_H_
+
+#include "roc_core/stddefs.h"
+
+namespace roc {
+namespace core {
+
+//! Get a random integer from a cryptographically secure pseudorandom number
+//! generator(CSPRNG).
+//! @remarks
+//!  Writes a random value in [from; to] to @p result.
+//!  May block indefinitely.
+//!  Falls back to a regular PRNG when using libuv version before 1.33.0.
+//! @returns true on success or false if accessing CSPRNG failed.
+bool secure_random(uint32_t from, uint32_t to, uint32_t& result);
+
+} // namespace core
+} // namespace roc
+
+#endif // ROC_CORE_SECURE_RANDOM_H_
--- a/src/modules/roc_fec/writer.cpp
+++ b/src/modules/roc_fec/writer.cpp
@@ -9,7 +9,7 @@
 #include "roc_fec/writer.h"
 #include "roc_core/log.h"
 #include "roc_core/panic.h"
-#include "roc_core/random.h"
+#include "roc_core/secure_random.h"
 #include "roc_packet/fec_scheme_to_str.h"
 
 namespace roc {
@@ -37,12 +37,17 @@ Writer::Writer(const WriterConfig& config,
     , buffer_pool_(buffer_pool)
     , repair_block_(allocator)
     , first_packet_(true)
-    , cur_sbn_((packet::blknum_t)core::random(0, packet::blknum_t(-1)))
-    , cur_block_repair_sn_((packet::seqnum_t)core::random(0, packet::seqnum_t(-1)))
     , cur_packet_(0)
     , fec_scheme_(fec_scheme)
     , valid_(false)
     , alive_(true) {
+    uint32_t rand_cur_sbn = 0, rand_cur_block_repair_sn = 0;
+    if (!core::secure_random(0, packet::blknum_t(-1), rand_cur_sbn)
+        || !core::secure_random(0, packet::seqnum_t(-1), rand_cur_block_repair_sn)) {
+        return;
+    }
+    cur_sbn_ = (packet::blknum_t)rand_cur_sbn;
+    cur_block_repair_sn_ = (packet::seqnum_t)rand_cur_block_repair_sn;
     if (!resize(config.n_source_packets, config.n_repair_packets)) {
         return;
     }
--- a/src/tests/roc_audio/test_resampler.cpp
+++ b/src/tests/roc_audio/test_resampler.cpp
@@ -18,7 +18,6 @@
 #include "roc_audio/resampler_reader.h"
 #include "roc_core/buffer_pool.h"
 #include "roc_core/heap_allocator.h"
-#include "roc_core/random.h"
 #include "roc_core/scoped_ptr.h"
 #include "roc_core/stddefs.h"
 
--- a/docs/sphinx/internals/code_structure.rst
+++ b/docs/sphinx/internals/code_structure.rst
@@ -107,8 +107,8 @@ Example directory structure employing targets:
     ├── target_posix
     │   └── roc_core
     │       ├── ...
-    │       ├── random.cpp
-    │       └── random.h
+    │       ├── fast_random.cpp
+    │       └── fast_random.h
     ├── target_posixtime
     │   └── roc_core
     │       ├── ...
--- a/src/modules/roc_audio/packetizer.cpp
+++ b/src/modules/roc_audio/packetizer.cpp
@@ -9,7 +9,7 @@
 #include "roc_audio/packetizer.h"
 #include "roc_core/log.h"
 #include "roc_core/panic.h"
-#include "roc_core/random.h"
+#include "roc_core/secure_random.h"
 
 namespace roc {
 namespace audio {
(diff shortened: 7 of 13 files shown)


gavv:
Please ping me when the PR will be ready for review.

fusuiyi123:
Hi @gavv the android build fails with
[build.log](https://github.com/roc-project/roc/files/4413027/build.log)
and from the libuv supporting platform, https://github.com/libuv/libuv/blob/a62f8ced7af51d1c4fd8abceb521bd24f362ab14/SUPPORTED_PLATFORMS.md, this may not be actively supported

gavv:
Could you please run our `rocproject/cross-aarch64-linux-android:api28` docker container and try build libuv manually inside it?

* They have some specific build instructions for android: https://github.com/libuv/libuv/tree/master#android. Probably it will solve the issue.

* They also have cmake build. Theoretically it also may help.

* And they also have gyp + ninja build... https://github.com/libuv/libuv/tree/master#using-ninja

If non of these helps, could you then try to build different libuv versions and find out which is the latest version the builds correctly?

fusuiyi123:
1. `source ./android-configure-arm64 NDK_PATH gyp [API_LEVEL]` worked
```
root@951ba020b0df:/Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0# source ./android-configure-arm64 /tmp/android-ndk-r18/ gyp 28
HOST_OS=linux
HOST_EXE=
HOST_ARCH=x86_64
HOST_TAG=linux-x86_64
HOST_NUM_CPUS=2
BUILD_NUM_CPUS=4
Toolchain installed to /Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/android-toolchain-arm64.
['-Dtarget_arch=arm64', '-DOS=android', '-f', 'make-android', '-I', 'common.gypi', 'test/test.gyp', '--depth=.', '-Goutput_dir=/Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/out', '--generator-output', '/Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/out', '-Dhost_arch=x64', '-Duv_library=static_library']
```
2. `make -C out` fails with
```
root@951ba020b0df:/Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0# make -C out
make: Entering directory `/Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/out'
  CC(target) /Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/out/Debug/obj.target/libuv/src/fs-poll.o
/Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/android-toolchain-arm64/bin/aarch64-linux-android-gcc: line 2: /Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/android-toolchain-arm64/bin/../../../../llvm/prebuilt/linux-x86_64/bin/clang: No such file or directory
make: *** [/Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/out/Debug/obj.target/libuv/src/fs-poll.o] Error 127
make: Leaving directory `/Users/fusuiyi/Desktop/opensource/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.35.0/src/libuv-v1.35.0/out'
```
where `cat android-toolchain-arm64/bin/aarch64-linux-android-gcc` gives `#!/bin/bash
exec `dirname $0`/../../../../llvm/prebuilt/linux-x86_64/bin/clang -target aarch64-linux-android "$@"`.
Still looking at it.

gavv:
> make -C out fails with

Did you run it inside `rocproject/cross-aarch64-linux-android:api28` docker container? The command output looks like it was running on macOS :/

fusuiyi123:
yes, it is in docker container mounted from pwd

fusuiyi123:
libuv 1.33.1 build succeed, project build fails though :/
```
       AR   build/aarch64-linux-android/clang-7.0.2-release/libroc_core.a
645   RANLIB   build/aarch64-linux-android/clang-7.0.2-release/libroc_core.a
646       LD   build/aarch64-linux-android/clang-7.0.2-release/libroc_unstripped.so.0.1
6473rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.33.1/lib/libuv.a(libuv_la-random.o): In function `uv__random':
648/opt/roc/3rdparty/aarch64-linux-android/clang-7.0.2-release/build/libuv-1.33.1/src/libuv-v1.33.1/src/random.c:40: undefined reference to `uv__random_getentropy'
649clang70++: error: linker command failed with exit code 1 (use -v to see invocation)
650scons: *** [build/aarch64-linux-android/clang-7.0.2-release/libroc_unstripped.so.0.1] Error 1
651The command "scripts/travis/docker.sh rocproject/cross-aarch64-linux-android:api28 scripts/travis/android/aarch64-linux-android.sh" exited with 2.
```

gavv:
Thanks, so it seems that the autotools libuv build is just broken on Android. I think this can be fixed by switching to cmake libuv build. I'm now trying to do it.

Useful documentation on cmake & android:

* https://cmake.org/cmake/help/v3.7/manual/cmake-toolchains.7.html#cross-compiling-for-android
* https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_ARCH.html
* https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_ARCH_ABI.html#variable:CMAKE_ANDROID_ARCH_ABI
* https://developer.android.com/ndk/guides/abis
* https://developer.android.com/ndk/guides/standalone_toolchain

gavv:
Done. See c5bb75f0c25a11d7c2dc6eb76c36b7f9fe91c7aa

Now we're using libuv 1.35.0 by default and use cmake for android builds. The issue with undefined references should be fixed, please rebase on fresh develop branch and try again.

fusuiyi123:
wow that's cool 👍 !!

docs/sphinx/internals/glossary.rst
```diff
@@ -83,10 +83,10 @@ RTP
     Synchronization source. Identifies a stream inside a session and should be unique only inside that session.
 
 *seqnum*
-    Sequence number. Identifies a packet inside a stream. All packets in a stream are numbered sequentially usually starting from a random number.
+    Sequence number. Identifies a packet inside a stream. All packets in a stream are numbered sequentially usually starting from a cryptographically strong random number.
```
gavv, line 86: I think we should omit "cryptographically strong" here. First, it's not necessary for understanding the terms (seqnum and timestamp). Second, it's not quite true: we use CSPRNG when possible (libuv > 1.33), but not necessary. It's an important detail. We should document it separately. I think in future we should add a page with security concerns. Third, "cryptographically strong" is a property of random *generator*, not a random *number*.

src/modules/roc_core/target_posix/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "roc_core/fast_random.h"
+#include "roc_core/panic.h"
+#include "roc_core/secure_random.h"
+
+namespace roc {
+namespace core {
+
+// Based on uv_random from libuv and arc4random_uniform() from OpenBSD.
+uint32_t secure_random(uint32_t from, uint32_t to) {
+#if UV_VERSION_MAJOR >= 1 && UV_VERSION_MINOR >= 33
```
gavv, line 21: When using preprocessor, it's preferred to use it for the entire function instead of a code block inside a function. This approach is less intrusive and easier to read and maintain. For example, [see here](https://www.kernel.org/doc/Documentation/process/coding-style.rst).

Also, don't forget about trailing comments on #else and #endif.

So it should become something like:

```
#if CONDITION
uint32_t secure_random(uint32_t from, uint32_t to) {
    ...
}
#else // !CONDITION
uint32_t secure_random(uint32_t from, uint32_t to) {
    ...
}
#endif // CONDITION
```

Use [this file](https://github.com/roc-project/roc/blob/develop/src/modules/roc_core/target_posixtime/roc_core/time.cpp) as an example.

gavv: The check `UV_VERSION_MAJOR >= 1 && UV_VERSION_MINOR >= 33` is actually not quite correct. It will not match `2.0` for example.

src/modules/roc_core/target_posix/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2020 Roc authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <stdlib.h>
```
gavv, line 9: Do we really need stdlib.h here?

src/modules/roc_core/target_posix/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+    const uint64_t upper = (uint64_t)to - from + 1;
+    const uint64_t min = -upper % upper;
+    uint64_t val;
+
+    for (;;) {
+        if (!uv_random(NULL, NULL, &val, sizeof(val), 0, NULL) && val >= min) {
+            break;
+        }
+    }
```
gavv, line 32: So if uv_random() will fail every time (return non-zero), e.g. because /dev/urandom is not available, this loop will hang.

Please always do proper error checking. If uv_random() returns an error, we should log the error message (you can find examples in roc_netio) and return false to the upper level.

The code that calls secure_random() should handle the error as well and fail the session construction. I already mentioned it in the issue, actually: https://github.com/roc-project/roc/issues/102#issuecomment-605116491

src/modules/roc_core/target_posix/roc_core/secure_random.h
```diff
@@ -0,0 +1,27 @@
...
+#define ROC_CORE_SECURE_RANDOM_H_
+
+#include "roc_core/stddefs.h"
+
+namespace roc {
+namespace core {
+
+//! Get secure random integer.
+//! @returns secure random value in range [from; to].
+uint32_t secure_random(uint32_t from, uint32_t to);
```
gavv, line 22: I'd rephrase it from "secure random integer" to something like "random integer from CSPRNG". It's also worth to mention that it can fallback to usual PRNG on older libuv versions. And that it may block for a while.

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -17,8 +17,8 @@
 #include "roc_audio/resampler_map.h"
 #include "roc_audio/resampler_reader.h"
 #include "roc_core/buffer_pool.h"
+#include "roc_core/fast_random.h"
 #include "roc_core/heap_allocator.h"
-#include "roc_core/random.h"
```
gavv, line 21: It seems we don't need this include at all.

src/tests/roc_library/test_sender_receiver.cpp
```diff
@@ -13,10 +13,10 @@
 
 #include "roc_address/socket_addr.h"
 #include "roc_core/buffer_pool.h"
+#include "roc_core/fast_random.h"
 #include "roc_core/heap_allocator.h"
 #include "roc_core/log.h"
 #include "roc_core/panic.h"
-#include "roc_core/random.h"
```
gavv, line 19: It seems we don't need this include at all.

src/modules/roc_core/target_posix/roc_core/fast_random.h
```diff
@@ -20,9 +20,9 @@ namespace core {
 //! Get random integer.
 //! Thread-safe.
 //! @returns random value in range [from; to].
-uint32_t random(uint32_t from, uint32_t to);
+uint32_t fast_random(uint32_t from, uint32_t to);
```
gavv, line 23: It's worth to mention in the comment, that it uses non-choreographically-strong, but faster PRNG.

src/modules/roc_core/target_posix/roc_core/secure_random.cpp
fusuiyi123: cool, to achieve this, what I can think of is change the definition to `int secure_random(uint32_t from, uint32_t to, unsigned &to_assign);` and pass source_, seqnum_, timestamp_ to to_assign and return succeed or not, do you have a better way to indicate the status as well as generate the random number? thanks!

fusuiyi123: or make it as uint32_t* and NULL indicates failure?

src/modules/roc_core/target_posix/roc_core/fast_random.h
```diff
@@ -6,23 +6,23 @@
...
+#ifndef ROC_CORE_FAST_RANDOM_H_
+#define ROC_CORE_FAST_RANDOM_H_
 
 #include "roc_core/stddefs.h"
 
 namespace roc {
 namespace core {
 
-//! Get random integer.
+//! Get non-choreographically-strong but faster random integer.
```
gavv, line 20: "choreographically" was a typo, obviously. Please don't blindly copy-paste what I'm writing :)

Also, "cryptographically strong" is the property of "(pseudo) random integer *generator*" (PRNG), not the random integer itself. So it should be something:

> Get a random integer from a non-cryptographically strong, but fast PRNG.

Also, it seems a more widely used term is "cryptographically secure" instead of "cryptographically strong".

src/modules/roc_core/target_posix/roc_core/secure_random.cpp
gavv:
```
uint32_t* ptr = new uint32_t;
*ptr = ret;
return ptr;
```

This is a very bad design decision on many levels:

1. We return a raw pointer and transfer ownership to the caller. It's easy to forget to deallocate it and get a memory leak. The function signature gives no hint on this (e.g. it could be named "alloc_secure_random"), and you even didn't mention such an important thing in the function comment.

2. The preferred way to *transfer ownership* is to return a smart pointer, which will save the caller from producing memory leaks. There are many cases when raw pointers are still useful, but they should be used with care.

    If you're using `delete` in your code and this code does not belong to some low-level memory management subsystem (e.g. you're writing an allocator), it's a hint that you're probably do something wrong. In most cases you just need a smart pointer.

    In this project, we almost never use raw `new` calls as well. Instead, we almost always allocate memory either from `core::Pool` or from `core::IAllocator` and combine them with smart pointers (`core::ScopedPtr` and `core::SharedPtr`).

3. But anyway, there is absolutely no need to allocate anything here at all. This is an unreasonable burden both to the programmer who uses your function and to performance (allocations are not cheap).

    There is nothing wrong with your first suggestion and it's a common practice in C:

        bool secure_random(uint32_t from, uint32_t to, uint32_t& result);

    (Note that in this project we mostly use `bool` instead of `int` to indicate a failure).

fusuiyi123: thank you for your advice! always learning from you!

src/modules/roc_core/target_posix/roc_core/secure_random.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "roc_core/stddefs.h"
+
+namespace roc {
+namespace core {
+
+//! Get a random integer from a cryptographically secure pseudorandom number
+//! generator(CSPRNG).
+//! @returns secure random value in range [from; to].
+bool secure_random(uint32_t from, uint32_t to, uint32_t& result);
```
gavv, line 23: We should improve the comment a bit:

1. `@returns` section is not quite correct. The function *returns* success status.
2. We should mention that it may block, it's important.
3. We should mention that it may fallback to a regular CSPRNG.

(Actually I already asked for this earlier in comments).

I suggest something like:

```
//! Get a random integer from a cryptographically secure pseudo-random number
//! generator (CSPRNG).
//! @remarks
//!   Writes a random value in range [from; to] to @p result.
//!   May block for a while.
//!   Falls back to a regular PRNG when using libuv versions before 1.33.0.
//! @returns
//!   true on success or false if accessing CSPRNG failed.
```

src/modules/roc_fec/writer.cpp
```diff
@@ -37,12 +37,17 @@ Writer::Writer(const WriterConfig& config,
     , buffer_pool_(buffer_pool)
     , repair_block_(allocator)
     , first_packet_(true)
-    , cur_sbn_((packet::blknum_t)core::random(0, packet::blknum_t(-1)))
-    , cur_block_repair_sn_((packet::seqnum_t)core::random(0, packet::seqnum_t(-1)))
     , cur_packet_(0)
     , fec_scheme_(fec_scheme)
     , valid_(false)
     , alive_(true) {
+    uint32_t rand_cur_sbn_ = 0, rand_cur_block_repair_sn_ = 0;
```
gavv, line 44: We use tailing underscore only for private class fields and methods. Local variables should not have it. See here: https://roc-project.github.io/roc/docs/development/coding_guidelines.html

So `rand_cur_sbn_` -> `rand_cur_sbn` and `rand_cur_block_repair_sn_` -> `rand_cur_block_repair_sn`.

gavv:
The build is [failing](https://travis-ci.org/github/roc-project/roc/jobs/671362819?utm_medium=notification&utm_source=github_status) because `dist.libuv.org` is unstable currently and scons can't download libuv. I've [added](https://github.com/roc-project/vendor/commit/10b73469a455526c95ec4124d41578fb10764ecb) libuv 1.35.0 to our `vendor/` directory, so now scons doesn't have to download it. Rebasing on fresh develop should fix the build.

src/modules/roc_core/target_posix/roc_core/secure_random.h
fusuiyi123: thanks 👍

gavv:
Thanks!

PR #347, opened 2020-04-12
Generate and install pkg-config file

### LINKED ISSUE #271: Generate and install pkg-config file
`scons install` should generate and install a `libroc.pc` file for pkg-config.

The file contents should be something like:

```
prefix=/usr
exec_prefix=${prefix}
libdir=/usr/lib64
includedir=${prefix}/include

Name: libroc
Requires: libuv libunwind
Version: 0.1.3
Description: Real-time audio streaming over the network.
URL: https://roc-project.github.io

Libs: -L${libdir} -lroc -lrt -lpthread -lnsl -ldl -lm
Cflags: -I${includedir}
```

But paths (`/usr`, `/usr/lib64`), version (`0.1.3`), and dependencies (`libuv libunwind`) should be generated dynamically according to the .version file, installation prefix, and enabled dependencies. All this info is already available inside scons.

The file itself should be installed accordingly to the installation prefix and pkg-config default path. On my system it is `/usr/lib64/pkgconfig`.

The generation function should be added somewhere under `site_scons/site_tools/roc`. This is our scons plugin which provided various helpers.

This task is needed for our Go bindings: https://github.com/roc-project/roc-go/issues/4

fusuiyi123:
Fixes #271


fusuiyi123:
since I am new to the SCons I am not sure if I added to the right file and followed the right approach.
I tried testing in ubuntu16.04 docker:
I have
```
root@53b5d196e524:/Users/fusuiyi/Desktop/opensource/roc# scons -Q --enable-werror --build-3rdparty=openfec install
      GET   3rdparty/x86_64-pc-linux-gnu/gcc-5.4.0-release/build/openfec-1.4.2.4
[download] https://github.com/roc-project/openfec/archive/v1.4.2.4.tar.gz
[extract] openfec_v1.4.2.4.tar.gz
[execute] cmake .. -DBUILD_STATIC_LIBS=ON -DDEBUG:STRING=OFF -DCMAKE_FIND_ROOT_PATH='' -DCMAKE_SYSROOT='' -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_C_COMPILER=/usr/bin/gcc-5 -DCMAKE_LINKER=/usr/bin/gcc-5 -DCMAKE_AR=/usr/bin/ar -DCMAKE_RANLIB=/usr/bin/ranlib -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS_RELEASE:STRING='-fPIC -fvisibility=hidden'
[execute] make -j2
[install] 3rdparty/x86_64-pc-linux-gnu/gcc-5.4.0-release/build/openfec-1.4.2.4/include
[install] 3rdparty/x86_64-pc-linux-gnu/gcc-5.4.0-release/build/openfec-1.4.2.4/lib/libopenfec.a
/usr/lib/x86_64-linux-gnu/pkgconfig
  INSTALL   /usr/lib/x86_64-linux-gnu/pkgconfig/libroc.pc
  INSTALL   /usr/include/roc
  INSTALL   /usr/lib/x86_64-linux-gnu/libroc.so.0.1
scons: *** [install] bin/x86_64-pc-linux-gnu/libroc.so.0.1: No such file or directory
```
/usr/lib/x86_64-linux-gnu/pkgconfig/libroc.pc is
```
prefix=/usr
exec_prefix=${prefix}
libdir=/usr/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: libroc
Requires: ragel cpputest gcc posixtime stdio pulseaudio libuv libunwind posix linux sox gengetopt glibc alsa
Version: 0.1.5
Description: Real-time audio streaming over the network.
URL: https://roc-project.github.io

Libs: -L${libdir} -lroc -lrt -lpthread -lnsl -ldl -lm
Cflags: -I${includedir}
```

SConstruct
```diff
@@ -1338,6 +1338,23 @@ env.AlwaysBuild(
             env.PrettyCommand('TIDY', 'src', 'yellow')
         )))
 
+if 'PKG_CONFIG_PATH' in env.Dictionary():
```
gavv, line 1341: Let's replace direct `env` checks with `env.HasArg` calls, like:

```
if env.HasArg('PKG_CONFIG_PATH')
```

It'd allow the user to specify the variables in either of these forms:

```
VAR=foo scons
scons VAR=foo
```

We'll also need `env.OverrideFromArg` call here: https://github.com/roc-project/roc/blob/453631ccbd1cb9ac657877faef2997cee8a78d46/SConstruct#L313-L315

Like:

```
env.OverrideFromArg('PKG_CONFIG', default='pkg-config')
```

SConstruct
```diff
@@ -1338,6 +1338,23 @@ env.AlwaysBuild(
...
+    pkg_config_path = env['PKG_CONFIG_PATH']
+elif 'PKG_CONFIG' in env.Dictionary() or env.Which('pkg-config'):
+    # https://linux.die.net/man/1/pkg-config the default is libdir/pkgconfig
+    pkg_config_path = os.path.join(env['ROC_SYSTEM_LIBDIR'], 'pkgconfig')
+else:
+    pkg_config_path = None
```
gavv, line 1347: Let's extract this pkg-config path detection to `cong.FindPkgConfigPath`. The method should be added to `site_scons/site_tools/roc/config.py` and used in SConstruct like this:

```

conf.FindPkgConfigPath()

env = conf.Finish()
```

Here is an example of such approach: https://github.com/roc-project/roc/blob/453631ccbd1cb9ac657877faef2997cee8a78d46/SConstruct#L478-L481

SConstruct
```diff
@@ -1338,6 +1338,23 @@ env.AlwaysBuild(
             env.PrettyCommand('TIDY', 'src', 'yellow')
         )))
 
+if 'PKG_CONFIG_PATH' in env.Dictionary():
+    pkg_config_path = env['PKG_CONFIG_PATH']
+elif 'PKG_CONFIG' in env.Dictionary() or env.Which('pkg-config'):
+    # https://linux.die.net/man/1/pkg-config the default is libdir/pkgconfig
+    pkg_config_path = os.path.join(env['ROC_SYSTEM_LIBDIR'], 'pkgconfig')
```
gavv, line 1345: If `$PKG_CONFIG` or `pkg-config` tool is found, before falling back to `libdir/pkgconfig`, let's first try to ask the tool itself which path it's using:

```
$ pkg-config --variable pc_path pkg-config
/usr/lib64/pkgconfig:/usr/share/pkgconfig
```

If the command works, we can use the first path it prints. If it doesn't, we can fallback to `libdir/pkgconfig`.

SConstruct
```diff
@@ -1338,6 +1338,23 @@ env.AlwaysBuild(
...
+else:
+    pkg_config_path = None
+if pkg_config_path:
+    pc_target = env.PkgConfig(
```
gavv, line 1349: I suggest to rename it to `env.GeneratePkgConfig`.

SConstruct
```diff
@@ -1338,6 +1338,23 @@ env.AlwaysBuild(
...
+                    build_dir='build',
+                    prefix=GetOption('prefix'),
+                    libdir=env['ROC_SYSTEM_LIBDIR'],
+                    filename='libroc.pc',
+                    version=env['ROC_VERSION'],
+                    dependencies=system_dependencies)
```
gavv, line 1355: Just using `system_dependencies` is not correct for two reasons:

* not all of its elements correspond to real libraries, e.g. `stdio` is not a library
* pkg-config files of the real libraries may have different names, e.g. we can have `pulseaudio` in `system_dependencies`, but we should pass `libpulse` to pkg-config

To solve this, I suggest the following approach.

In SConstruct, we have a lot of check like:

```
if 'libuv' in system_dependencies:

    env.ParsePkgConfig('--cflags --libs libuv')
    ...
```

or:

```
if 'libunwind' in system_dependencies:

    env.ParsePkgConfig('--cflags --libs libunwind')
    ...
```

These checks already have information what pkg-config dependencies do we need. So let's just modify them so that they accumulate the dependency list for us.

I suggest to replace `env.ParsePkgConfig` (see `site_scons/site_tools/roc/config.py`) with `env.AddPkgConfigDependency` which will be used like this:

```
    env.AddPkgConfigDependency('libunwind', '--cflags --libs')
```

`AddPkgConfigDependency` will do the same as `env.ParsePkgConfig` does currently, but if it succeeds, it will also add the dependency name to `env['PKG_CONFIG_DEPS']`.

And then we can just pass `env['PKG_CONFIG_DEPS']` to `env.GeneratePkgConfig`.

SConstruct
```diff
@@ -1338,6 +1338,23 @@ env.AlwaysBuild(
...
+    env.AddDistFile(pkg_config_path, pc_target)
```
gavv, line 1356: Let's move this call to `src/SConscript`. `SConstruct` file is mostly for configuration, and `SConscript` file is for performing real actions.

We should generate pkg-config only if building libroc was not diabled. In SConscript, we already have this section:

```
if not GetOption('disable_lib'):
    ...
```

So let's move this piece to the end of that section.

site_scons/site_tools/roc/commands.py
```diff
@@ -103,6 +103,33 @@ def Ragel(env, source):
...
+libdir=%(libdir)s
+includedir=${prefix}/include
+
+Name: libroc
+Requires: %(dependencies)s
+Version: %(version)s
+Description: Real-time audio streaming over the network.
+URL: https://roc-project.github.io
+
+Libs: -L${libdir} -lroc -lrt -lpthread -lnsl -ldl -lm
```
gavv, line 120: Hard-coding this list:

```
-lrt -lpthread -lnsl -ldl -lm
```

is not portable.

`-lrt` and `-ldl` are not available on all systems, for example.

Instead, let's find the places in SConstruct where they're added, for example:

```
if platform in ['linux']:
    env.AppendUnique(LIBS=['rt', 'dl', 'm'])
```

And when they're added, let's also add them to some array like `env['PKG_CONFIG_LIBS']`. We can add a new method for this, like `env.AddPkgConfigLib('rt')`.

gavv:
```
scons: *** [install] bin/x86_64-pc-linux-gnu/libroc.so.0.1: No such file or directory
```

Could you run scons without `-Q` and post the output?

SConstruct
fusuiyi123: thanks for your reply, I am a little bit unclear here.
I added `AddPkgConfigDependency` in `site_scons/site_tools/roc/config.py`
```python
def AddPkgConfigDependency(context, package, flags):
    if context.env.ParsePkgConfig('%s %s' % (package, flags)):
        context.env['PKG_CONFIG_DEPS'].append(package)
        return True

    return False
```
and in SConstruct replace `env.ParsePkgConfig` like this:
```python
if 'libuv' in system_dependencies:

    conf. AddPkgConfigDependency('libuv', '--cflags --libs')
    ......
    env = conf.Finish()
```
am I understanding correctly?

gavv: Yeah, something like this, but likely we can remove `ParsePkgConfig` entirely and merge its implementation into `AddPkgConfigDependency`.

fusuiyi123:
```
root@211f6f4baa3c:/Users/fusuiyi/Desktop/opensource/roc# scons --enable-werror --build-3rdparty=openfec install
scons: Reading SConscript files ...
Searching CXX executable... (cached) /usr/bin/g++-5
Searching CONFIG_GUESS script... (cached) /usr/share/misc/config.guess
Searching CC executable... (cached) /usr/bin/gcc-5
Searching CXXLD executable... (cached) /usr/bin/g++-5
Searching CCLD executable... (cached) /usr/bin/gcc-5
Searching AR executable... (cached) /usr/bin/ar (/usr/bin/x86_64-linux-gnu-ar)
Searching RANLIB executable... (cached) /usr/bin/ranlib (/usr/bin/x86_64-linux-gnu-ranlib)
Searching STRIP executable... (cached) /usr/bin/strip (/usr/bin/x86_64-linux-gnu-strip)
error: no result
Checking for C library uv... yes
error: no result
Checking for C library unwind... yes
Package atomic_ops was not found in the pkg-config search path.
Perhaps you should add the directory containing `atomic_ops.pc'
to the PKG_CONFIG_PATH environment variable
No package 'atomic_ops' found
error: no result
Checking for C library atomic_ops... no
error: libatomic_ops not found (see 'config.log' for details)
```

gavv:
Well, you need to install libatomic-ops-dev. It's a new dependency introduced recently.

fusuiyi123:
libroc.pc generated:
```
prefix=/usr
exec_prefix=${prefix}
libdir=/usr/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: libroc
Requires: libuv libunwind atomic_ops
Version: 0.1.5
Description: Real-time audio streaming over the network.
URL: https://roc-project.github.io

Libs: -L${libdir} -lrt -ldl -lm -lpthread
Cflags: -I${includedir}
```

fusuiyi123:
```
root@dcc8b5f06b99:/Users/fusuiyi/Desktop/opensource/roc# scons --enable-werror --build-3rdparty=openfec install
scons: Reading SConscript files ...
Searching CXX executable... (cached) /usr/bin/g++-5
Searching CONFIG_GUESS script... (cached) /usr/share/misc/config.guess
Searching CC executable... (cached) /usr/bin/gcc-5
Searching CXXLD executable... (cached) /usr/bin/g++-5
Searching CCLD executable... (cached) /usr/bin/gcc-5
Searching AR executable... (cached) /usr/bin/ar (/usr/bin/x86_64-linux-gnu-ar)
Searching RANLIB executable... (cached) /usr/bin/ranlib (/usr/bin/x86_64-linux-gnu-ranlib)
Searching STRIP executable... (cached) /usr/bin/strip (/usr/bin/x86_64-linux-gnu-strip)
error: no result
Checking for C library uv... yes
error: no result
Checking for C library unwind... yes
error: no result
Checking for C library atomic_ops... yes
error: no result
Checking for C library pulse... yes
error: no result
Checking for C library pulse-simple... yes
error: no result
Checking for C library sox... yes
Checking for executable ragel... (cached) /usr/bin/ragel
Checking for executable gengetopt... (cached) /usr/bin/gengetopt
error: no result
Checking for CXX library CppUTest... yes
python scripts/3rdparty.py 3rdparty/x86_64-pc-linux-gnu/gcc-5.4.0-release vendor '' release openfec-1.4.2.4 '' CXX=/usr/bin/g++-5 CXXLD=/usr/bin/g++-5 CC=/usr/bin/gcc-5 CCLD=/usr/bin/gcc-5 AR=/usr/bin/ar RANLIB=/usr/bin/ranlib
[download] https://github.com/roc-project/openfec/archive/v1.4.2.4.tar.gz
[extract] openfec_v1.4.2.4.tar.gz
[execute] cmake .. -DBUILD_STATIC_LIBS=ON -DDEBUG:STRING=OFF -DCMAKE_FIND_ROOT_PATH='' -DCMAKE_SYSROOT='' -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_C_COMPILER=/usr/bin/gcc-5 -DCMAKE_LINKER=/usr/bin/gcc-5 -DCMAKE_AR=/usr/bin/ar -DCMAKE_RANLIB=/usr/bin/ranlib -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS_RELEASE:STRING='-fPIC -fvisibility=hidden'
[execute] make -j2
[install] 3rdparty/x86_64-pc-linux-gnu/gcc-5.4.0-release/build/openfec-1.4.2.4/include
[install] 3rdparty/x86_64-pc-linux-gnu/gcc-5.4.0-release/build/openfec-1.4.2.4/lib/libopenfec.a
Searching for system library directory... (cached) /usr/lib/x86_64-linux-gnu
Searching PKG_CONFIG_PATH...(cached) error: no result
scons: done reading SConscript files.
scons: Building targets ...
install(/usr/include/roc)
install(/usr/lib/x86_64-linux-gnu/libroc.so.0.1)
scons: *** [install] bin/x86_64-pc-linux-gnu/libroc.so.0.1: No such file or directory
scons: building terminated because of errors.
```

fusuiyi123:
one concern is that I am not clear about the difference between
```python
env.function()
conf.function()
```
since `AddPkgConfigDependency` is added in `config.py` I have to use `conf. AddPkgConfigDependency `

gavv:
> since AddPkgConfigDependency is added in config.py I have to use conf. AddPkgConfigDependency

Yep, your usage is correct.

Configure() returns a new object with methods for configuration. Internally it contains a copy of the to original env.

When you call AddPkgConfigDependency(), it invokes pkg-config, parses its outputs and adds parsed search path, library names, etc, the internal copy of env in Configure object.

When you call `env = conf.Finish()` you assign that internal environment back to the original environment.

gavv:
> scons: *** [install] bin/x86_64-pc-linux-gnu/libroc.so.0.1: No such file or directory

Unfortunately I can't reproduce this on your branch.

fusuiyi123:
ok, so it there any other change needs to be made?

SConstruct
```diff
@@ -320,6 +320,7 @@ env.OverrideFromArg('STRIP', default='strip')
 env.OverrideFromArg('DOXYGEN', default='doxygen')
 env.OverrideFromArg('SPHINX_BUILD', default='sphinx-build')
 env.OverrideFromArg('BREATHE_APIDOC', default='breathe-apidoc')
+env.OverrideFromArg('PKG_CONFIG', default='pkg-config')
```
gavv, line 323: We already has this call for PKG_CONFIG here:

https://github.com/roc-project/roc/blob/05b998cce82fcc4e6ba9de96ab68821075747606/SConstruct#L312-L313

And we should add PKG_CONFIG_PATH.

BTW if PKG_CONFIG_PATH wont have a default, we can just add it to that list at line 312.

SConstruct
```diff
@@ -746,7 +751,7 @@ system_dependencies = all_dependencies - download_dependencies
 if 'libuv' in system_dependencies:
     conf = Configure(env, custom_tests=env.CustomTests)
 
-    env.ParsePkgConfig('--cflags --libs libuv')
+    conf.AddPkgConfigDependency('libuv', '--cflags --libs')
```
gavv, line 754: We also should use AddPkgConfigDependency for google-benchmark.

SConstruct
gavv: We also should add PKG_CONFIG_PATH to this list: https://roc-project.github.io/roc/docs/building/scons_options.html#variables

SConstruct
```diff
@@ -1380,6 +1388,12 @@ env.AlwaysBuild(
             env.PrettyCommand('TIDY', 'src', 'yellow')
         )))
 
+conf = Configure(env, custom_tests=env.CustomTests)
+
+conf.FindPkgConfigPath()
+
+env = conf.Finish()
```
gavv, line 1395: Can we move conf.FindPkgConfigPath() call to the end of this block above?

https://github.com/roc-project/roc/blob/05b998cce82fcc4e6ba9de96ab68821075747606/SConstruct#L1092-L1108

src/SConscript
```diff
@@ -73,6 +73,17 @@ if not GetOption('disable_lib'):
     for s in symlinks:
         env.AddDistFile(env['ROC_SYSTEM_LIBDIR'], s)
 
+    if 'PKG_CONFIG_PATH' in env.Dictionary():
```
gavv, line 76: Let's use env.HasArg here.

SConstruct
```diff
@@ -871,6 +876,7 @@ if 'pulseaudio' in system_dependencies:
                 env.Die("can't find %s" % path)
 
             pulse_env.Append(LIBS=libs)
+            env['PKG_CONFIG_LIBS'] += libs
```
gavv, line 879: 1. The first line appends `libs` to `pulse_env`, but the second line appends `libs` to `env.` `pulse_env` is used to build PulseAudio modules, and these dependencies are specific to them. So the correct code would be:

```
            pulse_env.Append(LIBS=libs)
            pulse_env['PKG_CONFIG_LIBS'] += libs
```

2. Instead of manually duplicating libs to LIBS and PKG_CONFIG_LIBS every time, I suggest to add a method env.AddPkgConfigLibs(libs), which will append given `libs` both to `LIBS` and `PKG_CONFIG_LIBS` (using AppendUnique) and use it everywhere instead of `Append(LIBS=...)` and `AppendUnique(LIBS=...)`.

SConstruct
```diff
@@ -607,6 +608,10 @@ env['ROC_MODULES'] = [
 
 env['ROC_TARGETS'] = []
 
+env['PKG_CONFIG_DEPS'] = []
+
+env['PKG_CONFIG_LIBS'] = []
```
gavv, line 613: Can we hide initialization of these variables inside scons plugin? E.g. automatically init them in AddPkgConfigDependency and AddPkgConfigLibs.

site_scons/site_tools/roc/commands.py
```diff
@@ -103,6 +103,34 @@ def Ragel(env, source):
...
+    src = """prefix=%(prefix)s
+exec_prefix=${prefix}
+libdir=%(libdir)s
+includedir=${prefix}/include
+
+Name: libroc
+Requires: %(dependencies)s
+Version: %(version)s
+Description: Real-time audio streaming over the network.
+URL: https://roc-project.github.io
```
gavv, line 118: Nitpick: I'm trying to keep our scons plugin (the code under site_scons/site_tools/roc) generic, keeping project-specific stuff in SConstruct and SConscript. So let's pass Name, Description, and URL as arguments instead of hard-coding them.

site_scons/site_tools/roc/config.py
```diff
@@ -290,6 +290,54 @@ def FindConfigGuess(context):
...
+    if env.HasArg('PKG_CONFIG'):
+        pkg_config = env['PKG_CONFIG']
+    elif env.Which('pkg-config'):
+        pkg_config = 'pkg-config'
+    else:
+        return False
+
+    pkg_config_paths = env.CommandOutput('%s --variable pc_path %s' % (pkg_config, pkg_config))
+    try:
+        for path in pkg_config_paths.split(':')[0]:
```
gavv, line 311: The trailing `[0]` shouldn't be here.

site_scons/site_tools/roc/config.py
```diff
@@ -290,6 +290,54 @@ def FindConfigGuess(context):
...
+        return True
+
+    if env.HasArg('PKG_CONFIG'):
+        pkg_config = env['PKG_CONFIG']
+    elif env.Which('pkg-config'):
+        pkg_config = 'pkg-config'
+    else:
+        return False
+
+    pkg_config_paths = env.CommandOutput('%s --variable pc_path %s' % (pkg_config, pkg_config))
```
gavv, line 309: 1. We should quote pkg_config variable because it can contain spaces, like:

```
'"%s" --variable pc_path "%s"' % (pkg_config, pkg_config)
```

or better use `quote()` function like we do it in thirdparty.py.

2. I guess the package name, i.e. the second %s, should be always `pkg-config`? And we shouldn't substitute it with `pkg_config` variable, which can be, for example, something like `/usr/local/bin/pkg-config`?

site_scons/site_tools/roc/config.py
```diff
@@ -290,6 +290,54 @@ def FindConfigGuess(context):
...
+
+    if env.HasArg('PKG_CONFIG'):
+        pkg_config = env['PKG_CONFIG']
+    elif env.Which('pkg-config'):
+        pkg_config = 'pkg-config'
+    else:
+        return False
+
+    try:
+        env.ParseConfig('%s %s %s' % (pkg_config, package, flags))
```
gavv, line 335: We also should quote `pkg_config` variable here.

SConstruct
gavv: Also, can we remove the old `env.ParsePkgConfig` method?

https://github.com/roc-project/roc/blob/05b998cce82fcc4e6ba9de96ab68821075747606/site_scons/site_tools/roc/parsers.py#L99-L111

fusuiyi123:
thanks for the review! I pulled the latest ubuntu16.04 and execute
```bash
scons -Q \
      --enable-werror \
      --enable-tests \
      --enable-benchmarks \
      --enable-examples \
      --build-3rdparty=openfec,google-benchmark:1.2.0 \
      install
```
I got `libroc.pc`
```
prefix=/usr
exec_prefix=${prefix}
libdir=/usr/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: roc
Requires: libuv libunwind atomic_ops
Version: 0.1.5
Description: Real-time audio streaming over the network.
URL: https://roc-project.github.io

Libs: -L${libdir} -lrt -ldl -lm -lpthread
Cflags: -I${includedir}
```
But I am still having the error as
```
Searching PKG_CONFIG_PATH...(cached) /usr/lib/x86_64-linux-gnu/pkgconfig
error: no result
error: no result
scons: done reading SConscript files.
scons: Building targets ...
install(/usr/include/roc)
install(/usr/lib/x86_64-linux-gnu/libroc.so.0.1)
scons: *** [install] bin/x86_64-pc-linux-gnu/libroc.so.0.1: No such file or directory
scons: building terminated because of errors.
```

fusuiyi123:
it installed successfully after I build first,
```
root@c552ee1884a6:/Users/fusuiyi/Desktop/opensource/roc# cat /usr/lib/x86_64-linux-gnu/pkgconfig/libroc.pc
prefix=/usr
exec_prefix=${prefix}
libdir=/usr/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: roc
Requires: libuv libunwind atomic_ops
Version: 0.1.5
Description: Real-time audio streaming over the network.
URL: https://roc-project.github.io

Libs: -L${libdir} -lrt -ldl -lm -lpthread
Cflags: -I${includedir}
```

site_scons/site_tools/roc/config.py
```diff
@@ -290,6 +295,67 @@ def FindConfigGuess(context):
...
+                env['PKG_CONFIG_PATH'] = path
+                break
+    except:
+        pass
+    finally:
+        if 'PKG_CONFIG_PATH' not in env.Dictionary():
+            # https://linux.die.net/man/1/pkg-config the default is libdir/pkgconfig
+            env['PKG_CONFIG_PATH'] = os.path.join(env['ROC_SYSTEM_LIBDIR'], 'pkgconfig')
+            if not os.path.isdir(env['PKG_CONFIG_PATH']):
+                os.makedirs(env['PKG_CONFIG_PATH'])
```
gavv, line 328: I think FindPkgConfigPath() should not call makedirs because it's task is to *find* the directory. We already have makedirs call in GeneratePkgConfig, so I guess we can just remove this `if`.

site_scons/site_tools/roc/config.py
```diff
@@ -290,6 +295,67 @@ def FindConfigGuess(context):
...
+        return False
+
+    try:
+        env.ParseConfig('%s %s %s' % (quote(pkg_config), package, flags))
+        env.AppendUnique(PKG_CONFIG_DEPS=[package])
+        return True
+    except:
+        return False
+
+def AddPkgConfigLibs(context, libs):
```
gavv, line 351: We can make this a method of `env` instead of a method of `conf` because AddPkgConfigLibs() does not perform any configuration (e.g. searching for headers, etc) and is usually used when we don't have `conf`.

Also, we can extract GeneratePkgConfig and AddPkgConfigLibs into a new file `pkgconfig.py`.

src/SConscript
```diff
@@ -73,6 +73,20 @@ if not GetOption('disable_lib'):
...
+            build_dir='build',
+            filename='libroc.pc',
+            prefix=GetOption('prefix'),
+            libdir=env['ROC_SYSTEM_LIBDIR'],
+            name='roc',
+            desc='Real-time audio streaming over the network.',
+            url='https://roc-project.github.io',
+            version=env['ROC_VERSION'],
+            dependencies=env.get('PKG_CONFIG_DEPS', []),
+            libs=env.get('PKG_CONFIG_LIBS', []))
```
gavv, line 87: I suggest to hide PKG_CONFIG_DEPS and PKG_CONFIG_LIBS inside GeneratePkgConfig (and remove dependencies and libs args), so that these variables will be an implementation detail.

site_scons/site_tools/roc/config.py
```diff
@@ -290,6 +295,67 @@ def FindConfigGuess(context):
...
+        context.Result(env['PKG_CONFIG_PATH'])
+        return True
+
+    if env.HasArg('PKG_CONFIG'):
+        pkg_config = env['PKG_CONFIG']
+    elif env.Which('pkg-config'):
+        pkg_config = 'pkg-config'
+    else:
+        context.Result('not found')
+        return False
```
gavv, line 313: If `pkg-config` tool wasn't found, we can still fallback to libdir/pkgconfig below instead of failing.

site_scons/site_tools/roc/config.py
```diff
@@ -290,6 +295,67 @@ def FindConfigGuess(context):
...
+    context.Message("Searching PKG_CONFIG_PATH...")
+
+    if env.HasArg('PKG_CONFIG_PATH'):
+        context.Result(env['PKG_CONFIG_PATH'])
+        return True
+
+    if env.HasArg('PKG_CONFIG'):
+        pkg_config = env['PKG_CONFIG']
+    elif env.Which('pkg-config'):
+        pkg_config = 'pkg-config'
```
gavv, line 310: It seems this branch (selecting system-wide `pkg-config` tool from PATH) is valid only if we're *not* cross-compiling. Because if we're, out sysroot likely has its own pkg-config directory.

So to pass `crosscompile` variable to this method:

```
conf.FindPkgConfigPath(crosscompile)
```

and if it's set to True, we should skip this `elif` branch.

Other branches, e.g. `env.HasArg('PKG_CONFIG')` or using `libdir/pkgconfig`, seems to be valid for cross-compiling.

gavv:
> it installed successfully after I build first,

OK, so I can reproduce it in a docker container with ubuntu 16.04. And yes, install works only after successful build. This problem is reproduced without your PR too.

Could you please file a new issue for that? This is a minor bug I think, I'll take a look at it later.

site_scons/site_tools/roc/config.py
fusuiyi123: I also added `crosscompile` check in `AddPkgConfigDependency`, let me know if it is desirable.

site_scons/site_tools/roc/config.py
```diff
@@ -290,6 +295,55 @@ def FindConfigGuess(context):
...
+    return True
+
+def AddPkgConfigDependency(context, crosscompile, package, flags):
+    env = context.env
+    if 'PKG_CONFIG_DEPS' not in env.Dictionary():
+        env['PKG_CONFIG_DEPS'] = []
+    if env.HasArg('PKG_CONFIG'):
+        pkg_config = env['PKG_CONFIG']
+    elif not crosscompile and env.Which('pkg-config'):
+        pkg_config = 'pkg-config'
```
gavv, line 336: > I also added crosscompile check in AddPkgConfigDependency, let me know if it is desirable.

Good question!

After some reading, I think we should improve `pkg-config` tool detection algorithm. I suggest to extract it into another method, FindPkgConfig, and call it from SConstruct before calling FindPkgConfigPath and AddPkgConfigDependency.

I suggest the following algorithm:

* if `PKG_CONFIG` is set explicitly (check using HasArg), use it;
* if `toolchain` is not empty;

  * try to search for `toolchain + "-pkg-config"` in PATH (using Which); if it's found, use it;
  * otherwise, if `PKG_CONFIG_PATH`, or `PKG_CONFIG_LIBDIR`, or `PKG_CONFIG_SYSROOT_DIR` environment variable is set (check using HasArg), try to search for "pkg-config" in PATH; if it's found, use it (rationale: we will use system-wise pkg-config, but it will operate with cross-compilation sysroot, so that's OK);
  * otherwise, fail; `PKG_CONFIG` is not found;

* otherwise, if `toolchain` is empty, try to search for "pkg-config" in PATH; if it's found, use it;

* otherwise, fail; `PKG_CONFIG` is not found.

If `PKG_CONFIG` was not found, AddPkgConfigDependency should always return False. FindPkgConfigPath may still succeed, e.g. if PKG_CONFIG_PATH was set explicitly or if it fall back to default path.

Reference:

* https://autotools.io/pkgconfig/cross-compiling.html
* https://stackoverflow.com/questions/9221236/pkg-config-fails-to-find-package-under-sysroot-directory

fusuiyi123:
the centos build failed with
`fatal error: sox.h: No such file or directory`
after `sudo ln -s /usr/include/sox/sox.h /usr/include/sox.h` it worked.
ref: https://github.com/pytorch/audio/issues/53#issuecomment-425621514

site_scons/site_tools/roc/pkgconfig.py
```diff
@@ -0,0 +1,70 @@
+import os
+
+def FindPkgConfig(env, toolchain):
```
gavv, line 3: Hm, I think this is should be in `config.py` (a method of conf).

site_scons/site_tools/roc/pkgconfig.py
```diff
@@ -0,0 +1,70 @@
...
+    if env.HasArg('PKG_CONFIG'):
+        return True
+
+    if toolchain:
+        if env.Which(toolchain + '-pkg-config'):
+            env['PKG_CONFIG'] = toolchain + '-pkg-config'
+            return True
+        if env.Which('PKG_CONFIG_PATH') \
+            or env.Which('PKG_CONFIG_LIBDIR') \
+            or env.Which('PKG_CONFIG_SYSROOT_DIR'):
```
gavv, line 13: `env.Which(foo)` -> `foo in os.environ`

(earlier I suggested HasArg, but that's incorrect; they should be in environment to be available for pkg-config and can't be scons arguments)

site_scons/site_tools/roc/pkgconfig.py
```diff
@@ -0,0 +1,70 @@
...
+            if env.Which('pkg-config'):
+                env['PKG_CONFIG'] = 'pkg-config'
+                return True
+        return False
+
+    if env.Which('pkg-config'):
```
gavv, line 19: I guess this should be:

```
if not toolchain and env.Which('pkg-config'):
```

site_scons/site_tools/roc/pkgconfig.py
```diff
@@ -0,0 +1,70 @@
+import os
+
+def FindPkgConfig(env, toolchain):
+    if env.HasArg('PKG_CONFIG'):
+        return True
+
+    if toolchain:
```
gavv, line 7: Let's add a comment with this link: https://autotools.io/pkgconfig/cross-compiling.html

SConstruct
```diff
@@ -743,10 +743,12 @@ if 'all' in download_dependencies:
 # dependencies that should be pre-installed on system
 system_dependencies = all_dependencies - download_dependencies
 
+env.FindPkgConfig(toolchain)
```
gavv, line 746: Let's move this to the end of this `conf` block: https://github.com/roc-project/roc/blob/38cd0704108edb73263ef504462c50995e6a32e9/SConstruct#L570

Note that this call should be before this block:

https://github.com/roc-project/roc/blob/38cd0704108edb73263ef504462c50995e6a32e9/SConstruct#L694-L699

otherwise, cloned envs wont have correct PKG_CONFIG var.

gavv:
> the centos build failed with
> fatal error: sox.h: No such file or directory
> after sudo ln -s /usr/include/sox/sox.h /usr/include/sox.h it worked.

I guess that pkg-config stuff is somehow broken in this PR.

I pulled rocproject/travis-centos and here is what I see:

```
[root@05cf94ebe888 sox]# ls -l /usr/include/sox/
total 92
-rw-r--r-- 1 root root 92326 Aug  9  2019 sox.h
[root@05cf94ebe888 sox]# pkg-config --cflags sox
-I/usr/include/sox  
```

So pkg-config should tell us to add /usr/include/sox the search path and just including `<sox.h>` should work (and works in develop).

So adding a symlink should not be needed if we use pkg-config correctly.

site_scons/site_tools/roc/pkgconfig.py
fusuiyi123: ok! I removed PKG_CONFIG_PATH in the override arg then. 
does `foo in os.environ` mean `foo in env.Dictionary()`?

site_scons/site_tools/roc/pkgconfig.py
fusuiyi123: here `toolchain`will be False because there is a `return False` in the end of above if toolchain check?

fusuiyi123:
> pkg-config --cflags sox
> -I/usr/include/sox

yes, I think that's the issue you suggested above!

site_scons/site_tools/roc/pkgconfig.py
fusuiyi123: I guess
```python
def FindPkgConfigPath(context):
    env = context.env

    context.Message("Searching PKG_CONFIG_PATH...")

    if env.HasArg('PKG_CONFIG_PATH'):
        context.Result(env['PKG_CONFIG_PATH'])
        return True
```
also needs to change to `'PKG_CONFIG_PATH' in os.environ` or `'PKG_CONFIG_PATH' in env.Dictionary()`?

site_scons/site_tools/roc/pkgconfig.py
gavv: Ah, I see, OK.

site_scons/site_tools/roc/pkgconfig.py
gavv: I think no. HasArg() checks that the var is either in os.environ or passed to scons as an argument, like:

```
scons ... VAR=value
```

Both ways are ok for us, and both ways we'll get the value of PKG_CONFIG_PATH and use it for installation rule. So here we should use HasArg.

However, in this check:

```
        if 'PKG_CONFIG_PATH' in os.environ \
            or 'PKG_CONFIG_LIBDIR' in os.environ \
            or 'PKG_CONFIG_SYSROOT_DIR' in os.environ:
            if env.Which('pkg-config'):
```

we want to ensure that PKG_CONFIG_PATH (or other vars) is passed to `pkg-config` tool. Environment variables are passed, but scons arguments aren't. So here we should check only for environment variables. If the user declared them, likely pkg-config will behave correctly with the cross-compilation sysroot and likely we can use it without asking the user to specify it explicitly.

Note that this is not a completely strict check, just a reasonable default. I expect that many cross-compilation environments have those vars exported, and when they are, our build script will just work out of the box. However, if our detection will go awry, the user can always specify PKG_CONFIG_PATH and PKG_CONFIG manually.

gavv: TLDR: the current checks look good :)

gavv: But since we use HasArg for PKG_CONFIG_PATH, we should return PKG_CONFIG_PATH to this loop:

```
for var in ['CXX', 'CC', 'AR', 'RANLIB', 'RAGEL', 'GENGETOPT', 'PKG_CONFIG', 'CONFIG_GUESS']:
    env.OverrideFromArg(var)
```

IIRC, we need OverrideFromArg if we want to allow passing a var as an argument. OverrideFromArg checks whether a var was specified as an argument (`scons ... VAR=...`), and if it was, it stores it into `env`.

gavv: The rest is good!

fusuiyi123: thank you! I added `PKG_CONFIG_PATH` back to the override arg list.

gavv:
LGTM!

NB: I've renamed libroc.pc to roc.pc.

Issue #348, opened 2020-04-25
UDP send optimization


gavv:
## Task

When a new outgoing packet is created on sender pipeline, it is passed to `netio::UdpSenderPort::write()`, which adds the packet to the outgoing queue and wakes up libuv thread by raising `write_sem_`. Then, `write_sem_cb_()` is called on libuv thread, fetches the packet from the queue, and asks libuv to send it by calling `uv_udp_send()`.

Such implementation allows to keep `write()` non-blocking even if the socket is not ready for write yet. libuv will wait until the socket becomes ready and then write the packet to it.

However, in most cases the socket *is* actually ready for write and we could reduce the latency a bit by avoiding an extra thread switch.

To do this, we can try to do a non-blocking `sendto()` directly in `write()`. If it succeeds, we're lucky and the packet is passed to kernel immediately without extra thread switch. And if it returns `EAGAIN`, we can fallback to the current implementation (add to queue and raise `write_sem_`).

## Implementation

1. Implement a new function `bool netio::sendto_nb(int fd, const void* buf, size_t bufsz, const address::Address& dst_addr)`. It will be a cross-platform wrapper for non-blocking `sendto()` syscall, for now implemented only for POSIX. It should return true on success. The implementation should be placed into `roc_netio/target_posix/roc_netio/sendto.h` and `roc_netio/target_posix/roc_netio/sendto.cpp`.

2. Add `uv_os_fd_t fd_` field to `netio::UdpSenderPort`. Fill it in `open()` using `uv_fileno()`. We can't use libuv handles from other threads, so we should store the raw socket handle instead.

3. Add the actual optimization to `UdpSenderPort::write()`. Before locking the mutex, check if the pending packet counter is zero. If it is, try to call `sendto_nb()` (use saved `fd_`, the provided packet, and it's destination address). If it succeeded, return; otherwise, fall back to the usual implementation (enqueue a packet and wake up thread).

DColadas:
Hey! I have had an eye at the project and I am interested. May I work on this issue?

gavv:
Great, you're welcome!

gavv:
Merged!

PR #350, opened 2020-04-28
Implement sendto_nb()

DColadas:
Hello,
I have checked this project and found it quite interesting. I spent some time solving the issue #348 (avoiding a thread switch if no more packets are enqueued in the UDP sender).
I am new to contributing in open source projects, so please tell me if I should add, fix or improve something!

### DIFF
--- a/src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
+++ b/src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
@@ -158,16 +158,18 @@ void UdpSenderPort::write(const packet::PacketPtr& pp) {
             return;
         }
 
-        if (rate_limiter_.allow()) {
-            const double nb_ratio = packet_counter_ != 0
-                ? (double)nb_packet_counter_ / packet_counter_
-                : 0.;
-            roc_log(LogDebug, "udp sender: total=%u nb=%u nb_ratio=%.5f",
-                    packet_counter_, nb_packet_counter_, nb_ratio);
-        }
+        if (config_.non_blocking_enabled) {
+            if (rate_limiter_.allow()) {
+                const double nb_ratio = packet_counter_ != 0
+                    ? (double)nb_packet_counter_ / packet_counter_
+                    : 0.;
+                roc_log(LogDebug, "udp sender: total=%u nb=%u nb_ratio=%.5f",
+                        packet_counter_, nb_packet_counter_, nb_ratio);
+            }
 
-        if (try_nonblocking_send_(pp)) {
-            return;
+            if (try_nonblocking_send_(pp)) {
+                return;
+            }
         }
 
         list_.push_back(*pp);
--- a/src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
+++ b/src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
@@ -11,6 +11,7 @@
 #include "roc_core/helpers.h"
 #include "roc_core/log.h"
 #include "roc_core/panic.h"
+#include "roc_netio/sendto.h"
 
 namespace roc {
 namespace netio {
@@ -28,6 +29,7 @@ UdpSenderPort::UdpSenderPort(const UdpSenderConfig& config,
     , pending_(0)
     , stopped_(true)
     , closed_(false)
+    , fd_()
     , packet_counter_(0) {
 }
 
@@ -98,6 +100,12 @@ bool UdpSenderPort::open() {
         return false;
     }
 
+    const int fd_err = uv_fileno((uv_handle_t*)&handle_, &fd_);
+    if (fd_err != 0) {
+        roc_panic("udp sender: uv_fileno(): [%s] %s", uv_err_name(fd_err),
+                  uv_strerror(fd_err));
+    }
+
     roc_log(LogInfo, "udp sender: opened port %s",
             address::socket_addr_to_str(config_.bind_address).c_str());
 
@@ -142,8 +150,16 @@ void UdpSenderPort::write(const packet::PacketPtr& pp) {
             return;
         }
 
+        if (try_nonblocking_send_(pp)) {
+            return;
+        }
+
         list_.push_back(*pp);
         ++pending_;
+
+        if (list_.size() > 1) {
+            return;
+        }
     }
 
     if (int err = uv_async_send(&write_sem_)) {
@@ -250,6 +266,25 @@ packet::PacketPtr UdpSenderPort::read_() {
     }
 
     return pp;
+}
+
+bool UdpSenderPort::try_nonblocking_send_(const packet::PacketPtr& pp) {
+    if (pending_ == 0) {
+        const packet::UDP& udp = *pp->udp();
+        const bool success =
+            sendto_nb(fd_, pp->data().data(), pp->data().size(), udp.dst_addr);
+        if (success) {
+            ++packet_counter_;
+            roc_log(LogTrace,
+                    "udp sender: sent packet non-blocking: num=%u src=%s dst=%s sz=%ld",
+                    packet_counter_,
+                    address::socket_addr_to_str(config_.bind_address).c_str(),
+                    address::socket_addr_to_str(udp.dst_addr).c_str(),
+                    (long)pp->data().size());
+            return true;
+        }
+    }
+    return false;
 }
 
 bool UdpSenderPort::fully_closed_() const {
--- a/src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
+++ b/src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
@@ -16,6 +16,12 @@
 namespace roc {
 namespace netio {
 
+namespace {
+
+const core::nanoseconds_t PacketLogInterval = 20 * core::Second;
+
+} // namespace
+
 UdpSenderPort::UdpSenderPort(const UdpSenderConfig& config,
                              ICloseHandler& close_handler,
                              uv_loop_t& event_loop,
@@ -30,7 +36,9 @@ UdpSenderPort::UdpSenderPort(const UdpSenderConfig& config,
     , stopped_(true)
     , closed_(false)
     , fd_()
-    , packet_counter_(0) {
+    , rate_limiter_(PacketLogInterval)
+    , packet_counter_(0)
+    , nb_packet_counter_(0) {
 }
 
 UdpSenderPort::~UdpSenderPort() {
@@ -148,6 +156,14 @@ void UdpSenderPort::write(const packet::PacketPtr& pp) {
 
         if (stopped_) {
             return;
+        }
+
+        if (rate_limiter_.allow()) {
+            const double nb_ratio = packet_counter_ != 0
+                ? (double)nb_packet_counter_ / packet_counter_
+                : 0.;
+            roc_log(LogDebug, "udp sender: total=%u nb=%u nb_ratio=%.5f",
+                    packet_counter_, nb_packet_counter_, nb_ratio);
         }
 
         if (try_nonblocking_send_(pp)) {
@@ -275,6 +291,7 @@ bool UdpSenderPort::try_nonblocking_send_(const packet::PacketPtr& pp) {
             sendto_nb(fd_, pp->data().data(), pp->data().size(), udp.dst_addr);
         if (success) {
             ++packet_counter_;
+            ++nb_packet_counter_;
             roc_log(LogTrace,
                     "udp sender: sent packet non-blocking: num=%u src=%s dst=%s sz=%ld",
                     packet_counter_,
--- a/src/modules/roc_netio/target_posix/roc_netio/sendto.cpp
+++ b/src/modules/roc_netio/target_posix/roc_netio/sendto.cpp
@@ -30,8 +30,8 @@ bool sendto_nb(int fd,
 #endif
         ) {
             roc_log(LogError, "sendto_nb: sendto: %s", core::errno_to_str().c_str());
-            return false;
         }
+        return false;
     }
     return true;
 }
--- a/src/modules/roc_netio/target_posix/roc_netio/sendto.cpp
+++ b/src/modules/roc_netio/target_posix/roc_netio/sendto.cpp
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2020 Roc authors
+ *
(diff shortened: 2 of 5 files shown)


src/modules/roc_netio/target_posix/roc_netio/sendto.cpp
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2019 Roc authors
```
gavv, line 2: 2020

src/modules/roc_netio/target_posix/roc_netio/sendto.h
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2019 Roc authors
```
gavv, line 2: 2020

src/modules/roc_netio/target_posix/roc_netio/sendto.cpp
```diff
@@ -0,0 +1,30 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_netio/sendto.h"
+#include "roc_core/errno_to_str.h"
+#include "roc_core/log.h"
+#include <sys/socket.h>
+#include <sys/types.h>
```
gavv, line 13: Nitpick: please add an empty line between roc and system include blocks.

gavv:
Since we're applying an optimization, it's important to see how it works. Could we also add a new metric to UdpSenderPort:

```
unsigned nb_packet_counter_;
```

that will count the number of packets successfully sent via non-blocking call?

And then, add a `core::RateLimiter` field and use it to periodically (e.g. once per 5 seconds) report both `nb_packet_counter_` and `packet_counter_`, using `roc_log(LogDebug)`.

With these reports, we'll see how often UdpSenderPort is able to use the optimized version.

Here is an example of such periodic reporting:

https://github.com/roc-project/roc/blob/3bf6c3f0f095280a7fe17aaa33fe8276b65ede37/src/modules/roc_audio/depacketizer.cpp#L72-L79

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -98,6 +99,12 @@ bool UdpSenderPort::open() {
         return false;
     }
 
+    const int fd_err = uv_fileno(reinterpret_cast<uv_handle_t*>(&handle_), &fd_);
+    if (fd_err == UV_EINVAL || fd_err == UV_EBADF) {
```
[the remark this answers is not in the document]
DColadas: I hadn't worked with `libuv` before and thought there was no guarantee of `0` not being an error. I just checked, and [every error is a negative number](http://docs.libuv.org/en/v1.x/errors.html). I'll change it.

gavv:
BTW, is seems that we can apply one more small optimization:

```
    {

        ...

        list_.push_back(*pp);
        ++pending_;

        if (list_.size() > 1) {
            return; // HERE
        }
    }

    if (int err = uv_async_send(&write_sem_)) {
        ...
```

If the list was already non-empty, we can avoid calling uv_async_send() because libuv thread was already notified and it will continue reading packets until the list becomes empty.

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -135,6 +142,30 @@ void UdpSenderPort::write(const packet::PacketPtr& pp) {
         roc_panic("udp sender: unexpected packet w/o data");
     }
 
+    if (pending_ == 0) {
```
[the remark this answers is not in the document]
DColadas: Sure! I thought about the data race, but I'm still not too familiar with how the thread is accessed and followed the instructions you wrote in the issue.

src/modules/roc_netio/target_posix/roc_netio/sendto.cpp
DColadas: Maybe it would be a good idea to add this, the preference for C-style casts... and other formatting decisions to the contributing guidelines (even adding a `clang-format` rule).

DColadas:
I've implemented the comments you wrote and the last optimization (I'll work on the RateLimiter later). As you can see I committed small changes many times. Should I squash them all into a single commit, or will that be done whenever we merge? 
Thanks for your help!

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -98,6 +100,9 @@ bool UdpSenderPort::open() {
         return false;
     }
 
+    const int fd_err = uv_fileno((uv_handle_t*)&handle_, &fd_);
+    roc_panic_if(fd_err != 0);
```
gavv, line 104: Would be nice to panic with meaningful error message with the error and strerror, like:

```
roc_panic("udp sender: uv_fileno(): [%s] %s", ...)
```

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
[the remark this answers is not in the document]
gavv: Yeah, I didn't think about it when I was writing the issue.

src/modules/roc_netio/target_posix/roc_netio/sendto.cpp
gavv: Agree.Not sure about clang-format rule though, e.g. sometimes we have to use reinterpret_cast to suppress warnings.

Code style is (partially) documented here: https://roc-project.github.io/roc/docs/development/coding_guidelines.html

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -252,6 +265,33 @@ packet::PacketPtr UdpSenderPort::read_() {
...
+            ++packet_counter_;
+            return true;
+        }
+        roc_log(LogTrace,
+                "udp sender: couldn't send packet non-blocking; falling back to "
+                "write_sem_: num=%u src=%s dst=%s sz=%ld",
+                packet_counter_ + 1,
+                address::socket_addr_to_str(config_.bind_address).c_str(),
+                address::socket_addr_to_str(udp.dst_addr).c_str(),
+                (long)pp->data().size());
```
gavv, line 290: Nitpick: I think it would be easier to debug things if every packet had only one log message. Currently it can be 1 or 3 messages per packet.

I suggest to log in try_nonblocking_send_() only if the non-blocking write succeeded. This way, either try_nonblocking_send_() or write_sem_cb_() will log one message.

gavv:
LGTM! Only two nitpicks above. Yes, please squash commits. The PR is small, so I think it can be a single commit. Do you plan to work on RateLimiter and stuff within this PR?

DColadas:
> LGTM! Only two nitpicks above. Yes, please squash commits. The PR is small, so I think it can be a single commit. Do you plan to work on RateLimiter and stuff within this PR?
I can add it using this same PR. I'll get it done these days.

DColadas:
> Since we're applying an optimization, it's important to see how it works. Could we also add a new metric to UdpSenderPort:
> 
> ```
> unsigned nb_packet_counter_;
> ```
> 
> that will count the number of packets successfully sent via non-blocking call?
> 
> And then, add a `core::RateLimiter` field and use it to periodically (e.g. once per 5 seconds) report both `nb_packet_counter_` and `packet_counter_`, using `roc_log(LogDebug)`.
> 
> With these reports, we'll see how often UdpSenderPort is able to use the optimized version.
> 
> Here is an example of such periodic reporting:
> 
> https://github.com/roc-project/roc/blob/3bf6c3f0f095280a7fe17aaa33fe8276b65ede37/src/modules/roc_audio/depacketizer.cpp#L72-L79

I just uploaded a version that logs the total number of sent packets, packets sent not blocking, and the ratio (not_blocking/total) every 5 seconds. Please feel free to check if it's what you were looking for!

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -143,6 +151,13 @@ void UdpSenderPort::write(const packet::PacketPtr& pp) {
         roc_panic("udp sender: unexpected packet w/o data");
     }
 
+    if (rate_limiter_.allow()) {
+        const double nb_ratio =
+            nb_packet_counter_ != 0 ? (double)nb_packet_counter_ / packet_counter_ : 0.;
+        roc_log(LogDebug, "udp sender: total=%u non-blocking=%u nb_ratio=%.5f",
+                packet_counter_, nb_packet_counter_, nb_ratio);
+    }
```
gavv, line 159: We should move it under the lock, otherwise we'll get a data race on packet_counter_.

Regarding threads in roc_netio.

UdpSenderPort::write() is called from pipeline thread, which performs sound processing and sound I/O. It case of sender, it gets sound from audio file or device, processes samples, formats packets, and passes them to UdpSenderPort.

UdpSenderPort::write_cb_() and other callbacks are called from libuv thread, which performs network I/O.

uv_async_send(), called from write() is a thread-safe way to wake up libuv from another thread. It schedules write_cb_() invocation on libuv thread. Under the hood, uv_async_send() uses eventfd or similar thing.

gavv:
> I just uploaded a version that logs the total number of sent packets, packets sent not blocking, and the ratio (not_blocking/total) every 5 seconds. Please feel free to check if it's what you were looking for!

Yep, that's what I was talking about.

gavv:
Could you please rebase on fresh develop? It was recently force-pushed to deliver a hot fix to master.

DColadas:
Damn! Looks like I messed up the rebase. Sorry about that, I'm not used to this kind of workflow. I'll fix it later!

DColadas:
Pretty sure everything is ready to merge now! Thanks for the patience, again. At the very least I am learning quite a lot.

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -11,10 +11,17 @@
...
 #include "roc_core/log.h"
 #include "roc_core/panic.h"
+#include "roc_netio/sendto.h"
 
 namespace roc {
 namespace netio {
 
+namespace {
+
+const core::nanoseconds_t PacketLogInterval = 5 * core::Second;
```
gavv, line 21: Just tested, the optimization works as expected. Though 5 seconds seems to be too frequent, could you please change it to 20 seconds?

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -142,8 +158,24 @@ void UdpSenderPort::write(const packet::PacketPtr& pp) {
             return;
         }
 
+        if (rate_limiter_.allow()) {
+            const double nb_ratio = nb_packet_counter_ != 0
+                ? (double)nb_packet_counter_ / packet_counter_
+                : 0.;
```
gavv, line 164: If we want to avoid division by zero, we should check `packet_counter_ != 0` instead of `nb_packet_counter_ != 0`.

src/modules/roc_netio/target_posix/roc_netio/sendto.h
```diff
@@ -0,0 +1,39 @@
...
+//! @file roc_netio/target_posix/roc_netio/sendto.h
+//! @brief Wrapper for non-blocking sendto() syscall.
+
+#ifndef ROC_NETIO_SENDTO_H_
+#define ROC_NETIO_SENDTO_H_
+
+#include "roc_address/socket_addr.h"
+#include "roc_core/iallocator.h"
+#include "roc_core/list_node.h"
+#include "roc_core/refcnt.h"
```
gavv, line 18: The last three includes are not needed here. What we need is `roc_core/stddefs.h` for `size_t`.

src/modules/roc_netio/target_libuv/roc_netio/udp_sender_port.cpp
```diff
@@ -142,8 +158,24 @@ void UdpSenderPort::write(const packet::PacketPtr& pp) {
...
+            roc_log(LogDebug, "udp sender: total=%u non-blocking=%u nb_ratio=%.5f",
```
gavv, line 165: Nitpick: `non-blocking` should be `non_blocking` or just `nb` (like a variable name). Just for consistency with the rest logging.

DColadas:
Giving it a second thought, I might have written a bug: every packet not successfully sent non-blocking should get enqueued to send later, right?
https://github.com/roc-project/roc/blob/dc6acfa92b66587e9b07684249502d25671c04db/src/modules/roc_netio/target_posix/roc_netio/sendto.cpp#L24-L36
Right now, if `EAGAIN` or `EWOULDBLOCK` happen, the packet will not be sent at all (`false` is only returned when other errors happen).

gavv:
Oh, really. `return false` should be in the outer `if`, not in the inner one.

gavv:
Another problem is that after this change, netio unit tests (`test_udp.cpp`) don't cover the slow (blocking) path anymore, since all non-blocking writes are likely to succeed on localhost.

To solve this, we can add a field to UdpSenderConfig that enables non-blocking writes feature (set to true by default), and add a separate test that disables it and check UDP still works.

DColadas:
I added the suggested changes. The test is copy and pasted from an existing one, but disabling the non-blocking writes. I guess it is enough, instead of having to try with different amounts of receivers.

Out of curiosity, is it preferable to squash the commits just before merging, instead of every time I push anything to this pull request? I suppose code reviews are easier if I keep the commits separate for now, instead of amending existing ones.

gavv:
Sorry for delay. LGTM!

> I guess it is enough, instead of having to try with different amounts of receivers.

Agree.

> Out of curiosity, is it preferable to squash the commits just before merging, instead of every time I push anything to this pull request? I suppose code reviews are easier if I keep the commits separate for now, instead of amending existing ones.

GitHub allows to see what was changed in the force-push and what was changed since last review. It's possible to mark file as viewed and it will be automatically unmarked if it was changed after a (force) push. So artificial commits for review are not necessary anymore.

If you're working on a large change-set, it's better to group changed into logical commits from start and constantly amend / rebase / force-push during review.

If you're working on a small PR though, it doesn't matter, since GitHub also has "squash & merge" button, and it's easy to squash all commits into one when merging.

Issue #362, opened 2020-05-23
Add core::Timer implementation based on timerfd


gavv:
### Intro

core::Timer is a simple class that allows to wait for deadline expiration in one thread and to update the deadline from other thread(s), concurrently.

https://github.com/roc-project/roc/blob/develop/src/modules/roc_core/timer.h
https://github.com/roc-project/roc/blob/develop/src/modules/roc_core/timer.cpp

Currently we have a generic implementation based on a seqlock and a semaphore with timed-wait operation. It works well, but has a disadvantage: changing deadline causes thread wake up. The waiting thread has to wake up just to see that the deadline is changed and to go to sleep with the new deadline.

It is possible to avoid this unnecessary wake up by implementing core::Timer using timerfd. This implementation will be Linux-specific, of course.

core::Timer is used in ctl::TaskQueue. This optimizations will allow to avoid unnecessary wake ups of the control thread when a control task is scheduled or cancelled.

### Steps

1. Add new target directory `target_notimers`. See #246 for details on adding a new target directory. Enable it in SConstruct for non-Linux platforms.

2. Move current core::Timer implementation into `roc_core/target_notimers`.

3. Add a new core::Timer implementation based on timerfd and place it into `roc_core/target_linux`.

4. Ensure that all tests are passing. Also check roc_ctl benchmarks.

5. Not necessary, but would be nice to add a (meaningful) benchmark for core::Timer and compare results for semaphore- and timerfd-based implementations. Example benchmarks: [1](https://github.com/roc-project/roc/blob/develop/src/tests/roc_core/bench_random.cpp), [2](https://github.com/roc-project/roc/blob/develop/src/tests/roc_core/bench_mpsc_queue.cpp).

### Requirements

These requirements are already met in semaphore-based core::Timer, and we should preserve them in the new implementation:

* try_set_deadline() should be thread-safe. On concurrent calls, all calls except one are allowed to fail, and at least one should guaranteedly succeed.

* try_set_deadline() should be lock-free, except the timerfd syscall, which probably can't be called "lock-free", but hopefully is something close to it. try_set_deadline() should not block the caller and should work well when called concurrently.

* try_set_deadline() should try to avoid the timerfd syscall if wait_deadline() is not currently waiting.

moon9:
Hello,

I want to help solve this issue.

gavv:
Great, you're welcome.

moon9:
Make pull request with changes [#394 ](url)

[...]

PR #394, opened 2020-06-13
Add core::Timer implementation based on timerfd

moon9:
For roc_core/target_linux implement timer based on timerfd.
Old timer implementation moved to roc_core/target_notimers.
Target added for non-linux platform.

/link #362

Implementation based on timerfd have little overhead by "struct timespec -> nanoseconds_t -> struct timespec" convertion.

BM_Random_Fast - +7%
BM_Random_Secure - +2%

% Time increased in relation to target_notimers.


gavv:
Thanks for the PR! I'll review it in upcoming days.

src/modules/roc_core/target_linux/roc_core/timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include "roc_core/timer.h"
+
+namespace roc {
+namespace core {
+
+#if defined(CLOCK_MONOTONIC)
+#define CLOCK_TYPE CLOCK_MONOTONIC
+#else
+#define CLOCK_TYPE CLOCK_REALTIME
+#endif
```
gavv, line 23: Are there really any platforms where timerfd is available, but CLOCK_MONOTONIC is not?

src/modules/roc_core/target_linux/roc_core/timer.h
```diff
@@ -0,0 +1,60 @@
...
+class Timer : public NonCopyable<> {
+public:
+    Timer();
+
+    virtual ~Timer();
+
+    //! Set timer deadline.
+    //! Can be called concurrently, but only one concurrent call will succeed.
+    //! Returns false if the call failed because of another concurrent call.
+    //! Is lock-free if Semaphore::post() is so (which is true of modern plarforms).
```
gavv, line 35: The note on Semaphore::post() in not relevant here.

src/modules/roc_core/target_linux/roc_core/timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+    if (!deadline_.try_store(new_deadline)) {
+        return false;
+    }
+
+    struct itimerspec new_value = convert_deadline();
+
+    int res = timerfd_settime(timerfd_, TFD_TIMER_ABSTIME, &new_value, NULL);
+    if (res == -1) {
+        roc_panic("timer: timerfd_settime(timerfd_, TFD_TIMER_ABSTIME, "
+                  "&new_value, NULL): %s",
```
gavv, line 47: Nitpick: we usually omit non-interesting function arguments in log messages, so it can be:

```
timer: timerfd_settime(TFD_TIMER_ABSTIME): %s
```

or even just:

```
timer: timerfd_settime(): %s
```

gavv: If we really can't rely that CLOCK_MONOTINIC is always available and should be able to work with CLOCK_REALTIME, then according to the man page we should be ready to get ECANCELED here: https://man7.org/linux/man-pages/man2/timerfd_settime.2.html#NOTES

src/modules/roc_core/target_linux/roc_core/timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+Timer::Timer()
+    : timerfd_(timerfd_create(CLOCK_TYPE, 0))
+    , deadline_(0) {
+    if (timerfd_ == -1) {
+        roc_panic("timer: timerfd_create(CLOCK_TYPE, 0): %s", errno_to_str().c_str());
+    }
+}
+
+Timer::~Timer() {
+    close(timerfd_);
```
gavv, line 34: Nitpick: we should check for errors here.

src/modules/roc_core/target_linux/roc_core/timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+    uint64_t ticks = 0;
+    ssize_t readed = read(timerfd_, &ticks, sizeof(uint64_t));
+    if (readed == -1 || ticks == 0) {
+        roc_panic("timer: read(timerfd_, &ticks, sizeof(uint64_t)): %s",
+                  errno_to_str().c_str());
+    }
+}
+
+struct itimerspec Timer::convert_deadline() {
+    const nanoseconds_t deadline = deadline_.wait_load();
```
gavv, line 70: Since we call wait_load() here, try_set_deadline() is not wait-free any more. We should get rid of this call.

src/modules/roc_core/target_linux/roc_core/timer.h
```diff
@@ -0,0 +1,60 @@
...
+    //! Should be called from a single thread.
+    //! Assumes that wait_deadline() calls are serialized.
+    //! Deadline may be changed concurrently from other threads.
+    void wait_deadline();
+
+private:
+    struct itimerspec convert_deadline();
+    struct itimerspec infinity_deadline();
+    struct itimerspec immediately_deadline();
+    struct itimerspec finity_deadline();
```
gavv, line 51: Private methods should end with underscore. However I'd suggest to make them free functions (private to .cpp file, in anonymous namespace) instead of methods.

src/modules/roc_core/target_linux/roc_core/timer.h
```diff
@@ -0,0 +1,60 @@
...
+
+namespace roc {
+namespace core {
+
+//! Thread-safe timer.
+class Timer : public NonCopyable<> {
+public:
+    Timer();
+
+    virtual ~Timer();
```
gavv, line 30: No need for virtual destructor here.

src/modules/roc_core/target_linux/roc_core/timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+void Timer::wait_deadline() {
+    const nanoseconds_t deadline = deadline_.wait_load();
+
+    if (deadline >= 0 && deadline <= timestamp()) {
+        return;
+    }
+
+    uint64_t ticks = 0;
+    ssize_t readed = read(timerfd_, &ticks, sizeof(uint64_t));
```
gavv, line 62: We should handle EINTR here.

If we have to support CLOCK_REALTIME, we should also handle zero return value from read().

gavv:
```
BM_Random_Fast - +7%
BM_Random_Secure - +2%
```

Why are you providing results for these benchmarks? They are unrelated to core::Timer. We don't have benchmarks for core::Timer, though it would be nice to add them, as mentioned in the issue.

src/modules/roc_core/target_linux/roc_core/timer.cpp
moon9: Get reference from time.cpp. If need, I can remove this and set only monotonic.

gavv:
Could you please rebase on fresh develop?

moon9:
> 
> 
> Could you please rebase on fresh develop?

Of course I'll do it.

src/modules/roc_core/target_linux/roc_core/timer.h
moon9: ok

src/modules/roc_core/target_linux/roc_core/timer.cpp
moon9: ECANCELED can get from read if TFD_TIMER_CANCEL_ON_SET flag is set. TFD_TIMER_CANCEL_ON_SET is off now.

src/modules/roc_core/target_linux/roc_core/timer.cpp
moon9: ok

src/modules/roc_core/target_linux/roc_core/timer.cpp
moon9: ok

src/modules/roc_core/target_linux/roc_core/timer.h
moon9: ok

src/modules/roc_core/target_linux/roc_core/timer.h
moon9: ok

src/modules/roc_core/target_linux/roc_core/timer.cpp
moon9: Added EINTR case.

gavv:
Please don't use merge, use rebase + force push instead. Or you can you achieve the same by reset to fresh develop + cherry-pick your commits + force push.

See here: https://roc-streaming.org/toolkit/docs/development/version_control.html

Issue #363, opened 2020-05-24
Add command-line option to enable tracing in tests


gavv:
Our tests has `-v` option, which enables debug logging (LogDebug level). Example:

```
bin/x86_64-pc-linux-gnu/roc-test-core -v
```

It would be very useful to have another option to enable even more verbose trace logging (LogTrace). Ideally, by specifying `-vv` (like in our tools), but a separate option is okay too.

The tricky part here is that the command-line is parsed by CppUTest and we should find a way to register our own options or implement some workaround.

The parsing happens in `test_main.cpp`.

pvimal816:
Anyone working on this issue?

gavv:
@pvimal816 AFAIK it's free.

> My approach is to pass '-v' two time when we want more verbosity level. Then we just count number of time this option passes in test_main.cpp itself and accordingly set logLevel.

Well, sound good, if:

* CppUTest can handle this
* the parsing code will be portable (e.g. getopt function is not)
* we will correctly handle common cases like mixing with other options, e.g. `-vv` or `-vX` where X is another CppUTest option

pvimal816:
I have checked that following command which gets accepted by CppUTest:
`bin/x86_64-pc-linux-gnu/roc-test-core -v -v`

But CppUTest generates error on following command:
`bin/x86_64-pc-linux-gnu/roc-test-core -vv`

So we can use first command to implement our feature.
But here can I assume that accessing `argv` variable in `test_main.cpp` does not make code non-portable?

gavv:
Well if `-vv` does not work, which is the usual way to specify an option multiple times, then I suggest to use some other letter, e.g. `-t` (t for trace).

> Can I assume that accessing argv variable in test_main.cpp does not make code non-portable?

Yeah, argc/argv should be available everywhere. (Unlike getopt function which is POSIX).

pvimal816:
Hi gavv,

 Sorry to interrupt you. But can you please check this Travis CI error and guide me a little.  --- https://travis-ci.org/github/roc-streaming/roc-toolkit/jobs/703784921

gavv:
@pvimal816 Yep, I saw your messages on IRC, but I can't reply you when you're offline. You just wont receive the reply, IRC doesn't support offline messages.

The problem is specific to brew on old macos versions, I'm already looking into it. It's not specific to your PR.

[...]

PR #400, opened 2020-06-30
Added support for -vv option for tests to provide more verbose logs.

pvimal816:
#363

### DIFF
--- a/src/tests/test_main.cpp
+++ b/src/tests/test_main.cpp
@@ -23,7 +23,7 @@ int main(int argc, const char** argv) {
 
     core::CrashHandler crash_handler;
 
-    /* Check wether "-t" option is set.
+    /* Check whether "-t" option is set.
      * If yes then change it to "-v" and remember this in "more_verbose" variable.
      *
      * Reason of changing "-t" to "-v" is that we also want to instruct
--- a/src/tests/test_main.cpp
+++ b/src/tests/test_main.cpp
@@ -14,6 +14,7 @@
 #include "roc_core/exit.h"
 #include "roc_core/heap_allocator.h"
 #include "roc_core/log.h"
+#include "roc_core/stddefs.h"
 
 using namespace roc;
 
@@ -22,9 +23,28 @@ int main(int argc, const char** argv) {
 
     core::CrashHandler crash_handler;
 
+    /* Check wether "-t" option is set.
+     * If yes then change it to "-v" and remember this in "more_verbose" variable.
+     *
+     * Reason of changing "-t" to "-v" is that we also want to instruct
+     * CppUTest to give more verbose output
+     */
+
+    bool more_verbose = false;
+
+    for (int i = 0; i < argc; i++) {
+        if (strcmp(argv[i], "-t") == 0) {
+            argv[i] = "-v";
+            more_verbose = true;
+            break;
+        }
+    }
+
     CommandLineArguments args(argc, argv);
 
-    if (args.parse(NULL) && args.isVerbose()) {
+    if (more_verbose) {
+        core::Logger::instance().set_level(LogTrace);
+    } else if (args.parse(NULL) && args.isVerbose()) {
         core::Logger::instance().set_level(LogDebug);
     } else {
         core::Logger::instance().set_level(LogNone);
--- a/docs/sphinx/building/developer_cookbook.rst
+++ b/docs/sphinx/building/developer_cookbook.rst
@@ -160,8 +160,14 @@ Run a single test:
 .. code::
 
    $ ./bin/x86_64-pc-linux-gnu/roc-test-pipeline -v -g receiver_source -n one_session_long_run
+   
+Enable trace logging:
 
-Run behcnmarks for the module manually:
+.. code::
+
+   $ ./bin/x86_64-pc-linux-gnu/roc-test-core -t
+
+Run benchmarks for the module manually:
 
 .. code::
 


pvimal816:
Travis build for mac-os fails for some reason I can't comprehend. In perticular it fails to download some file. I checked for that URL which indeed works well.

gavv:
Could you please rebase your changes on develop instead of merging and force-push the result to this PR? Or you can hard-reset to fresh develop and then cherry-pick your commits.

pvimal816:
> Could you please rebase your changes on develop instead of merging and force-push the result to this PR? Or you can hard-reset to fresh develop and then cherry-pick your commits.

OK, I will do that.

pvimal816:
after resetting to fresh develop this

src/tests/test_main.cpp
```diff
@@ -22,9 +23,31 @@ int main(int argc, const char** argv) {
...
+    /*
+     * Reason of changing "-vv" to "-v" is that we also want to instruct
+     * CppUTest to give more verbose output
+     */
+
+    bool moreVerbose = false;
+    const char** argv2 = new const char*[(unsigned int)argc];
+    for (int i = 0; i < argc; i++) {
+        if (strcmp(argv[i], "-vv") == 0) {
+            argv2[i] = "-v";
```
gavv, line 39: I don't think this is a good approach. `-vv` is a special case of the generic feature of mixing flags; if we support `-vv`, I'd expect that we also support any mixed combination, so that these calls would be equivalent:

```
./roc-test-core -vvc
./roc-test-core -cvv
./roc-test-core -vcv
./roc-test-core -vc -v
./roc-test-core -cv -v
...
```

(and this example mentions only `-c` option, while there are other options too)

This would require a much more complicated option parsing than just replacing one string with another.

On the other hand, CppUTest itself doesn't support this feature, so e.g. `-vc` and `-cv` aren't accepted, and only `-c -v` or `-v -c` are accepted.

So instead of providing incomplete (and thus surprising for the user) support of mixing flags, maybe just provide a different flag without mixing support?

E.g. we can add a new flag `-t` for our feature and expect that it will be always provided as a separate entry in argv.

pvimal816: Ok. I will implement seperate flag `-t`. 

Is there any other suggestion regrading implementation?

gavv: Nope. Another bad thing with current implementation is using bare new/delete instead of RAII (e.g. core::Array), but the new implementation will be different, I guess, so I didn't mention it.

pvimal816: Let me discuss my new approach briefly.
I will assume that `-t` flag will be provided seperately for enabling trace. Now because `CPPUTEST` does not recognise that option but we still want to instruct it to enable verbose log we have to pass `-v` flag to it. So I will just change `-t` to `-v`. 

But this might create a problem of duplication of option when there is already a `-v` flag provided by user.

pvimal816: Let me discuss my new approach briefly.
I will assume that `-t` flag will be provided seperately for enabling trace. Now because `CPPUTEST` does not recognise that option but we still want to instruct it to enable verbose log we have to pass `-v` flag to it. So I will just change `-t` to `-v`. 

But this might create a problem of duplication of option when there is already a `-v` flag provided by user. I tried to provide duplicate flags to CPPUTEST in our original version of codes and indeed it works fine as of now. But in future they might change the behaviour to handle this scenario. So please suggest whether I should implement it this way or make sure that `CPPUTEST` will never get duplicate option because of this feature.

gavv: I think we can leave duplicate `-v` flags, this is an expected behavior and it's unlikely that they'll break it. So just replacing `-t` with `-v` should be fine.

gavv:
Macos build broke again tonight (brew update), I've already pushed a fix to develop. You'll have to rebase to fix travis build.

src/tests/test_main.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include <CppUTest/CommandLineArguments.h>
 #include <CppUTest/CommandLineTestRunner.h>
+#include <string.h>
```
gavv, line 11: Nitpick: please replace this with `roc_core/stddefs.h`.

src/tests/test_main.cpp
```diff
@@ -22,9 +23,27 @@ int main(int argc, const char** argv) {
...
     core::CrashHandler crash_handler;
 
+    /* Check wether "-t" option is set.
+     * If yes then change it to "-v" and remember this in "moreVerbose" variable.
+     *
+     * Reason of changing "-t" to "-v" is that we also want to instruct
+     * CppUTest to give more verbose output
+     */
+
+    bool moreVerbose = false;
```
gavv, line 33: Nitpick: we use snake_case, so it should be `more_verbose`.

src/tests/test_main.cpp
```diff
@@ -22,9 +23,27 @@ int main(int argc, const char** argv) {
...
+
+    for (int i = 0; i < argc; i++) {
+        if (strcmp(argv[i], "-t") == 0) {
+            argv[i] = "-v";
+            moreVerbose = true;
```
gavv, line 38: Nitpick: we can add `break` here.

src/tests/test_main.cpp
```diff
@@ -22,9 +23,28 @@ int main(int argc, const char** argv) {
 
     core::CrashHandler crash_handler;
 
+    /* Check wether "-t" option is set.
```
gavv, line 26: A typo: should be "whether"

gavv:
LGTM! Sorry, forgot to mention it before, could you please also update developer cookbook?
https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/building/developer_cookbook.rst#running-tests-manually

Just add an example to to run tests with tracing enabled.

pvimal816:
Surely. I will update it.

pvimal816:
@gavv Is this fine? [developer cookbook](https://github.com/pvimal816/roc-toolkit/blob/d014ccade3db7a5e1ea5e23ec0a14b674ba6bca3/docs/sphinx/building/developer_cookbook.rst)

gavv:
Yeah, but I'd move this example above the benchmarks examples, to group tests examples together. Also maybe call it like "enable trace logging" or something, since just "trace" may be not precise enought.

gavv:
https://travis-ci.org/github/roc-streaming/roc-toolkit/builds/706737935

gavv:
Thanks!

PR #401, opened 2020-07-05
Timer for linux based on timerfd

moon9:
For roc_core/target_linux implement timer based on timerfd.
Old timer implementation moved to roc_core/target_notimers.
Target added for non-linux platform.

/link #362


gavv:
Thanks. BTW it wasn't necessary to reopen PR, you can just force-push to the same branch.

src/modules/roc_core/target_linux/roc_core/timer.cpp
```diff
@@ -0,0 +1,143 @@
...
+        roc_panic("timer: close(): %s", errno_to_str().c_str());
+    }
+}
+
+bool Timer::try_set_deadline(nanoseconds_t new_deadline) {
+    if (!deadline_.try_store(new_deadline)) {
+        return false;
+    }
+
+    sem_.wait();
```
gavv, line 86: Could you explain what are we trying to achieve using this sem_.wait() call?

moon9: Is synchronization try_set_deadline and wait_deadline, because it have complexity. While one thread execute wait_deadline, other execute try_set_deadline - significant deadline can be drop. It was in tests before. Semaphore guard significant part of code, where concurrent threads can corrupt deadline.

gavv: But do you really understand the requirement of making try_set_deadline() lock-free (except the syscall)? This issue requires some basic understanding of this stuff.

I see that you're using sem_ to create a critical section, but why did you end up with this solution? The current implementation of try_set_deadline() may be blocked by other try_set_deadline() or wait_deadline() calls, which is exactly what we want to avoid for the reasons I mentioned earlier.

moon9: If you want lock-free, this functions must look like:

try_set_deadline
{
syscall_set(new_deadline);
}

wait_deadline
{
syscall_wait();
}

Other ideas as check on waiting and other is complexity and come as critical section. As I say, without it we loose significant deadlines and catch deadlock on wait_deadline.

We can't do so many actions as atomic operation.

gavv: I feel that lock-free implementation is possible, since there is no big difference from cond-based Timer implementation, which is lock-free. However I don't have the full algorithm in mind so far. I'll look into it a bit later, after finishing more urgent issues of 0.2 release. Ping me if you'll have any ideas.

gavv:
Closing the PR for now.

Issue #396, opened 2020-06-28
Run tests in Android simulator in CI


gavv:
Currently we have Android build in Travis, but it doesn't run unit tests. It is defined [here](https://github.com/roc-streaming/roc-toolkit/blob/61a31b5887829630d0d82259587ad3a1bc14cd48/.travis.yml#L93-L107).

It would be nice to run tests on Travis using Android simulator.

We already have `env-android` docker image that provides Android SDK and emulator, and a helper script to manage simulator devices:

* [documentation](https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/development/continuous_integration.rst#android-environment)
* [dockerfile](https://github.com/roc-streaming/dockerfiles/tree/master/images/env-android)
* [dockerhub page](https://hub.docker.com/r/rocstreaming/env-android)
* [helper script](https://github.com/roc-streaming/dockerfiles/blob/master/images/env-android/device.sh)

This image is used in our java bindings to run java tests in simulator, which was implemented here: https://github.com/roc-streaming/roc-java/pull/21.

Now we want to employ a similar approach for our C++ tests. In java bindings we're using gradle, and to run C++ tests we can use adb as mentioned here: https://github.com/roc-streaming/roc-java/pull/21#issuecomment-643603332

Travis scripts are located here: https://github.com/roc-streaming/roc-toolkit/tree/develop/scripts/travis

An example of running tests manually from travis scripts can be [found here](https://github.com/roc-streaming/roc-toolkit/tree/develop/scripts/travis/linux-arm). Those scripts also run tests in an emulated environment, but they use qemu instead of android simulator.

Related: #390.

MatteoArella:
I can easily work on this

gavv:
Great!

gavv:
Done, except failing tests on i686: #424.

gavv:
Just caught an error on x86_64:

```
--- installing system image android-29;default;x86_64

Warning: An error occurred while preparing SDK package Intel x86 Atom_64 System Image: Connection reset.

The command "scripts/travis/android.sh API=29 NDK_VERSION=21.1.6352462 ABI=x86_64 scripts/travis/android/linux-android.sh" exited with 1.
```

https://travis-ci.org/github/gavv/roc-toolkit/jobs/729580567

@MatteoArella any ideas?

MatteoArella:
> Just caught an error on x86_64:
> 
> ```
[...]

Issue #390, opened 2020-06-01
Add more Android versions to CI


gavv:
Currently our travis build uses Android NDK 21 with API level 28, which corresponds to Android 9.

It would be nice to cover other Android versions as well:

* Android 7
* Android 8 (maybe; if the travis job wont take too much time)
* Android 10

We should determine what NDK versions and API levels corresponds to these Android versions.

Then we should add missing NDK versions to cross-linux-android dockerfile: https://github.com/roc-streaming/dockerfiles/tree/master/cross-linux-android (just add a line to images.csv).

Then we should add travis scripts for corresponding NDK and API versions to https://github.com/roc-streaming/roc-toolkit/tree/develop/scripts/travis/android and call them from .travis.yml.

Finally, we should update docs:

* https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/portability/supported_platforms.rst (update the list of tested versions)
* https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/portability/cross_compiling.rst (mention that we have different NDK versions defined by docker tags plus the "latest" tag for the latest tested NDK)
* https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/building/user_cookbook.rst (explicitly specify NDK version in Android recipes)
* https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/development/continuous_integration.rst (add new images to the table)

Bwoltz:
- Android Level 10 - API level 29
- Android Level 8.1.0 - API level 27
- Android Level 8.0.0 - API level 26
- Android Level 7.1/7.1.1 - API level 25
- Android Level 7.0 - API level 24
[<https://source.android.com/setup/start/build-numbers>](https://source.android.com/setup/start/build-numbers)
[https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels>](https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels)>

I'm not familiar with Android, but do you need to use older versions of NDK? Can't you use the latest version for all your builds and set the minimum API level (24) accordingly?

Sorry if this is a dumb question, just trying to learn. Thanks!

gavv:
Hi!

Yes, using a single NDK is perfectly OK if it supports all necessary API levels.

It seems that NDK 21 supports all necessary API levels so we can use it for all of them:

```
[...]

Issue #388, opened 2020-05-26
Employ build hooks in toolchain docker images


gavv:
We've recently merged a mechanism, based on [docker build hooks](https://docs.docker.com/docker-hub/builds/advanced/#build-hook-examples),  for building multiple variants of a single dockerfile and pushing them to multiple docker tags.

For details, see here:

* [documentation](https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/development/continuous_integration.rst#docker-hub-build-hooks)
* [discussion](https://github.com/roc-streaming/dockerfiles/pull/4)

It's currently used in [`env-android`](https://github.com/roc-streaming/dockerfiles/tree/master/images/env-android) and [`toolchain-linux-android`](https://github.com/roc-streaming/dockerfiles/tree/master/images/toolchain-linux-android) images.

We should use this new mechanism for other cross-compilation images:

* `toolchain-arm-bcm2708hardfp-linux-gnueabi`
* `toolchain-arm-linux-gnueabihf`
* `toolchain-aarch64-linux-gnu`

Instead of duplicating dockerfiles for each tag and changing hard-coded parameters, we should use a single dockerfile for each image, parameterized using ARG mechanism, plus images.csv file which will be used by hooks to populate ARGs.

Dockerfiles are located here: https://github.com/roc-streaming/dockerfiles/tree/master/images

Documentation can be found here: https://roc-streaming.org/toolkit/docs/development/continuous_integration.html

jsomwaru:
I'm interested in the project. Do you think I can take this issue. I'm a little rusty at C++ so I think this would be a good start for me while i get up to speed with the project.

gavv:
@jsomwaru Sure, you're welcome.

jsomwaru:
Would you like this done for other images as well instead of just the ones mentioned above? The Ubuntu images could also be managed the same way.

gavv:
Dockerfiles for different ubuntu tags have different package sets, so I think using a single dockerfile wouldn't be beneficial here: it will anyway have conditions for each supported tag.

gavv:
Thanks!

PR #402, opened 2020-07-07
Add env-android documentation

MatteoArella:
This PR contains CI docs about ``env-android`` docker image.

### DIFF
--- a/docs/sphinx/development/continuous_integration.rst
+++ b/docs/sphinx/development/continuous_integration.rst
@@ -62,6 +62,8 @@ Android cross-compilation
 Image                                      APIs        ABIs                                Compilers
 ========================================== =========== =================================== =============
 rocstreaming/toolchain-linux-android:ndk21 21-29       armeabi-v7a, arm64-v8a, x86, x86_64 clang-9.0.8
+rocstreaming/env-android:jdk8              21-29       armeabi-v7a, arm64-v8a, x86, x86_64 clang-9.0.8
+rocstreaming/env-android:jdk11             21-29       armeabi-v7a, arm64-v8a, x86, x86_64 clang-9.0.8
 ========================================== =========== =================================== =============
 
 Run locally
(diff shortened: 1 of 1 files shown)


docs/sphinx/development/continuous_integration.rst
```diff
@@ -61,6 +61,8 @@ Android cross-compilation
 Image                                      APIs        ABIs                                Compilers
 ========================================== =========== =================================== =============
 rocstreaming/toolchain-linux-android:ndk21 21-29       armeabi-v7a, arm64-v8a, x86, x86_64 clang-9.0.8
+rocstreaming/env-android:jdk8              21-29       armeabi-v7a, arm64-v8a, x86, x86_64 clang-9.0.8
+rocstreaming/env-android:jdk11             21-29       armeabi-v7a, arm64-v8a, x86, x86_64 clang-9.0.8
```
gavv, line 65: I suggest to extract these two into a separate table "Full Android environment", to distinguish them from toolchain-linux-android. And we can rename "Linux cross-compilation" and "Android cross-compilation" tables to "Linux toolchains" and "Android toolchains" to make the difference more clear.

Also, it seems that listing APIs, ABIs, and compilers for env-android doesn't make sense because it may install any needed version? If so, I'd suggest to remove those columns. Instead, we can add a column with JDK version.

docs/sphinx/development/continuous_integration.rst
```diff
@@ -61,6 +61,8 @@ Android cross-compilation
...
 ========================================== =========== =================================== =============
 
 Run locally
```
gavv, line 68: Could you please also rename "Run locally" section to "Run builds locally"?

docs/sphinx/development/continuous_integration.rst
```diff
@@ -90,3 +92,140 @@ Explanation:
...
+* API
+* BUILD_TOOLS_VERSION
+* NDK_VERSION
+* CMAKE_VERSION
+
+Example:
+
+.. code::
+
+    $ docker run -t --rm -v "${PWD}:${PWD}" -w "${PWD}" --env API=28 \
```
gavv, line 122: Since people will copy-paste this example, maybe also add `-v android-sdk:/sdk` to it?

docs/sphinx/development/continuous_integration.rst
MatteoArella: Ok :+1:

docs/sphinx/development/continuous_integration.rst
MatteoArella: Yes sure

MatteoArella:
I've also forgot to document dockerhub hooks. I'll add that in next commit!

gavv:
Great. BTW see also my message on IRC.

gavv:
Thanks! The web-site is built from master, so the new docs are not available there yet, however we can browse them via github. I've added links to the new docs to related issues: #396, #390, #388.

PR #404, opened 2020-07-11
add Android APIs 24, 26 and 29

### LINKED ISSUE #390: Add more Android versions to CI

MatteoArella:
This PR would close #390.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,10 +113,22 @@ jobs:
       before_install:
         - docker pull rocstreaming/env-android
       script:
-        - scripts/travis/android.sh ANDROID_API=28 ANDROID_NDK_VERSION=21.1.6352462 scripts/travis/android/aarch64-linux-android28.sh
-        - scripts/travis/android.sh ANDROID_API=28 ANDROID_NDK_VERSION=21.1.6352462 scripts/travis/android/armv7a-linux-androideabi28.sh
-        - scripts/travis/android.sh ANDROID_API=28 ANDROID_NDK_VERSION=21.1.6352462 scripts/travis/android/x86_64-linux-android28.sh
-        - scripts/travis/android.sh ANDROID_API=28 ANDROID_NDK_VERSION=21.1.6352462 scripts/travis/android/i686-linux-android28.sh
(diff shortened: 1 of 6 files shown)


gavv:
Android job now takes 42 minutes, and it'll become even longer after we'll add tests. I suggest to remove API 26 (Android 8), so that we'll test three versions: two latest widely used versions (currently 9 and 10), and the oldest supported version (currently 7).

I hope this will reduce the time below 30 minutes. However after adding tests it may become too long again. In this case we may try to split the job or speed it up by caching downloaded content between builds.

gavv:
> or speed it up by caching downloaded content between builds

I've created a new issue for that: #405.

gavv:
Related: #406.

gavv:
LGTM! Oh, and thanks for fixing the variables prefix.

Issue #375, opened 2020-05-24
Problem with UndefinedBehaviorSanitizer


gavv:
How to reproduce:

```
$ scons -Q --build-3rdparty=openfec,pulseaudio,sox,cpputest,google-benchmark \
  --compiler=clang --enable-debug --enable-werror \
  --enable-examples --enable-tests --enable-benchmarks \
  --sanitizers=all --enable-pulseaudio-modules \
  test
```

Output:

```
........
     TEST   roc_library
!......src/library/src/sender.cpp:181:46: runtime error: load of value 4294967295, which is not a valid value for type 'roc_interface'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/library/src/sender.cpp:181:46 in 
src/library/src/config_helpers.cpp:213:18: runtime error: load of value 4294967295, which is not a valid value for type 'roc_interface'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/library/src/config_helpers.cpp:213:18 in 
src/library/src/sender.cpp:77:46: runtime error: load of value 4294967295, which is not a valid value for type 'roc_interface'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/library/src/sender.cpp:77:46 in 
src/library/src/sender.cpp:109:46: runtime error: load of value 4294967295, which is not a valid value for type 'roc_interface'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/library/src/sender.cpp:109:46 in 
src/library/src/sender.cpp:142:46: runtime error: load of value 4294967295, which is not a valid value for type 'roc_interface'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/library/src/sender.cpp:142:46 in 
.....src/library/src/endpoint.cpp:64:42: runtime error: load of value 4294967295, which is not a valid value for type 'roc_protocol'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/library/src/endpoint.cpp:64:42 in 
src/library/src/config_helpers.cpp:231:18: runtime error: load of value 4294967295, which is not a valid value for type 'roc_protocol'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/library/src/config_helpers.cpp:231:18 in 
...src/library/src/receiver.cpp:115:46: runtime error: load of value 4294967295, which is not a valid value for type 'roc_interface'
[...]

PR #412, opened 2020-07-15
Update scons script to stop tests in case of runtime error

pvimal816:
Pass -fno-sanitize-recover flag if santitizers are enabled.
Blacklist some functions for sanitizers which are causing false-positives.


SConstruct
```diff
@@ -1335,8 +1335,8 @@ if sanitizers:
         env.Die("sanitizers are not supported for compiler '%s'" % compiler)
 
     for name in sanitizers:
-        flags = ['-fsanitize=%s' % name]
-
+        flags = ['-fsanitize=%s' % name, '-fno-sanitize-recover=%s' % name, '-fsanitize-blacklist=blacklist.txt']
```
gavv, line 1338: Let's specify sanitizer blacklist option only once instead of repeating it.

BTW, is repeating sanitize-recover option with different values okay, or should we pass it once with a list?

blacklist.txt
```diff
@@ -0,0 +1,9 @@
+fun:*roc_sender_connect*
```
gavv, line 1: Can we isolate UB in a few helper conversion/validation functions, similar to proto_from_user, and blacklist only them? I'd prefer to avoid disabling sanitizers for the whole API function.

Also, is it possible to disable only specific sanitizers? 

Also, let's move this file to src/ and rename to something like sanitizer_blacklist.txt

gavv:
#375

gavv:
https://travis-ci.org/github/roc-streaming/roc-toolkit/builds/708303180

SConstruct
pvimal816: > Let's specify the sanitizer blacklist option only once instead of repeating it.
Yes, it was by mistake. I fixed it.

> BTW, is repeating sanitize-recover option with different values okay, or should we pass it once with a list?
From the documentation, it seems good to provide all options in a list. So I fixed it.

blacklist.txt
pvimal816: > Can we isolate UB in a few helper conversion/validation functions, similar to proto_from_user, and blacklist only them? I'd prefer to avoid disabling sanitizers for the whole API function.

No. After the test runner passes invalid value for enum argument even accessing enum variable in caller causes runtime error.

> Also, is it possible to disable only specific sanitizers?

Yes, It's possible and I have updated it.

> Also, let's move this file to src/ and rename to something like sanitizer_blacklist.txt

Done.

SConstruct
```diff
@@ -1333,10 +1333,14 @@ sanitizers = env.ParseList(GetOption('sanitizers'), supported_sanitizers)
 if sanitizers:
     if not compiler in ['gcc', 'clang']:
         env.Die("sanitizers are not supported for compiler '%s'" % compiler)
+    
+    if compiler=='clang':
+        env.AppendUnique(CXXFLAGS=['-fsanitize-blacklist=src/sanitizer_blacklist.txt'])
```
gavv, line 1338: It would be also nice to make `src/sanitizer_blacklist.txt` the dependency of all source or object files, so that they are rebuilt when this file changes.

Here is an example of making a file the dependency of .so file: https://github.com/roc-streaming/roc-toolkit/blob/develop/src/SConscript#L56

One way to do it is to iterate all sources and/or objects arrays in SConscript and explicitly call [Depends()](https://www.scons.org/doc/0.96.91/HTML/scons-user/x995.html) for each file. I'm not sure whether we may call it for sources, probably it will work properly only for objects. In this case we'll probably need to find all places when sources are passed to builders like SharedLibrary() and Program() directly, and instead explicitly construct Object() for each source file, call Depends() for each object, and then pass objects array instead of sources array to builders.

Another way is to write a scanner that will produce the needed dependency: https://scons.org/doc/1.2.0/HTML/scons-user/c3755.html. This solution should be much more compact but I didn't try it before. I suggest to try it first.

SConstruct
```diff
@@ -1333,10 +1333,14 @@ sanitizers = env.ParseList(GetOption('sanitizers'), supported_sanitizers)
...
+
+    flags = "-fno-sanitize-recover=%s" % (", ".join(str(x) for x in sanitizers))
```
gavv, line 1340: Hm, I guess the space in `", "` should be removed?

blacklist.txt
gavv: > No. After the test runner passes invalid value for enum argument even accessing enum variable in caller causes runtime error.

I see. I checked it out and found a workaround: we can change signature of, for example, `interface_from_user()` from `roc_interface in` to `const roc_interface& in`. In this case `roc_sender_*()` functions that invoke `interface_from_user()` may be removed from blacklist. Same should work for `proto_from_user()`.

SConstruct
pvimal816: Maybe we have to use first one.

Cause there seems a bug with `Scanner` factory function.
[See this](https://stackoverflow.com/q/62946778/8272206)

gavv: Interesting, did you see the recent edit to the answer?

pvimal816: Yes I just checked it out. Now it is fine.

pvimal816: I have one doubt.

When I use following code to specify dependencies in `SConstruct` it works as expected(after update to blacklist file cmdline.cpp gets recompiled).
`env.Depends('#build/x86_64-pc-linux-gnu/clang-6.0.0-debug/tools/roc_recv/cmdline.o', '#src/sanitizer_blacklist.txt')`

But When I use the following codes instead to specify that all objects depend on blacklist file it does not work as expected. In particular, after updating the blacklist file `cmdline.o` does not get rebuilt. Instead, it says everything is up to date.

`
dlist = env.GlobRecursive(".", "*.o")
env.Depends(env.File(dlist), '#src/sanitizer_blacklist.txt')
`

I have temporarily disabled implicit cache.

gavv: Are you sure dlist contains expected object files? IIRC GlobRecursive searches in source directory, not in build directory.

In general, just searching for all .o files feels like a bad idea. For example, corresponding . o file may not be created on filesystem yet if this is the first run after clean. Thus, the dependency on blacklist will be added only at second scons invocation. And changing dependencies may cause unwanted rebuilds of object files. Also, object file naming is platform specific.

So instead of searching .o files, I'd either collect scons targets returned by Object() builder, or use scanner if possible.

pvimal816: Yes, that really is not a good way. I was just trying to get familiar with scons.

gavv: Just checked: yes, GlobRecursive searches in source dir: https://github.com/roc-streaming/roc-toolkit/blob/develop/site_scons/site_tools/roc/fs.py#L20 (because of srcnode())

pvimal816: But I printed `dlist` and output contained `build/x86_64-pc-linux-gnu/clang-6.0.0-release/tools/roc_send/cmdline.o` among many other object files.

gavv: No idea :) Maybe try to iterate over dlist and call File() and Depends() for each entry individually. Also you can check that the File() object really refers to the correct file by getting it's abspath property. Sometimes the relative path looks good, but it's relative to a wrong base. You can also try `scons ... --debug=explain`

gavv:
It seems macos builds are broken on develop again: https://travis-ci.org/github/roc-streaming/roc-toolkit/builds/708683310

Likely scons version changed in brew. Don't have time for a closer look right now.

pvimal816:
@gavv It seems that `gcc` does not support `fsanitize-blacklist` option. But it supports `attribute` for functions. So can I use that. I tested it both for `gcc` and `clang` and it works fine.

gavv:
Sounds good. We can add them to `src/modules/roc_core/target_gcc/roc_core/attributes.h`.

gavv:

pvimal816:
> Sounds good. We can add them to `src/modules/roc_core/target_gcc/roc_core/attributes.h`.

Sorry. I didn't got your point. Why we need seperate file to declare attribute of function. I am talking about attribute like one mentioned [here](https://stackoverflow.com/a/27982389/8272206)

Maybe this will not work on other platforms. Let see.

gavv:
That file declares compiler-specific attribute definitions. We should hide the attribute behind a macro to allow future portability across compilers beyond gcc and clang. In such cases, we mostly use [target directories](https://roc-streaming.org/toolkit/docs/internals/code_structure.html#targets) instead of ifdefs.

pvimal816:
> That file declares compiler-specific attribute definitions. We should hide the attribute behind a macro to allow future portability across compilers beyond gcc and clang. In such cases, we mostly use [target directories](https://roc-streaming.org/toolkit/docs/internals/code_structure.html#targets) instead of ifdefs.

Actually `gcc` does not support `fsanitize-blacklist` option and `clang` has little different syntax for function attribute(that we want) than that of `gcc`. So is it okay if I set the environment variable from the `SConstruct` which I can check in `attribute.h` to make a decision?

pvimal816:
GCC does not support the `fsanitize-blacklist` option but both `clang` and `GCC` supports attributes to suppress sanitize for a particular function. So to be consistent I discarded blocklist approach completely.

gavv:
Your current approach looks good.

SConstruct
```diff
@@ -1332,10 +1332,15 @@ sanitizers = env.ParseList(GetOption('sanitizers'), supported_sanitizers)
 if sanitizers:
     if not compiler in ['gcc', 'clang']:
         env.Die("sanitizers are not supported for compiler '%s'" % compiler)
+    
+    if compiler=='clang':
+        env.Append(CPPDEFINES=['CLANG'])
+
+    if compiler == 'gcc':
+        env.Append(CPPDEFINES=['GCC'])
```
gavv, line 1340: No need to define this. We can use predefined macros:

```
#if defined(__clang__)
...
#elif defined(__GNUC__)
...
#endif // __GNUC__

File: src/modules/roc_core/target_gcc/roc_core/attributes.h [hunk 6]
```diff
@@ -28,4 +28,12 @@
 #define ROC_ATTR_PRINTF(n_fmt_arg, n_var_arg)                                            \
     __attribute__((format(printf, n_fmt_arg, n_var_arg)))
 
+#ifdef GCC
+#define ATTRIBUTE_NO_SANITIZE_UB __attribute__((no_sanitize_undefined))
```
gavv, line 32: Let's rename it to ROC_ATTR_NO_SANITIZE_UB for consistency with other macros.

src/modules/roc_core/target_gcc/roc_core/attributes.h
```diff
@@ -28,4 +28,12 @@
...
+#elif CLANG
+#define ATTRIBUTE_NO_SANITIZE_UB __attribute__((no_sanitize("undefined")))
+#else
+#define ATTRIBUTE_NO_SANITIZE_UB
```
gavv, line 36: Please add a doxygen comment.

gavv:
It seems we lost `-fno-sanitize-recover`.

gavv:
Ping me when it'll be ready for review.

pvimal816:
Now it is ready for review.

SConstruct
```diff
@@ -1332,10 +1332,9 @@ sanitizers = env.ParseList(GetOption('sanitizers'), supported_sanitizers)
 if sanitizers:
     if not compiler in ['gcc', 'clang']:
         env.Die("sanitizers are not supported for compiler '%s'" % compiler)
-
+
```
gavv, line 1335: Nitpick: extra spaces.

SConstruct
```diff
@@ -1332,10 +1332,9 @@ sanitizers = env.ParseList(GetOption('sanitizers'), supported_sanitizers)
 if sanitizers:
     if not compiler in ['gcc', 'clang']:
         env.Die("sanitizers are not supported for compiler '%s'" % compiler)
-
+    
     for name in sanitizers:
-        flags = ['-fsanitize=%s' % name]
-
+        flags = ['-fsanitize=%s' % name, '-fno-sanitize-recover=%s' % name]
```
gavv, line 1337: https://github.com/roc-streaming/roc-toolkit/pull/412#discussion_r455096510

Don't we want to specify `-fno-sanitize-recover` only once with a list?

src/SConscript
```diff
@@ -218,3 +218,4 @@ if GetOption('enable_pulseaudio_modules'):
 
     env.Alias('pulse', targets, env.Action(''))
     env.AlwaysBuild('pulse')
+
```
gavv, line 221: Nitpick: extra empty line.

src/modules/roc_core/target_gcc/roc_core/attributes.h
```diff
@@ -28,4 +28,18 @@
...
     __attribute__((format(printf, n_fmt_arg, n_var_arg)))
 
+//! Used to suppress undefined behavior sanitizer for a particular function
+#if defined(__clang__)                                                                   \
+    && (__clang_major__ > 3 || __clang_major__ == 3 && __clang_minor__ >= 8)
+#define ROC_ATTR_NO_SANITIZE_UB __attribute__((no_sanitize("undefined")))
+
+#elif !defined(__clang__) && defined(__GNUC__)                                           \
+    && (__GNUC__ > 4 || __GNUC__ == 4 && __GNUC_MINOR >= 9)
+#define ROC_ATTR_NO_SANITIZE_UB __attribute__((no_sanitize_undefined))
```
gavv, line 38: Can we replace these version checks with something like this? (I didn't test it).

```c
#ifdef __has_feature
 #if __has_feature(memory_sanitizer)
  #if defined(__clang__)
   #define ROC_ATTR_NO_SANITIZE_UB ...
  #elif defined(__GNUC__)
   #define ROC_ATTR_NO_SANITIZE_UB ...
  #endif // defined(__GNUC__)
 #endif // __has_feature(memory_sanitizer)
#endif // __has_feature

#ifndef ROC_ATTR_NO_SANITIZE_UB
 //! Doxygen comment here.
 #define ROC_ATTR_NO_SANITIZE_UB
#endif
```

src/modules/roc_core/target_gcc/roc_core/attributes.h
```diff
@@ -28,4 +28,18 @@
 #define ROC_ATTR_PRINTF(n_fmt_arg, n_var_arg)                                            \
     __attribute__((format(printf, n_fmt_arg, n_var_arg)))
 
+//! Used to suppress undefined behavior sanitizer for a particular function
```
gavv, line 31: IIRC doxygen comment should be right before #define line.

Also, IIRC doxygen doesn't define macros like `__clang__` or `__GNUC__`, so the comment should be in `#else` block.

Also, doxygen comment should end with a dot (`.`), because we use auto-brief mode. In this mode, doxygen interprets the first sentence, i.e. everything before first dot, as a `@brief`.

src/modules/roc_core/target_gcc/roc_core/attributes.h
pvimal816: I have tested the suggested approach. But GCC does not support `__has_feature` macro.

gavv: Oh, really. Can we use __has_attribute(no_sanitize_undefined) for GCC?

pvimal816: > Oh, really.

Am I wrong? Does GCC indeed support `__has_feature` macro?



> Can we use __has_attribute(no_sanitize_undefined) for GCC?

Yes.

gavv: It seems that [the oldest](https://roc-streaming.org/toolkit/docs/portability/supported_platforms.html) supported clang and gcc already has __has_feature and __has_attribute, so we can assume they're always present.

gavv: A typo: `__has_feature(memory_sanitizer)` -> `__has_feature(undefined_behavior_sanitizer)`

gavv: > Am I wrong? Does GCC indeed support __has_feature macro?

No, you're right, __has_feature is clang-specific.

gavv: > It seems that the oldest supported clang and gcc already has __has_feature and __has_attribute, so we can assume they're always present.

Hm, maybe I'm wrong and we need an ifdef: https://gcc.gnu.org/gcc-5/changes.html

pvimal816: [This](https://gcc.gnu.org/gcc-5/changes.html) page mentions that `__has_attribute` macro is supported from `gcc-5` onwards.

SConstruct
gavv: Checked it, passing multiple options is also ok.

gavv:
Could you please rollback unnecessary white-space only changes (adding spaces and removing newlines) in scons scripts?

Also, I have tested this PR, and it seems that UBSAN does not print backtrace before exiting by default (unlike ASAN).

It can be enabled by exporting:

    UBSAN_OPTIONS=print_stacktrace=1

We already have crash handler that prints backtrace on segfaults, so it would be logical to enable UBSAN backtrace by default too.

AFAIK we can do it by adding a .cpp file with `__ubsan_default_options()` function (with a few attributes) which returns "print_stacktrace=1", and linking that .cpp into all executables. (But I think we shouldn't link it into libroc and pulseaudio modules; only tests, benchmarks, and tools).

Could yo try to add it too, please?

pvimal816:
> AFAIK we can do it by adding a .cpp file with `__ubsan_default_options()` function (with a few attributes) which returns "print_stacktrace=1", and linking that .cpp into all executables. (But I think we shouldn't link it into libroc and pulseaudio modules; only tests, benchmarks, and tools).

Is this an alternative of just declaring an environment variable `UBSAN_OPTIONS=print_stacktrace=1` to enable a backtrace?
If yes then why would we use this option as it requires us to link a new file with many existing files?

gavv:
> Is this an alternative of just declaring an environment variable UBSAN_OPTIONS=print_stacktrace=1 to enable a backtrace for undefined behavior sanitizer?

Yep.

>  If yes then why would we use this option as it requires us to link a new file with many existing files?

Because otherwise we'll have to export it in scons when running tests, in travis scripts when running tests directly without scons (there are a few places; currently without sanitizers though, but this may change in future), when running tests manually on hardware (e.g. I deploy test binaries to raspberry pi and run them by hand), and also ask user to export it if the user gets sanitizer error in tools or tests.

I think it would be much simpler just to hard-code it directly into binaries to be sure that it's always enabled no matter how the binary is run.

gavv:
BTW, we already have two examples of .cpp files linked into each test binary: test_main.cpp and bench_main.cpp. The idea is to build an object file from .cpp and then to add it to the list of sources of every executable.

pvimal816:
> I think it would be much simpler just to hard-code it directly into binaries to be sure that it's always enabled no matter how the binary is run.

Hmm. I'll try to do it.

src/library/src/ubsan_options.cpp
```diff
@@ -0,0 +1,21 @@
...
+    extern "C" ROC_ATTR_NO_SANITIZE_ADDR ROC_ATTR_NO_SANITIZE_UB                         \
+        __attribute__((visibility("default"))) __attribute__((used))
+
+SANITIZER_HOOK_ATTRIBUTE
+const char* __ubsan_default_options();
+
+SANITIZER_HOOK_ATTRIBUTE
+const char* __ubsan_default_options() {
+    return "print_stacktrace=1";
+}
```
gavv, line 21: It seems that the only attribute really needed here is `__attribute__((visibility("default")))`. Others can be removed. Or maybe I'm missing something? I've tested it on recent gcc and clang on my machine.

Also, let's move definition of `__attribute__((visibility("default")))` to `roc_core/attributes.h`.

After doing this, I think we can remove `SANITIZER_HOOK_ATTRIBUTE` macro. Since we don't need `ROC_ATTR_NO_SANITIZE_ADDR`, we can remove it too.

Finally, let's rename this file to more generic `sanitizer_options.cpp` (we may add options to other sanitizers in future) and move it from ` src/library/src` to `src` (it's not specific to libroc and event not used in libroc at all).

src/SConscript
```diff
@@ -116,6 +116,7 @@ if GetOption('enable_tests') or GetOption('enable_benchmarks'):
     cenv.MergeVars(tool_env)
     cenv.MergeVars(test_env)
     cenv.Append(CPPDEFINES=('ROC_MODULE', 'roc_test'))
+    ubsan_options = cenv.Object('library/src/ubsan_options.cpp')
```
gavv, line 119: It's better to add it to build only when sanitizers are enabled during build.

Also, the name is a bit confusing to me, I'd rename it to something like `sanitizer_options_obj` or `sanitizer_obj` or similar because "options" suggest that the variable contains some flags or settings, and actually it contains an object file.

src/SConscript
```diff
@@ -173,12 +174,13 @@ if GetOption('enable_tests') or GetOption('enable_benchmarks'):
...
 
 if not GetOption('disable_tools'):
+    ubsan_options = cenv.Object('library/src/ubsan_options.cpp')
     for tooldir in env.GlobDirs('tools/*'):
         cenv = env.Clone()
         cenv.MergeVars(tool_env)
         cenv.Append(CPPDEFINES=('ROC_MODULE', tooldir.name))
         cenv.Append(CPPPATH=['tools', '#src/tools/%s' % tooldir.name])
-
+
```
gavv, line 183: Nitpick: extra spaces.

src/SConscript
```diff
@@ -217,4 +219,4 @@ if GetOption('enable_pulseaudio_modules'):
         env.AddDistFile(env['ROC_PULSE_MODULEDIR'], target)
 
     env.Alias('pulse', targets, env.Action(''))
-    env.AlwaysBuild('pulse')
+    env.AlwaysBuild('pulse')
```
gavv, line 222: Nitpick: missing trailing newline.

Hint: you can detect such space-only changes by reviewing your changes on GitHub web interface after pushing them. It highlights them.

src/library/src/ubsan_options.cpp
pvimal816: > It seems that the only attribute really needed here is `__attribute__((visibility("default")))`. Others can be removed. Or maybe I'm missing something? I've tested it on recent gcc and clang on my machine.

I'm not much familiar with this. But please refer to comment(line 20-23) [here](https://gn.googlesource.com/gn/+/e375226345b2be2bfc8c4549702ee3e437b5c136/build/sanitizers/sanitizer_options.cc).

gavv: I see. So the reasoning on no_sanitize is to ensure that __ubsan_default_options() is not instrumented since it's called from sanitizer (though I also don't understand the exact problem). Then I'd suggest to add only ROC_ATTR_NO_SANITIZE_UB (since I supposed it's called only from UBSAN).

As for `__attribute__((used))`, I suppose it was needed to suppress a warning, but in our case it seems there is no warning.

src/modules/roc_core/target_gcc/roc_core/attributes.h
```diff
@@ -43,4 +41,7 @@
 #define ROC_ATTR_NO_SANITIZE_UB
 #endif
 
+//! TODO: add description.
+#define ROC_ATTR_VISIBILITY_DEFAULT __attribute__((visibility("default")))
```
gavv, line 45: I suggest to name it `ROC_ATTR_EXPORT`. It's a common name for such macros.

gavv:
Thanks!!

PR #414, opened 2020-07-19
Implement push_back_range for String List

alexandremgo:
Implementation of the method push_back_range which appends a string from two pointers begin and end to a string list.
Used in #300 .


gavv:
https://github.com/roc-streaming/roc-toolkit/pull/411#issuecomment-660657018

gavv:
LGTM, thanks!

gavv:
Follow-up commit: f4eb62b2d8662ddad87df01106e75c8d4c2ba5bf

Issue #198, opened 2019-05-27
Auto-detect appropriate audio backend


gavv:
If the driver and device are omitted, roc-recv and roc-send tools select the default driver and device. When it is done, they always prefer PulseAudio over ALSA if both were enabled during build.

If the tools were built with PulseAudio support enabled but are running on a system without PulseAudio daemon installed, this becomes a problem.

To fix this, we should check at run time whether PulseAudio daemon is really available on the system and don't try to use it if it's not.

We already have `sndio::BackendDispatcher` that selects an appropriate backend by calling `probe()` on all supported backends and selecting the first one that returned true. When the driver and device are omitted, it calls `probe(NULL, NULL)`.

Hence, all we need is to update `PulseaudioBackend::probe()` so that it will return false if the driver and device are NULL and the PulseAudio daemon is *not* available on the system.

It's a question how we can check this. Probably we can do it somehow using libpulse API.

See also #251. It provides an overview of roc_sndio concepts.

gavv:
See also #251. It provides an overview of roc_sndio concepts.

srp3003:
I am new to open source software development, may I have a go on this issue?

enigmaro:
Hello @gavv , I just looked at this project and it seems very exciting to contribute to. I was wondering if I could pick up on this issue if it has not been assigned to anyone else.

gavv:
@enigmaro Great, you're welcome, the issue is free.

gavv:
Hi, saw your message on IRC, but you're offline already, so answering here.

> Is it a valid approach to check if the PulseAudio daemon is running during run time to check if a simple context connection can be made to the default PulseAudio Server?

Sounds good. Let's try it and see how it'll work.

gavv:
BTW if you want to receive offline messages on IRC, see here https://roc-streaming.org/toolkit/docs/about_project/contacts.html

gavv:
@enigmaro Hi! Do you still have plans on this?

PR #417, opened 2020-07-22
auto-detect audio backend

enigmaro:
PulseaudioBackend probe performs runtime check for Pulseaudio Daemon.
SoxBackend chooses next available driver upon failure to open sink or source.


gavv:

gavv:
#198

src/modules/roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_backend.cpp
```diff
@@ -17,6 +17,22 @@
 namespace roc {
 namespace sndio {
 
+namespace {
+
+bool is_pulse_running() {
+    pa_context* context =
+        pa_context_new(pa_mainloop_get_api(pa_mainloop_new()), "ROC Daemon Check");
+
+    if (int err = pa_context_connect(context, NULL, PA_CONTEXT_NOFLAGS, NULL)) {
```
gavv, line 26: Several concerns here:

1. We call pa_mainloop_get_api(), but never call pa_mainloop_free(), it's a leak.

2. We don't check for return values of pa_mainloop_new(), pa_mainloop_get_api(), and pa_context_new(). If they return NULL, we'll crash. We should return false instead.

3. Is just calling pa_context_connect() really enough? Generally speaking, it's an asynchronous function, and I'd expect that we should run the main loop (currently we never start it) and wait for state callback, and then check whether connection failed or succeeded. Or is there a reason why we don't need to wait callback in this specific case?

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -252,8 +268,11 @@ ISink* SoxBackend::open_sink(core::IAllocator& allocator,
...
     }
 
-    if (!sink->open(driver, output)) {
-        return NULL;
+    while (!sink->open(driver, output)) {
+        driver = get_fallback_driver(driver);
+        if (!driver) {
+            return NULL;
+        }
     }
```
gavv, line 276: This code will try to switch driver even if the user did specify the driver explicitly.

E.g. the user specified `pulse://default`, pulse is not running, and here we'll try `alsa://default`. This is not what the user asked. We should only try multiple drivers if the user didn't specify the driver at all.

A few lines above we call `select_defaults()`, which selects default driver in case if no driver was specified. But with our new approach there is no more such thing as a single default driver. So instead of selecting one default driver there, we should rework that piece and try drivers in a loop, until open succeeds, as you did it here.

Also, when we're trying multiple drivers, we should log the reason. Something like "failed to open sink FOO using driver BAR, trying next driver". Or some other message, the idea is to explain the user why we're doing what we're doing.

src/modules/roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_backend.cpp
gavv: It seems that correctly waiting for pa_context_connect() callback requires quite complicated code. However, we actually already have this code in PulseaudioSink implementation.

Maybe it would be wiser to get rid of probe() at all?

Currently, BackendDispatcher calls probe() in a loop until it succeeds, and then just call open_source() or open_sink() on the given backend. It seems we can easily replace it with a loop which calls open_source() or open_sink() until it succeeds. We just need to remove probe() and instead pass flags directly to open_source() and open_sink(). We also should move corresponding flag checks from probe to open, of course.

This approach will allow us to re-use existing pulseaudio logic, and also will avoid creating pulseaudio connection twice, which is nice.

src/modules/roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_backend.h
```diff
@@ -12,6 +12,8 @@
 #ifndef ROC_SNDIO_PULSEAUDIO_BACKEND_H_
 #define ROC_SNDIO_PULSEAUDIO_BACKEND_H_
 
+#include <pulse/pulseaudio.h>
```
gavv, line 15: If pulseaudio headers are not used in `pulseaudio_backend.h`, it's better to move include to `pulseaudio_backend.cpp`, to reduce pollution of the global namespace of `pulseaudio_backend.h` users. This is a general rule for improving build time.

src/modules/roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_backend.cpp
enigmaro: @gavv , thanks for your comments. Yes, you are right about pa_context_connect() being an async call. I overlooked that. Your suggestion might be wiser to implement. I will work on it soon.

gavv:
Thanks for update, I'll take a look.

BTW, how did you test it?

enigmaro:
I tested this on my fedora 30 laptop machine and RPi3. I wrote shell scripts to spawn the roc-recv on my laptop and roc-send on the RPi3 (vice-versa as well). I tried them first with default commands and then added the testing args for specifying driver and device name. Finally, also added testing args for file devices. I tested this with pulseaudio enabled first on both machines and then disabled. I know this is not extensive, Do you suggest that I should write dedicated pipeline tests for this?

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -252,26 +257,51 @@ ISink* SoxBackend::open_sink(core::IAllocator& allocator,
...
+        } else {
+            roc_log(LogInfo,
+                    "failed to open sink=(%s) using driver=(%s), trying next driver",
+                    output, driver);
+            driver = get_fallback_driver(driver);
+            if (!driver) {
+                return NULL;
+            }
+        }
+    } while (!sink->open(driver, output));
```
gavv, line 289: The new algorithm looks good, but to me it's a bit hard to decrypt it from the implementation.

If I understand the implementation correctly, it can be simplified to this (pseudo-code):

```
if (driver != NULL) {
    if (!check_handler(driver, output, ...)) {
        return NULL;
    }
    return try_open(driver, output, ...);
} else {
    for (size_t n = 0; n < SIZE(default_driver_priorities); n++) {
        if (!check_handler(default_driver_priorities[n], "default", ...)) {
            continue;
        }
        sink = try_open(default_driver_priorities[n], "default", ...);
        if (!sink) {
            continue;
        }
        return sink;
    }
    return NULL;
}
```

Note that if driver is NULL, then device/file name is also NULL because driver is decoded from IO URI scheme, and ether scheme is non-empty, or the whole URI is empty. Earlier, before URIs were introduced, it was not always true, and there are artefacts in code from that time.

It seems that select_default_driver(), select_default_device(), select_defaults(), get_fallback_driver() can be removed.

(select_default_device can be removed because all drivers in default_driver_priorities are device drivers, and thus it will always return string "default"; we can just use that string).

Also, I think we can rename default_driver_priorities to just default_drivers and add a comment that they're sorted in probing order. And let's remove "null" driver from the list. If no driver found, it's better to fail than to silently do nothing. I don't remember why I added it.

gavv: Also, could you please add a check that if the user specified a driver (driver != NULL), but is_driver_hidden() returns true for that driver, then open will fail? It seems we forgot to add it earlier.

gavv:
> I tested this on my fedora 30 laptop machine and RPi3. I wrote shell scripts to spawn the roc-recv on my laptop and roc-send on the RPi3 (vice-versa as well). I tried them first with default commands and then added the testing args for specifying driver and device name. Finally, also added testing args for file devices. I tested this with pulseaudio enabled first on both machines and then disabled. I know this is not extensive, Do you suggest that I should write dedicated pipeline tests for this?

Sounds good. Before merging the PR, I'll test in my environment too.

Currently we don't have any tests that work with real PulseAudio/ALSA devices, I think adding them is out of the scope of this PR. We have a bunch of tasks for adding "real-time" tests that will live in a separate repo and run on real hardware: https://github.com/roc-streaming/roc-toolkit/labels/rt-tests. I think it's a good place for tests with real audio devices too. For now we don't have an issue for that, but this would be definitely be a good addition. If you'd like to work on it in future, feel free to file an issue.

src/modules/roc_sndio/ibackend.h
```diff
@@ -42,20 +42,19 @@ class IBackend {
         FilterDevice = (1 << 3)
```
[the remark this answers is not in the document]
enigmaro: There used to be a check in Pulseaudio::Backend::probe() which checked to see if the filter_flags corresponded to a FilterSink which now is moved to Pulseaudio::open_sink().Was this check not needed at all in that case?

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -140,6 +140,41 @@ bool is_driver_hidden(const char* driver) {
     return false;
 }
 
+bool is_fmt_handler_valid(const char* inout, const char* driver, int filter_flags) {
```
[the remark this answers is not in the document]
enigmaro: Im thinking of renaming is_fmt_handler_valid() to either fmt_handler_flags_match(..) or is_handler_matching(..). I just didn't want the function name to grow too big so I was favoring the second one. Let me know what you prefer.

src/modules/roc_sndio/ibackend.h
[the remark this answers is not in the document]
gavv: That check guaranteed that we wont select PulseaudioBackend when opening sources (because it supports only sinks). The previous logic was to select backend first, and then call open, so it was important to select right backend.

Now we're just calling open on all backends iteratively. PulseaudioBackend::open_source() will return null, and backend dispatched will go to the next backend, so everything is ok.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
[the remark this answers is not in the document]
gavv: Both sound good.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
enigmaro: > The new algorithm looks good, but to me it's a bit hard to decrypt it from the implementation.
> 
> If I understand the implementation correctly, it can be simplified to this (pseudo-code):
> 
> ```
> if (driver != NULL) {
>     if (!check_handler(driver, output, ...)) {
>         return NULL;
>     }
>     return try_open(driver, output, ...);
> } else {
>     for (size_t n = 0; n < SIZE(default_driver_priorities); n++) {
>         if (!check_handler(default_driver_priorities[n], "default", ...)) {
>             continue;
>         }
>         sink = try_open(default_driver_priorities[n], "default", ...);
>         if (!sink) {
>             continue;
>         }
>         return sink;
>     }
>     return NULL;
> }
> ```
> 
> Note that if driver is NULL, then device/file name is also NULL because driver is decoded from IO URI scheme, and ether scheme is non-empty, or the whole URI is empty. Earlier, before URIs were introduced, it was not always true, and there are artefacts in code from that time. -- Is it always true that the file name will be NULL when driver is NULL? Because when we have URIs that point to a filepath like 'file:/home/roc/file.wav' ... the driver name is NULL but the device/file name is the input right?. Sorry if I am misunderstanding you, Just wanted to confirm my belief.

enigmaro:
> > I tested this on my fedora 30 laptop machine and RPi3. I wrote shell scripts to spawn the roc-recv on my laptop and roc-send on the RPi3 (vice-versa as well). I tried them first with default commands and then added the testing args for specifying driver and device name. Finally, also added testing args for file devices. I tested this with pulseaudio enabled first on both machines and then disabled. I know this is not extensive, Do you suggest that I should write dedicated pipeline tests for this?
> 
> Sounds good. Before merging the PR, I'll test in my environment too.
> 
> Currently we don't have any tests that work with real PulseAudio/ALSA devices, I think adding them is out of the scope of this PR. We have a bunch of tasks for adding "real-time" tests that will live in a separate repo and run on real hardware: https://github.com/roc-streaming/roc-toolkit/labels/rt-tests. I think it's a good place for tests with real audio devices too. For now we don't have an issue for that, but this would be definitely be a good addition. If you'd like to work on it in future, feel free to file an issue.

That sounds interesting. I will work on creating tests once I am done with the couple other issues that I am working on now.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
gavv: Oh, really, you're right!

So...

* If driver is not NULL, output is also not NULL (then use given driver and output).
* If driver is NULL, output may be NULL (then find default driver and default output) or not NULL (then find appropriate driver for given output).

Could you please document these rules in comments to IBackend?

In other words, three cases are possible:

`driver == NULL && output == NULL` => loop through default driver list and use "default" output
`driver == NULL && output != NULL` => just open given driver and output (SoX will select driver automatically)
`driver != NULL && output != NULL` => just open given driver and output 

And condition in the pseudo-code above may be fixed as:

```
if (output != NULL) {
    < just open >
} else {
    roc_panic_if_not(driver == NULL);
    < loop >
}
```

The fourth case, `driver != NULL && output == NULL`, is not possible with current command-line interface, if I'm not mistaken. It would mean "use given driver and select default device for it". For device drivers, same can be achieved by using notation `pulse://default`, `alsa://default`, etc. For file drivers, default output makes little sense. Using stdin/stdout ("-") as default input/output doesn't look like a good idea.

So I think it's right that the fourth case is not supported. And since command-line forbids it, a panic should be okay.

gavv:
> That sounds interesting. I will work on creating tests once I am done with the couple other issues that I am working on now.

Awesome!

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
enigmaro: Okay great, I will do that. Is it okay to add these rules as a multiline comment in IBackend before the declarations of open_sink and open_source?

gavv: You can add it to class-level comment. IIRC comment should be attached either to a class or a method to be handled by doxygen.

src/modules/roc_sndio/ibackend.h
```diff
@@ -23,17 +23,21 @@ namespace roc {
...
+
+/*!
+ * Allows three cases of driver and device combinations.
+ * When driver is NULL and device is NULL, loop through default driver list occurs and we
+ * use "default" device to call open. When driver is NULL and device is not NULL, open is
+ * performed for given driver and output. When driver is not NULL and device is not NULL,
+ * open is performed for given driver and output. A fourth case is not meaningful, hence
+ * it is not supported and is forbidden by command-line.
+ */
 class IBackend {
```
gavv, line 35: A few nitpicks:

* Please convert it to a single doxygen comment block in one style (C++-style comments), like:

    ```
    //! Backend interface.
    //!
    //! Allows three cases of driver [.....]
    class IBackend {
    ```

* It's worth to mention that we're talking about open_source() and open_sink() methods.

* I think it would be more readable if we format the cases as a list.

src/modules/roc_sndio/ibackend.h
```diff
@@ -23,17 +23,21 @@ namespace roc {
...
     //! Probing flags.
     enum FilterFlags {
-        //! Input or output may be a sink.
-        FilterSink = (1 << 0),
-
-        //! Input or output may be a source.
-        FilterSource = (1 << 1),
 
         //! Input or output may be a file.
         FilterFile = (1 << 2),
```
gavv, line 43: Nitpick: since we removed values from beginning, we should adjust the remaining values to start from `(1 << 0)` again.

src/modules/roc_sndio/ibackend.h
```diff
@@ -23,17 +23,21 @@ namespace roc {
...
+ * performed for given driver and output. When driver is not NULL and device is not NULL,
+ * open is performed for given driver and output. A fourth case is not meaningful, hence
+ * it is not supported and is forbidden by command-line.
+ */
 class IBackend {
 public:
     virtual ~IBackend();
 
     //! Probing flags.
     enum FilterFlags {
```
gavv, line 40: Nitpick: we should fix comment "Probing flags." because there is no more probing.

src/modules/roc_sndio/ibackend.h
```diff
@@ -23,17 +23,21 @@ namespace roc {
...
-        //! Input or output may be a sink.
-        FilterSink = (1 << 0),
-
-        //! Input or output may be a source.
-        FilterSource = (1 << 1),
```
gavv, line 41: Nitpick: extra empty line here:

```
    enum FilterFlags {

        //! Input or output may be a file.
```

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -252,27 +202,59 @@ ISink* SoxBackend::open_sink(core::IAllocator& allocator,
...
+        }
+
+        if (!sink->open(driver, output)) {
+            return NULL;
+        }
+        return sink.release();
     }
 
-    return sink.release();
+    else {
```
gavv, line 221: Nitpick:

```
    }

    else {
```

should be formatted as:

```
    } else {
```

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -252,27 +202,59 @@ ISink* SoxBackend::open_sink(core::IAllocator& allocator,
...
+        roc_panic_if_not(driver == NULL);
+        roc_log(LogInfo, "auto-detecting appropriate backend");
```
gavv, line 223: Let's move this to LogDebug.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -252,27 +202,59 @@ ISink* SoxBackend::open_sink(core::IAllocator& allocator,
...
+        output = "default";
+        for (size_t n = 0; n < ROC_ARRAY_SIZE(default_drivers); n++) {
+            if (sox_find_format(default_drivers[n], sox_false)) {
+                const sox_format_handler_t* handler =
+                    sox_write_handler("default", default_drivers[n], NULL);
+                if (!check_handler_flags(handler, filter_flags)) {
+                    return NULL;
```
gavv, line 230: Let's also write explanation to log when returning NULL here.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -202,47 +184,15 @@ void SoxBackend::set_frame_size(core::nanoseconds_t frame_length,
...
-
     first_created_ = true;
 
     driver = map_to_sox_driver(driver);
-
-    if (!select_defaults(driver, output)) {
-        return NULL;
-    }
-
     core::ScopedPtr<SoxSink> sink(new (allocator) SoxSink(allocator, config), allocator);
```
gavv, line 196: It's not very good that we're reusing sink variable and call open() multiple times. It's too easy to make a mistake in sink implementation and leave the sink in some invalid state after a failed open, which would affect next open call. So better re-create sink each time we try open.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -252,27 +202,59 @@ ISink* SoxBackend::open_sink(core::IAllocator& allocator,
...
+
+        const sox_format_handler_t* handler = sox_write_handler(output, driver, NULL);
+        if (!check_handler_flags(handler, filter_flags)) {
+            return NULL;
+        }
+
+        if (!sink->open(driver, output)) {
+            return NULL;
+        }
+        return sink.release();
```
gavv, line 218: Can we extract this block (is hidden check + hander flags check + open) into a function and reuse here and in a loop? This will make open_source() much shorter.

We can also make it a template function that can work with both sources and sinks and use it in both open_source() and open_sink().

gavv:
It seems after removing probe, we have one new problem.

If the user specifies `pulse://somename`, we'll first go to PulseaudioBackend, and if it returns NULL, we'll then go to SoxBackend.

It means that if there is some bug in PulseaudioSink which causes open failure, instead of reporting failure, we now will silently switch to SoX backend and pretend that everything is OK and continue working, but with higher latency.

That's not very good. If Roc is built with PulseAudio support, I'd prefer to be sure that it will always use PulseaudioSink and will never use SoxSink for PulseAudio, to avoid confusion.

The problem is that now we can't distinguish two cases: 1) backend doesn't support given driver and device; 2) backend supports them but open failed.

In case (1), we should try another backend. In case (2), we should report failure and exit. Earlier we distinguished them because (1) was handled by probe, and (2) was handler by open. But now we just get NULL from open in both cases.

To fix this, we can, for example, introduce status codes returned by open_source() and open_sink(), like: StatusOK, StatusNotSupported, StatusOpenFailed; and handle them in backend dispatcher loop.

Thoughts?

enigmaro:
> It seems after removing probe, we have one new problem.
> 
> If the user specifies `pulse://somename`, we'll first go to PulseaudioBackend, and if it returns NULL, we'll then go to SoxBackend.
> 
> It means that if there is some bug in PulseaudioSink which causes open failure, instead of reporting failure, we now will silently switch to SoX backend and pretend that everything is OK and continue working, but with higher latency.
> 
> That's not very good. If Roc is built with PulseAudio support, I'd prefer to be sure that it will always use PulseaudioSink and will never use SoxSink for PulseAudio, to avoid confusion.
> 
> The problem is that now we can't distinguish two cases: 1) backend doesn't support given driver and device; 2) backend supports them but open failed.
> 
> In case (1), we should try another backend. In case (2), we should report failure and exit. Earlier we distinguished them because (1) was handled by probe, and (2) was handler by open. But now we just get NULL from open in both cases.
> 
> To fix this, we can, for example, introduce status codes returned by open_source() and open_sink(), like: StatusOK, StatusNotSupported, StatusOpenFailed; and handle them in backend dispatcher loop.
> 
> Thoughts?

I agree we cannot distinguish the two but if we want to distinguish case (2) and (1), there is another problem I think. We won't be able to say if open failed because of a bug or because the backend supports the driver and device but daemon was not running right? Status codes seem like a good way to go, but I think we might need to handle them in PulseaudioBackend::open_sink() as well to distinguish whether open failed because of a bug or because pulse daemon failed. Does that concern make sense?

enigmaro:
Oh no, I think my previous comment was wrong. I think what you said makes sense. This is when the user specifies the driver so yes, I think what you said makes complete sense. Its not important if the daemon failed in the case when user doesn't mention default driver and device anyway. So yes, I can try what you said.

gavv:
Hmm, good catch, and it seems it ruins the idea. If driver and device are NULL, and PulseaudioBackend::open_sink() returned StatusOpenFailed, we can't report error, we have to go to SoxBackend. However SoxBackend doesn't know that it shouldn't try pulseaudio driver.

One way to fix it is to move iteration through default drivers from backends to backend dispatcher. Backend dispatcher can iterate all backends during initialization, ask each one for the list of default drivers, and collect all default drivers in a single list. Then, when the user asks to open a sink without specifying driver, backend dispatcher will iterate default driver list and try to open default device using each of them.

The last step can be implemented using status codes (StatusNotSupported means go to next backend; StatusOpenFails means stop and report error).

However, if we're going this way and are collecting driver list in backend dispatcher, we can make one step further and just build a mapping of drivers to backends in backend dispatcher. E.g. a list of triplets `<driver, backend, is_default>`.

Then we can use this list for two purposes 1) iterate default drivers; 2) find appropriate backend by driver name.

enigmaro:
So what would happen when open fails with PulseaudioBackend and user doesn't specify driver. Do we mark that as StatusNotSupported and continue try open with other drivers with Sox then?

gavv:
If the user doesn't specify driver, we should iterate default drivers until pen succeeds. Since we will do it in backend dispatcher, not in backend, we can be sure that each driver (e.g. "pulseaudio") will be tried only once.

enigmaro:
> Hmm, good catch, and it seems it ruins the idea. If driver and device are NULL, and PulseaudioBackend::open_sink() returned StatusOpenFailed, we can't report error, we have to go to SoxBackend. However SoxBackend doesn't know that it shouldn't try pulseaudio driver.
> 
> One way to fix it is to move iteration through default drivers from backends to backend dispatcher. Backend dispatcher can iterate all backends during initialization, ask each one for the list of default drivers, and collect all default drivers in a single list. Then, when the user asks to open a sink without specifying driver, backend dispatcher will iterate default driver list and try to open default device using each of them.
> 
> The last step can be implemented using status codes (StatusNotSupported means go to next backend; StatusOpenFails means stop and report error).
> 
> However, if we're going this way and are collecting driver list in backend dispatcher, we can make one step further and just build a mapping of drivers to backends in backend dispatcher. E.g. a list of triplets `<driver, backend, is_default>`.
What do you suppose the `is_default` flag can help for? Aren't we only querying backends for default drivers? So won't they all be default drivers anyway?
> 
> Then we can use this list for two purposes 1) iterate default drivers; 2) find appropriate backend by driver name.

gavv:
In two last paragraphs, I'm talking about more generic idea. We can not only accumulate default drivers, but build a list of all drivers and their backends. This mapping can be used for opening both default and other drivers, and also can be used to iterate default drivers. Instead of iterating backends  in open, we will iterate drivers in this list. So opening default and non-default drivers will use the same mechanism, for better consistency. What do you think?

enigmaro:
> In two last paragraphs, I'm talking about more generic idea. We can not only accumulate default drivers, but build a list of all drivers and their backends. This mapping can be used for opening both default and other drivers, and also can be used to iterate default drivers. Instead of iterating backends in open, we will iterate drivers in this list. So opening default and non-default drivers will use the same mechanism, for better consistency. What do you think?

I think that makes sense. I will go ahead and cumulate the list of default drivers by querying the backends in the dispatcher, do you have any directions on how we can accumulate all non-default drivers for a given backend though? Is it correct to assume that PulseaudioBackend will not return any non-default drivers when a call is made to query the supported drivers? Also, in the case of SoxBackend, the non-default drivers would be the hidden drivers?

gavv:
We already have `get_driver(StringList&)` that can used to obtain the list of all drivers supported by backend.

We can enhance it by replacing StringList with an array of structs:

```
bool get_driver(Array<DriverInfo>&);

struct DriverInfo {
    char name[16];
    unsigned flags;
    IBackend* backend;
};

enum DriverFlags {
    DriverDefault, // try this driver when no driver specified
    DriverFile, // this file driver
    DriverDevice, // this is device driver
    DriverSource, // this driver supports sources
    DriverSink, // this driver supports sinks
};
```

(You can choose different names).

PulseaudioBackend will add a single driver with name "pulse" and flags `DriverDefault|DriverDevice|DriverSink`.

SoxBackend will add multiple drivers with different flags. It will add DriverDefault flag for drivers that are currently in default_drivers list.

Hidden drivers are drivers supported by sox but excluded from driver list, i.e. roc pretends that there are no such drivers at all. The reason is that they are redundant or wont work properly.

enigmaro:
@gavv -- I plan on making the Array of DriverInfo objects a member of the BackendDispatcher class. However, this requires an allocator to be initialized right? 

When BackendDispatcher is instantiated in roc_{recv,send,conv}, I see that it passes the context.allocator() to open_sink() and open_source() methods. Would it be okay to pass the same allocator for Array<DriverInfo> as well? The other way I assume is to create a separate HeapAllocator but I am not sure which one is better?

enigmaro:
Just realized I could do this with a small embedded capacity.

gavv:
Yeah, all allocations should go through context allocator. Currently sndio is not used in C API, but it will be in future, and in C API it's possible to have multiple context objects, each one with its own allocator. Maybe we'll allow to provide custom allocator for context.

Currently BackendDispatcher is a singleton. We can make it non-singleton and pass allocator to ctor instead of open.

enigmaro:
@gavv, I changed the implementation based on our discussion. However, I think we can differentiate cases without the need for status codes.

I separated the cases when user specifies input and does not. When user does not specify input, we iterate until success for a default driver (sorted by priority) and ensured that each driver is tried exactly once. For eg., if we build with pulse support and if opening pulse fails with PulseaudioBackend, SoxBackend will not try to open pulse again.

Otherwise, when user provides input, we now find the appropriate backend for the given device driver and try open. 
If it is a file input, currently we just use SoxBackend to open file type devices. If an appropriate backend cannot be found for a given driver this means that the driver is not supported by any of the available backends and we report this and exit. If open fails, we report failure for open and exit. 

Is there any other reason that we would want these status codes? It seems to me like they may be redundant unless I am missing a point. We can discuss this further when you review the code.

gavv:
@enigmaro Sorry for huge delay, I finally have time for Roc again. I'm now reviewing the changes.

> Is there any other reason that we would want these status codes? It seems to me like they may be redundant unless I am missing a point. We can discuss this further when you review the code.

I think you're right and status coded are not needed with the new approach. StatusNotSupported is now impossible because backend dispatcher wont try unsupported drivers. So only tow codes remain: StatusOK and StatusOpenFailed, and it can be just a boolean. So no changes are needed here.

src/modules/roc_sndio/backend_dispatcher.cpp
```diff
@@ -58,16 +58,30 @@ const char* select_input_output(const address::IoURI& uri) {
...
 #ifdef ROC_TARGET_PULSEAUDIO
     register_backend_(PulseaudioBackend::instance());
 #endif // ROC_TARGET_PULSEAUDIO
 #ifdef ROC_TARGET_SOX
     register_backend_(SoxBackend::instance());
 #endif // ROC_TARGET_SOX
+    init_driver_info_();
 }
 
 void BackendDispatcher::set_frame_size(core::nanoseconds_t frame_length,
```
gavv, line 87: There is a problem with set_frame_size(), related to making BackendDispatcher non-singleton. I've described it in a separate issue: #427. No more changes needed in this PR, just FYI.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -202,126 +238,72 @@ void SoxBackend::set_frame_size(core::nanoseconds_t frame_length,
...
 
-bool SoxBackend::get_drivers(core::StringList& list, int filter_flags) {
+bool SoxBackend::get_drivers(core::Array<DriverInfo>& list, int filter_flags) {
     core::Mutex::Lock lock(mutex_);
 
+    for (size_t n = 0; n < ROC_ARRAY_SIZE(default_drivers); n++) {
+        const char* driver = map_from_sox_driver(default_drivers[n]);
+        add_driver_mapping(list, driver, this,
+                           DriverDevice | DriverDefault | DriverSource | DriverSink);
+    }
```
gavv, line 283: If format is not enabled is SoX (at compile time or as a plugin), we shouldn't add corresponding driver; but in this loop we're adding all drivers from default_drivers unconditionally.

We do we need a separate loop for default drivers? Can we use a single loop below and just add corresponding flag if is_default() reports true, instead of skipping driver?

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
```diff
@@ -175,6 +166,51 @@ void log_handler(unsigned sox_level,
...
+    driver_info.set(driver, backend, driver_flags);
+    list.push_back(driver_info);
+}
+
+template <class T>
+bool check_and_open(const char* driver,
+                    const char* inout,
+                    int filter_flags,
+                    const core::ScopedPtr<T>& sinksource) {
+    if (driver && is_driver_hidden(driver)) {
```
gavv, line 188: It seems it can't happen with the new backend dispatcher algorithm: if driver is non-null, backend dispatcher will use only backend that explicitly specified it supports given driver. So we can remove this check.

gavv (commented):
Thanks for update! See comments below.

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.cpp
enigmaro: The reason I used a separate loop for default drivers is for the order of the priority in which they are supposed to be opened. If the loop below is used the format names from the handler are not returned in the same order. Without this order we will fallback to different drivers arbitrarily. Do you suggest an alternate method for this?

gavv: Ah, I see! Let's keep it as is. Could you please add corresponding comment to the first loop?

gavv:
I've finished and merged this PR. Thanks!

PR #419, opened 2020-07-25
Android tests on emulator

MatteoArella:
This PR add android tests on emulator (#396) for platforms based on i686 and x86_64 architectures.


gavv:
Thanks!

Comments:

* Travis job time exceeded 50 mins and is killed. Let's split it into two: "Android x86" and "Android ARM".

* It seems it's time to extract common part of android scripts into a generic script. Could you please extract either a script for running tests or a script for the whole thing (build + test) and re-use it?

gavv:

MatteoArella:
> Travis job time exceeded 50 mins and is killed. Let's split it into two: "Android x86" and "Android ARM".

Ok.

> It seems it's time to extract common part of android scripts into a generic script. Could you please extract either a script for running tests or a script for the whole thing (build + test) and re-use it?

Sure, I'll refactor into a single script :+1:

MatteoArella:
Some android tests are failing at multicast test bind:

```
src/tests/roc_netio/test_bind.cpp:174: error: Failure in TEST(bind, multicast)
	CHECK(add_udp_receiver(net_loop, rx_config, queue)) failed
```

On my workstation all tests are successful; I suppose that failed tests are due to travis networking issues, since on different builds tests aren't always failing on the same API/ABI platform.

@gavv what do you think about that?

.travis.yml
```diff
@@ -103,7 +103,29 @@ jobs:
...
+      dist: bionic
+      sudo: required
+      services:
+        - docker
+      before_install:
+        - docker pull rocstreaming/env-android
+        - sudo apt-get update
+        - sudo apt-get install -y qemu-kvm libvirt-bin bridge-utils virtinst virt-manager
+        - sudo adduser $USER libvirt
+        - sudo adduser $USER kvm
```
gavv, line 118: Why do we need kvm on travis if we're running tests inside docker?

scripts/travis/android/linux-android.sh
```diff
@@ -0,0 +1,54 @@
...
+        ;;
+    "aarch64")
+        arch=arm64-v8a
+        ;;
+    "armv7a")
+        arch=armeabi-v7a
+        sysroot="$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/arm-linux-androideabi"
+        ;;
+    *) echo "unrecognized android ABI."; exit 1;;
+esac
```
gavv, line 22: For readability, I'd suggest to rewrite this a bit more explicitly:

```
case "$ABI" in
    "x86_64")
        arch=...
        host=...
        sysroot=...
        ;;
    "i686")
        arch=...
        host=...
        sysroot=...
        ;;
    ....
esac
```

In other words, duplicate host in sysroot in every branch. The duplication is small, and it will be immediately clear how each abi is mapped to these variables.

scripts/travis/android/linux-android.sh
```diff
@@ -0,0 +1,54 @@
...
+target_prefix="/data/local/tmp"
+
+device create --name roc-device-$host$API --image default --api $API --arch $arch
+device start --name roc-device-$host$API
+
+adb shell "su 0 mkdir -p ${target_prefix}/lib"
+adb push "$sysroot/libc++_shared.so" $target_prefix/lib/libc++_shared.so 1> /dev/null
+
+readarray -d '' tests < <(find bin/$host$API \
+        -name 'roc-test-*' -not -name 'roc-test-library' -print0)
```
gavv, line 46: Why do we skip roc-test-library here? In qemu tests, we skip it because roc-test-library is multi-threaded, and qemu-arm has problems with threads.

gavv:
> Some android tests are failing at multicast test bind:
> 
> ```
> src/tests/roc_netio/test_bind.cpp:174: error: Failure in TEST(bind, multicast)
> 	CHECK(add_udp_receiver(net_loop, rx_config, queue)) failed
> ```
> 
> On my workstation all tests are successful; I suppose that failed tests are due to travis networking issues, since on different builds tests aren't always failing on the same API/ABI platform.
> 
> @gavv what do you think about that?

It seems that we can't bind to 224.0.0.1 in emulator. This test is supposed to be offline (I hope). Does it fail every time or from time to time?

It would be helpful to run `roc-test-netio -v` (on travis) and see the exact error message. It would be also helpful to run `ip maddr` inside emultator (if it's available there) and see the output. On my machine it has 224.0.0.1 entry for each interface, including `lo`.

If this test works on your machine inside emulator inside docker, than probably it's a configuration issue and we can somehow configure docker and/or emulator on travis to allow multicast.

If we can't, we should somehow detect that multicast is not available on current environment and skip the test in this case. To understand how, we need to know what exactly is failing (`-v` will show that).

Also, probably we can't use 224.0.0.1 in emulator, but can use some other multicast address?

.travis.yml
MatteoArella: kvm must be available on host system.
I was running tests on travis beta platform and sometimes hardware acceleration wasn't available so it was needed to install kvm explicitly; if hardware acceleration is always available on travis stable platform we can safely remove these lines.

scripts/travis/android/linux-android.sh
MatteoArella: Sure :+1:

scripts/travis/android/linux-android.sh
MatteoArella: I was skipping roc-test-library according to linux-arm tests scripts.
On platform i686 with android API 24 roc-test-library never make any progress and timeout always expires; in other tested cases (i686 platform with android API 28-29 and x86_64 platform with all android APIs) roc-test-library was successful; so maybe we can skip roc-test-library only for i686 platform with API 24.

MatteoArella:
> > Some android tests are failing at multicast test bind:
> > ```
> > src/tests/roc_netio/test_bind.cpp:174: error: Failure in TEST(bind, multicast)
> > 	CHECK(add_udp_receiver(net_loop, rx_config, queue)) failed
> > ```
> > 
> > 
> > On my workstation all tests are successful; I suppose that failed tests are due to travis networking issues, since on different builds tests aren't always failing on the same API/ABI platform.
> > @gavv what do you think about that?
> 
> It seems that we can't bind to 224.0.0.1 in emulator. This test is supposed to be offline (I hope). Does it fail every time or from time to time?

It fails every time:
- on i686 platform multicast tests are often successful with API 28 but only sometimes they are successful with API 24 and 29;
- on x86_64 platform multicast tests are successful most of the times but few times they fail on API 29.

> It would be helpful to run `roc-test-netio -v` (on travis) and see the exact error message. It would be also helpful to run `ip maddr` inside emultator (if it's available there) and see the output. On my machine it has 224.0.0.1 entry for each interface, including `lo`.

Here is an extract of the output of a failing test with verbose flag:

```
TEST(bind, multicast)15:26:02.925 [2601] [dbg] roc_netio: network loop: starting event loop
15:26:02.990 [2601] [inf] roc_netio: udp receiver: opened port 224.0.0.1:60708
15:26:03.047 [2601] [err] roc_netio: udp receiver: uv_udp_set_membership(): [ENODEV] no such device
15:26:03.049 [2601] [err] roc_netio: network loop: can't add port 224.0.0.1:0: can't start receiver
15:26:03.049 [2601] [inf] roc_netio: udp receiver: closing port 224.0.0.1:51599
15:26:03.052 [2601] [inf] roc_netio: udp receiver: closed port 224.0.0.1:51599
15:26:03.052 [2601] [dbg] roc_netio: network loop: asynchronous close finished: port 224.0.0.1:51599

src/tests/roc_netio/test_bind.cpp:174: error: Failure in TEST(bind, multicast)
	CHECK(add_udp_receiver(net_loop, rx_config, queue)) failed
```

so it seems that there are no network interfaces available for multicast; even though according the output got running `ip maddr` inside android emulator

```
1:	lo
	inet  224.0.0.1
	inet6 ff02::1
	inet6 ff01::1
3:	wlan0
	link  01:00:5e:00:00:01 users 2
	inet  224.0.0.1
	inet6 ff02::1
	inet6 ff01::1
6:	ip_vti0
	inet6 ff02::1
	inet6 ff01::1
7:	ip6_vti0
	inet6 ff02::1
	inet6 ff01::1
8:	sit0
	inet6 ff02::1
	inet6 ff01::1
9:	ip6tnl0
	inet6 ff02::1
	inet6 ff01::1
12:	radio0
	link  33:33:00:00:00:01
	link  01:00:5e:00:00:01
	link  33:33:ff:41:05:26
	link  33:33:ff:e9:e2:5c
	link  01:00:5e:00:00:fb
	link  33:33:00:00:00:fb
	inet  224.0.0.251
	inet  224.0.0.1
	inet6 ff02::fb
	inet6 ff02::1:ffe9:e25c
	inet6 ff02::1:ff41:526 users 2
	inet6 ff02::1
	inet6 ff01::1
```

224.0.0.1 multicast address is always available for all interfaces (`ip a` also returns MULTICAST flag enabled for all interfaces).

Adding also a route for multicast traffic with `ip route add 224.0.0.0/4 dev wlan0` won't solve this issue.

.travis.yml
gavv: Oh, I see. I don't know. If you're not sure, let's keep it.

scripts/travis/android/linux-android.sh
gavv: > so maybe we can skip roc-test-library only for i686 platform with API 24

Yes, let's do it. Could you please create an issue for that? Would be great if you could also provide logs and backtrace. The backtrace can be captured by attaching to hanged test using gdb or by sending SIGABRT to specific tid (not pid).

gavv:
So we get ENODEV when we're calling `setsockopt(IPV_ADD_MEMBERSHIP)` with address "0.0.0.0", which means "all interfaces".

Also it seems that multicast address ("224.0.0.1") is not present on all interfaces, according to `ip maddr`.

So here are some ideas to try:

* change multicast_interface from "0.0.0.0" to "127.0.0.1" in tests; same for IPv6;

* OR: add "224.0.0.1" to *every* interface;

* OR: try to add `table local` to the command: https://github.com/joyent/libuv/issues/1401#issuecomment-115709844

TBH I don't understand how this can depend on the API level. Maybe if we'll run tests long enough, they will eventually fail on any level?

gavv:
It seems that the travis job "Android x86" is still too long. I'd prefer to keep jobs under 30 minutes. Maybe also split "i686" and "x86_64"?

.travis.yml
MatteoArella: It seems that kvm is always available on travis stable platform so I've removed explicit kvm installation.

scripts/travis/android/linux-android.sh
MatteoArella: Sure I'll create an issue for that problem :+1: 
I've skipped roc-test-library for i686 platform for all APIs since it's not always successful.

MatteoArella:
> So we get ENODEV when we're calling `setsockopt(IPV_ADD_MEMBERSHIP)` with address "0.0.0.0", which means "all interfaces".
> 
> Also it seems that multicast address ("224.0.0.1") is not present on all interfaces, according to `ip maddr`.
> 
> So here are some ideas to try:
> 
>     * change multicast_interface from "0.0.0.0" to "127.0.0.1" in tests; same for IPv6;
> 
>     * OR: add "224.0.0.1" to _every_ interface;
> 
>     * OR: try to add `table local` to the command: [joyent/libuv#1401 (comment)](https://github.com/joyent/libuv/issues/1401#issuecomment-115709844)
> 
> 
> TBH I don't understand how this can depend on the API level. Maybe if we'll run tests long enough, they will eventually fail on any level?

Adding `table local` solved the problem!

gavv:
Sorry for delay. Thanks, LGTM!

Issue #410, opened 2020-07-13
Use thread local storage in fast_random()


gavv:
Currently, [core::fast_random()](https://github.com/roc-streaming/roc-toolkit/blob/da0383488b087e632d6fbcf9bd99be00670173c0/src/modules/roc_core/target_posix/roc_core/fast_random.cpp#L38) is implemented using pthread_once + pthread_mutex. This approach works, but creates an additional contention point between real-time threads, which is better to be avoided.

Instead of using pthread_once + pthread_mutex, we can implement that function using thread local storage (pthread_key_create and friends). We can just use a separate PRNG for every thread.

Prg1005:
Hi @gavv I would like to work on this issue, please let me know.

gavv:
@Prg1005 Great, you're welcome!

pvimal816:
> Instead of using pthread_once + pthread_mutex, we can implement that function using thread local storage (pthread_key_create and friends). We can just use a separate PRNG for every thread.

Can we use `__thread` keyword to have thread local seed. Then I think we don't need pthread_key_create.

gavv:
> @gavv 
> > Instead of using pthread_once + pthread_mutex, we can implement that function using thread local storage (pthread_key_create and friends). We can just use a separate PRNG for every thread.
> 
> Can we use `__thread` keyword to have thread local seed. Then I think we don't need pthread_key_create.

Regarding __thread, see my last comment in PR.

gavv:
I've created a new issue: #426.

PR #420, opened 2020-07-27
Replaced pthread_mutex with pthread_key_create

Prg1005:
PR for issue #410 

Using rand_seed global variable as a local storage key value, this avoids the dependency of a destructor. 

pthread_key_delete is needed but I am not sure when to call it!

### DIFF
--- a/src/modules/roc_core/target_posix/roc_core/fast_random.cpp
+++ b/src/modules/roc_core/target_posix/roc_core/fast_random.cpp
@@ -58,7 +58,7 @@ uint32_t fast_random(uint32_t from, uint32_t to) {
 
     pthread_once(&rand_key_once, do_init);
 
-    unsigned short *val = (unsigned short *) pthread_getspecific(rand_key);
+    unsigned short* val = (unsigned short*)pthread_getspecific(rand_key);
     if (val == NULL) {
         val = rand_seed;
         pthread_setspecific(rand_key, &rand_seed);
--- a/src/modules/roc_core/target_posix/roc_core/fast_random.cpp
+++ b/src/modules/roc_core/target_posix/roc_core/fast_random.cpp
@@ -18,14 +18,28 @@ namespace roc {
 namespace core {
 
 namespace {
+pthread_key_t rand_key;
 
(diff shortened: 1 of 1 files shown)


src/modules/roc_core/target_posix/roc_core/fast_random.cpp
```diff
@@ -18,15 +18,29 @@ namespace roc {
 namespace core {
 
 namespace {
+pthread_key_t rand_key;
```
gavv, line 21: Nitpick: missing empty line after "namespace {".

gavv:
pthread_key_delete becomes necessary when our library is loaded using dlopen() or similar mechanism. In this case, we should remove the key when the library is unloaded. However there is nor portable way to do this, AFAIK.

I've looked into pthread_key_create implementation in glibc, and it appears that using it in a library is not that good idea. The number of keys may be limited by implementation, and the limit may be low enough to be reached in practice. It's 1024 in glibc and even lower on some BSDs.

It means two things:

1. We can't assume that pthread_key_create() never fails. Probably the app using our library, or other libraries loaded by the app, already took all available keys. In this case it will return EAGAIN.
2. Calling pthread_key_delete() is really important (when using dlopen).

Given all these, it seems that using POSIX TLS doesn't worth it in our simple case, and it's better to avoid it here.

We could use compiler-specific __thread or C11/C++11 thread_local, but this would require to have a separate implementation of fast_random() for older compilers, which I think also doesn't worth it.

So I think we should close this issue and PR, unfortunately.

We can leave it as is for now, or try another approach. Instead of using nrand48() + TLS, we can adopt simple atomic-based lock-free PRNG algorithm. Actually it would be even better solution, since it'll be guaranteed to be lock-free (unlike TLS) and will be fully portable (we already have portable atomics).

The implementation can be very simple. For example, in mulberry32 linked above, it's enough to replace `state += 0x6D2B79F5` with an atomic add-and-fetch. Here is a real-world example of lock-free PRNG: https://github.com/liblfds/liblfds/blob/master/liblfds/liblfds7.1.0/liblfds710/inc/liblfds710/lfds710_prng.h

@Prg1005 thoughts?

gavv:
I've created a new issue: #426.

Issue #378, opened 2020-05-24
Introduce SampleSpec class


gavv:
We have numerous components in roc_audio and roc_sndio modules, which get two parameters in their constructor: sample_rate and either num_channels or channel_mask (often ch_mask).

They are good candidates to be united in a class with two fields and helper methods. It will simplify and unify the code and later will allow us to easily replace channel mask with more sophisticated channel set information (#86).

The new class should be called SampleSpec and placed into roc_audio module. It should contain two private fields, a constructor, getters, and setters.

We should also use SampleSpec instead of input_sample_rate + input_channels in pipeline::SenderConfig, and instead of output_sample_rate + output_channels in pipeline::ReceiverCommonConfig. And the same for pipeline::ConverterConfig.

After adding the class, we should check if we can also convert free functions packet::num_channels(), packet::ns_to_size(), and packet::size_to_ns() to methods of this class. If in all places where they're used we're actually using them with SampleSpec fields, we can do it.

soapdogg:
Hi, I would like to work on this issue.

gavv:
@soapdogg Hi, you're welcome!

gavv:
@soapdogg Hi, do you plan finishing the PR?

soapdogg:
Hi, 

Probably not. I have got caught up in other work. Feel free to unassign me from this task.

gavv:
Thanks for letting know. Unassigning, so that someone could pick it up. Feel free to ping me if you decide to get back to it.

nythoang:
Hi gavv, I stumbled on Roc while looking for my first open source project to work on. I read your about page and really like what you guys are doing. May I work on this issue?

gavv:
@nythoang Sure, you're welcome! Please take a look at #430 (last commits by Eric and discussions), it may be helpful.

nythoang:
Hi Victor, I ran the building commands in User Cookbook, then ran "scons -Q" to do a full build but got an error: "fatal error: of_openfec_api.h: No such file or directory". I also tried to install all dependencies with the commands in Developer Cookbook, with no success. Do you have an idea how I can solve this? For now I can run tests and build if I do --disable-openfec, so it is enough to get me started. Thank you!

gavv:
> Edit: I couldn't run the build locally without disabling openfec, but it runs fine with docker, so that's resolved for me!

Feed free to post build logs when you run build locally, ideally it should be possible to build without docker too.

gavv:
[...]

PR #430, opened 2020-10-14
Add SampleSpec class with just members, constructor, getters and setters

soapdogg:
Pull request for issue #378


gavv:
Thanks for the PR! I'll take a look in upcoming days.

The CI build is failing, this needs to be fixed. See [1](https://roc-streaming.org/toolkit/docs/development/continuous_integration.html) and [2](https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/building/developer_cookbook.rst) for hints how you can run everything locally.

Also please note that we should use `snake_case` for functions / methods / variables.

soapdogg:
Thanks for the feedback. I'll send out a revision soon!

src/modules/roc_audio/sample_spec.h
```diff
@@ -0,0 +1,54 @@
...
+
+class SampleSpec {
+public:
+    SampleSpec();
+    SampleSpec(size_t sample_rate, channel_mask_t channels);
+
+    size_t getSampleRate() const;
+    void setSampleRate(size_t sample_rate);
+    channel_mask_t getChannels() const;
+    void setChannels(channel_mask_t channels);
```
gavv, line 36: Some code style issues:

* Please use snake_case for method names.
* Could you rename getters from getFoo() to just foo()?
* Let's rename getChannels/setChannels to channel_mask()/set_channel_mask().
* Every public member should have Doxygen comment.
* Private methods should end with underscore: calc_num_channels -> calc_num_channels_.

src/modules/roc_audio/pcm_funcs.cpp
```diff
@@ -43,9 +43,9 @@ size_t pcm_encode_samples(void* out_data,
                           size_t out_offset,
                           const sample_t* in_samples,
                           size_t in_n_samples,
-                          packet::channel_mask_t in_chan_mask) {
-    const packet::channel_mask_t out_chan_mask = packet::channel_mask_t(1 << NumCh) - 1;
-    const packet::channel_mask_t inout_chan_mask = in_chan_mask | out_chan_mask;
+                          SampleSpec& sample_spec) {
+    const channel_mask_t out_chan_mask = channel_mask_t(1 << NumCh) - 1;
+    const channel_mask_t inout_chan_mask = sample_spec.getChannels() | out_chan_mask;
```
gavv, line 48: * Let's rename sample_spec to in_sample_spec to make it clear is it for input or output.
* Let's cache sample_spec.getChannels() in a variable, we can call it in_chan_mask.

Same for pcm_decode_samples, but with "out" instead of "in".

src/modules/roc_audio/sample_spec.h
```diff
@@ -0,0 +1,54 @@
...
+namespace audio {
+
+//! Bitmask of channels present in audio packet.
+typedef uint32_t channel_mask_t;
+
+//! Default sample rate, number of samples per second.
+const size_t DefaultSampleRate = 44100;
+
+//! Default channel mask.
+const channel_mask_t DefaultChannelMask = 0x3;
```
gavv, line 26: All defaults should stay in one place, roc_pipeline/config.h, because we want to avoid spreading various hard-coded values across the code base and be sure that they're not duplicated.

Thus, default SampleSpec constructor may initialize fields with zero, or, better, we can remove default constructor at all, if it wouldn't create too much problems.

And config structs in roc_pipeline should initialize their SampleSpec fields with appropriate defaults. Only roc_pipeline should know about those defaults.

gavv:
Hi, thanks again for the PR and sorry for delay. Big work!

You can find some comments below.

gavv:
Also:

* please fix travis build; the code should be compatible with old non-C++11 compilers; if desired, you can run travis build locally: https://roc-streaming.org/toolkit/docs/development/continuous_integration.html#run-locally

* please go through all places where we pass SampleSpec to methods and ctors and replace `SampleSpec` or `SampleSpec &` with `const SampleSpec &`.

* please go through all places where we declare SampleSpec field in a class and make it const if it's not modified after ctor

gavv:
IFrameEncoder and IFrameDecoder (and their implementations) don't use sample rate, only channel mask. And it would be nice to make it clear that they only do channel remapping, but not rate conversion. For this reason, could we revert the change in those classes and keep them using channel_mask_t?

src/modules/roc_audio/profiler.cpp
```diff
@@ -30,14 +29,13 @@ Profiler::Profiler(core::IAllocator& allocator,
...
     , last_chunk_samples_(0)
     , moving_avg_(0)
-    , sample_rate_(sample_rate)
-    , num_channels_(packet::num_channels(channels))
+    , sample_spec_(sample_spec)
     , valid_(false)
     , buffer_full_(false) {
-    if (num_channels_ == 0) {
+    if (sample_spec.num_channels() == 0) {
         roc_panic("profiler: n_channels is zero");
```
gavv, line 36: Nitpick: n_channels -> num_channels

src/modules/roc_audio/sample_spec.h
```diff
@@ -0,0 +1,54 @@
...
+    channel_mask_t getChannels() const;
+    void setChannels(channel_mask_t channels);
+    size_t num_channels() const;
+    packet::timestamp_diff_t timestamp_from_ns(core::nanoseconds_t ns) const;
+    core::nanoseconds_t timestamp_to_ns(packet::timestamp_diff_t ts) const;
+    size_t ns_to_size(core::nanoseconds_t frame_length) const;
+    core::nanoseconds_t size_to_ns(size_t frame_size) const;
+
+private:
+    size_t calc_num_channels();
```
gavv, line 44: Nitpick: This method can be made const.

src/modules/roc_packet/units.h
```diff
@@ -63,43 +63,7 @@ inline bool timestamp_le(timestamp_t a, timestamp_t b) {
...
-}
 
-//! Convert frame size to frame duration.
-inline core::nanoseconds_t
-size_to_ns(size_t frame_size, size_t sample_rate, packet::channel_mask_t ch_mask) {
-    return timestamp_to_ns(timestamp_diff_t(frame_size / num_channels(ch_mask)),
-                           sample_rate);
-}
 
 //! FEC block number in a packet stream.
```
gavv, line 68: Nitpick: Please remove extra empty lines: https://github.com/roc-streaming/roc-toolkit/blob/597f25c4508ad1e46b22559285ea38b07ecac077/src/modules/roc_packet/units.h#L65-L66

src/modules/roc_rtp/validator.h
```diff
@@ -57,7 +58,7 @@ class Validator : public packet::IReader, public core::NonCopyable<> {
     packet::PacketPtr prev_packet_;
 
     const ValidatorConfig config_;
-    const size_t sample_rate_;
+    const audio::SampleSpec& sample_spec_;
```
gavv, line 61: `const audio::SampleSpec& sample_spec_` -> `const audio::SampleSpec sample_spec_`

It's not safe to store a reference to outside value, its lifetime is unknown.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -22,7 +22,7 @@ SoxSink::SoxSink(core::IAllocator& allocator, const Config& config)
     , valid_(false) {
     SoxBackend::instance();
 
-    size_t n_channels = packet::num_channels(config.channels);
+    size_t n_channels = config.sample_spec.num_channels();
```
gavv, line 25: I suggest to get rid of n_channels variable at all.

src/modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -137,7 +137,8 @@ void SoxSink::write(audio::Frame& frame) {
 }
 
 bool SoxSink::setup_buffer_() {
-    buffer_size_ = packet::ns_to_size(frame_length_, sample_rate(), channels_);
+    audio::SampleSpec ss = audio::SampleSpec(sample_rate(), sample_spec_.getChannels());
+    buffer_size_ = ss.ns_to_size(frame_length_);
```
gavv, line 141: Here, we use sample_rate() to get up-to-date sample rate from `output_->signal`, which is updated in open().

Instead of constructing temporary SampleSpec here, let's update sample_spec_.rate in open() (set it to `output_->signal.rate`), so that we'll have correct sample_spec_ after open() and can use it here.

gavv: Same for SoxSource::setup_buffer_()

src/tests/roc_audio/test_depacketizer.cpp
```diff
@@ -39,6 +39,13 @@ const PcmFuncs& pcm_funcs = PCM_int16_2ch;
 } // namespace
 
 TEST_GROUP(depacketizer) {
+
+    channel_mask_t channels = {0x3};
+
+    SampleSpec sample_spec = SampleSpec(100, channels);
```
gavv, line 45: Let's keep sample rate (100) and channel mask (0x3) global constants (anon enum above). This way it's easy to find all test parameters in one place.

src/tests/roc_audio/test_mixer.cpp
```diff
@@ -23,13 +23,14 @@ namespace {
...
        ChannelMask = 0x1, MaxBufSz = 500 };
 
-const core::nanoseconds_t MaxBufDuration =
-    MaxBufSz * core::Second / (SampleRate * packet::num_channels(ChannelMask));
 
 core::HeapAllocator allocator;
 core::BufferPool<sample_t> buffer_pool(allocator, MaxBufSz, true);
 core::BufferPool<sample_t> large_buffer_pool(allocator, MaxBufSz * 10, true);
 
+const audio::SampleSpec sample_spec = audio::SampleSpec(SampleRate, ChannelMask);
```
gavv, line 31: Global constants should be named in CamelCase, here and in other tests.

src/tests/roc_audio/test_resampler.cpp
```diff
@@ -207,14 +202,14 @@ TEST(resampler, supported_scalings) {
         for (size_t pn = 0; pn < ROC_ARRAY_SIZE(profiles); pn++) {
             for (size_t fn = 0; fn < ROC_ARRAY_SIZE(frame_sizes); fn++) {
                 for (size_t irate = 0; irate < ROC_ARRAY_SIZE(rates); irate++) {
+                    SampleSpec sample_spec = SampleSpec(rates[irate], ChMask);
```
gavv, line 205: SampleSpec variable can be made const, here and below in this file.

src/tests/roc_audio/test_watchdog.cpp
```diff
@@ -63,11 +62,14 @@ class TestFrameReader : public IReader, public core::NonCopyable<> {
 
 TEST_GROUP(watchdog) {
     TestFrameReader test_reader;
+    channel_mask_t channels = {0x3};
```
gavv, line 65: Let's keep channel mask global constant.

src/tests/roc_pipeline/test_task_pipeline.cpp
```diff
@@ -1730,7 +1732,7 @@ TEST(task_pipeline, process_frame_min_samples_between_frames) {
 
     UNSIGNED_LONGS_EQUAL(1, pipeline.num_processed_frames());
 
-    UNSIGNED_LONGS_EQUAL(2, pipeline.num_pending_tasks());
+    //UNSIGNED_LONGS_EQUAL(2, pipeline.num_pending_tasks());
```
gavv, line 1735: Why is it commented?

gavv (commented):
A few more comments.

src/tests/roc_audio/test_packetizer.cpp
```diff
@@ -92,18 +93,18 @@ class PacketChecker {
         payload_decoder_.begin(pp->rtp()->timestamp, pp->rtp()->payload.data(),
                                pp->rtp()->payload.size());
 
-        sample_t samples[SamplesPerPacket * NumCh] = {};
+        sample_t samples[400] = {};
```
gavv, line 96: We shouldn't hard-code this.

src/tests/roc_audio/test_packetizer.cpp
```diff
@@ -156,7 +157,12 @@ class FrameMaker {
 
 } // namespace
 
-TEST_GROUP(packetizer) {};
+TEST_GROUP(packetizer) {
+    unsigned int PayloadType = 123;
+    unsigned int SamplesPerPacket = 200;
+    SampleSpec sample_spec = SampleSpec(SampleRate, ChMask);
+    const core::nanoseconds_t PacketDuration = SamplesPerPacket * core::Second / SampleRate;
```
gavv, line 164: Again, let's keep these global constants at the top of the file.

src/tests/roc_audio/test_packetizer.cpp
```diff
@@ -229,40 +235,40 @@ TEST(packetizer, multiple_buffers_one_packet) {
...
 TEST(packetizer, multiple_buffers_multiple_packets) {
-    enum {
-        NumFrames = 10,
-        NumSamples = (SamplesPerPacket - 1),
-        NumPackets = (NumSamples * NumFrames / SamplesPerPacket)
-    };
+    
+    unsigned int NumFrames = 10;
+    unsigned int NumSamples = (SamplesPerPacket - 1);
+    unsigned int NumPackets = (NumSamples * NumFrames / SamplesPerPacket);
```
gavv, line 251: These variables can be made const.

src/modules/roc_pipeline/receiver_session.cpp
```diff
@@ -134,12 +136,11 @@ ReceiverSession::ReceiverSession(const ReceiverSessionConfig& session_config,
...
         }
-
+        audio::SampleSpec sample_spec = audio::SampleSpec(format->sample_spec.getSampleRate(), session_config.sample_spec.getChannels());
         resampler_.reset(audio::ResamplerMap::instance().new_resampler(
                              session_config.resampler_backend, allocator,
                              sample_buffer_pool, session_config.resampler_profile,
-                             common_config.internal_frame_length, format->sample_rate,
-                             session_config.channels),
+                             common_config.internal_frame_length, sample_spec),
                          allocator);
```
gavv, line 144: When I was reviewing ReceiverSession, it was always confusing what is `session_config.sample_spec`.

We have two main sample specs:

* `format->sample_spec` - spec for packets coming from sender
* `common_config.output_sample_spec` - spec for samples coming to mixer, and then passed to audio device

The role of ReceiverSession is to transform the first to the second.

And what is `session_config.sample_spec`? Something intermediate? It's unclear.

This problem is not specific to this PR, it was present before too.

What I suggest is to get rid of `session_config.sample_spec` at all, i.e. remove ReceiverSessionConfig::sample_spec. Every time we need to configure a component with a sample spec, we will use either `format->sample_spec`, `common_config.output_sample_spec`, or create a temporary spec combining rate from one and channels from another.

Can we do it?

gavv (commented):
And some final comments.

PR #434, opened 2020-12-05
prevent task starvation

### LINKED ISSUE #380: Prevent task starvation in ctl::TaskQueue
ctl::TaskQueue is a thread-safe task queue, allowing to schedule task for immediate or delayed execution.

When it selects a task for execution, it always first checks the queue for immediate execution, and if it's empty, it then checks the queue for delayed execution.

While this is okay in most cases, there is a potential problem here. If the tasks for immediate execution will be added frequently enough during some period, tasks for delayed execution wont be handled during this period at all, even if their deadline is already expired.

A smoother behavior would be a more fair selection of tasks, for example: process one task from immediate queue, then process one task from delayed queue, and so on.

The new behavior should be covered with a unit test. Existing unit tests should be updated accordingly.

stevenzhu94:
closes #380

### DIFF
--- a/src/modules/roc_ctl/task_queue.cpp
+++ b/src/modules/roc_ctl/task_queue.cpp
@@ -49,6 +49,7 @@ TaskQueue::ICompletionHandler::~ICompletionHandler() {
 TaskQueue::TaskQueue()
     : started_(false)
     , stop_(false)
+    , fetch_ready_(true)
     , ready_queue_size_(0) {
     started_ = Thread::start();
 }
--- a/src/modules/roc_ctl/task_queue.cpp
+++ b/src/modules/roc_ctl/task_queue.cpp
@@ -49,7 +49,6 @@ TaskQueue::ICompletionHandler::~ICompletionHandler() {
 TaskQueue::TaskQueue()
     : started_(false)
     , stop_(false)
-    , fetch_ready_(true)
     , ready_queue_size_(0) {
     started_ = Thread::start();
 }
@@ -158,6 +157,7 @@ void TaskQueue::stop_and_wait() {
 
 bool TaskQueue::process_tasks_() {
     core::Mutex::Lock lock(task_mutex_);
+    bool fetch_ready_ = true;
 
     for (;;) {
         Task* task = NULL;
--- a/src/modules/roc_ctl/task_queue.cpp
+++ b/src/modules/roc_ctl/task_queue.cpp
@@ -49,6 +49,7 @@ TaskQueue::ICompletionHandler::~ICompletionHandler() {
 TaskQueue::TaskQueue()
     : started_(false)
     , stop_(false)
+    , fetch_ready_(true)
     , ready_queue_size_(0) {
     started_ = Thread::start();
 }
@@ -157,7 +158,6 @@ void TaskQueue::stop_and_wait() {
 
 bool TaskQueue::process_tasks_() {
     core::Mutex::Lock lock(task_mutex_);
-    bool fetch_ready_ = true;
 
     for (;;) {
         Task* task = NULL;
@@ -166,15 +166,17 @@ bool TaskQueue::process_tasks_() {
             task = fetch_ready_task_();
             if (!task) {
                 task = fetch_sleeping_task_();
+            } else {
+                fetch_ready_ = !fetch_ready_;
             }
         } else {
             task = fetch_sleeping_task_();
             if (!task) {
                 task = fetch_ready_task_();
+            } else {
+                fetch_ready_ = !fetch_ready_;
             }
         }
-
-        fetch_ready_ = !fetch_ready_;
 
         if (!task) {
             if (update_wakeup_timer_() == 0) {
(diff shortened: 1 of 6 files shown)


src/modules/roc_ctl/task_queue.cpp
```diff
@@ -157,14 +157,25 @@ void TaskQueue::stop_and_wait() {
...
+                task = fetch_sleeping_task_();
+            }
+        } else {
             task = fetch_sleeping_task_();
+            if (!task) {
+                task = fetch_ready_task_();
+            }
         }
 
+        fetch_ready_ = !fetch_ready_;
```
gavv, line 177: There is a minor issue here. We're inverting fetch_ready_ no matter of what type of task we have processed actually, ready or sleeping (both types are possible since if one queue is empty, we check another).

Thus, the following situation is possible:

* FIRST CALL: fetch_ready_ is true, ready queue is empty, sleeping queue is non-empty; we process sleeping task and set fetch_ready_ to false

* new tasks is added to ready queue

* SECOND CALL: fetch_ready_ is false, ready queue is non-empty, sleeping queue is non-empty; we process sleeping task again

So, we will process two sleeping tasks in a row, although ready queue is non-empty.

Not a big deal, but I think the algorithm would be a bit clearer if we took into account what type of task we have just processed and always process different type of task on next invocations, it they are present.

src/tests/roc_ctl/test_task_queue.cpp
```diff
@@ -1290,5 +1290,53 @@ TEST(task_queue, reschedule_cancelled) {
...
+
+    enum { NumTasks = 20 };
+
+    UNSIGNED_LONGS_EQUAL(0, tq.num_tasks());
+
+    TestHandler handler;
+    handler.expect_success(true);
+    handler.expect_n_calls(NumTasks);
+
+    TestTaskQueue::Task* tasks = new TestTaskQueue::Task[NumTasks];
```
gavv, line 1305: Since NumTasks is constant, we can use fixed-size array on stack.

src/tests/roc_ctl/test_task_queue.cpp
```diff
@@ -1290,5 +1290,53 @@ TEST(task_queue, reschedule_cancelled) {
...
+
+    TestTaskQueue::Task* temp = NULL;
+    for (size_t i = 0; i < NumTasks; i++) {
+        tq.unblock_one();
+        temp = handler.wait_called();
+        UNSIGNED_LONGS_EQUAL(i + 1, tq.num_tasks());
+    }
+
+    // check that the sleeping task wasn't last to get processed
+    CHECK(temp != &tasks[0]);
```
gavv, line 1330: Could we implement some stricter check here?

Currently we're checking that sleeeping task wont wait for NumTasks ready tasks. But we allow it to wait for NumTasks-1 ready tasks, for example.

Can we instead schedule a few sleeping and ready tasks and check that each sleeping task with expired deadline never waits more than one ready task?

gavv (commented):
Thanks for PR! Here are some comments.

It seems that the travis in develop is broken, I'll take a look.

src/modules/roc_ctl/task_queue.cpp
gavv: BTW it means that it's indeed better to make the flag field instead of var. But it doesn't have to be atomic, since it's accessed only from ctl thread.

gavv:
> It seems that the travis in develop is broken, I'll take a look.

Fixed, please rebase.

stevenzhu94:
Had to do some printouts to figure out why the tests were failing. Turns out there were some synchronization issues when creating the tests in a loop, so I changed the test to explicitly schedule and unblock one by one.

src/modules/roc_ctl/task_queue.cpp
stevenzhu94: I added else blocks after the task null checks so that the bool only flips on successful fetches

src/tests/roc_ctl/test_task_queue.cpp
stevenzhu94: changed it to use fixed-size array on stack

src/tests/roc_ctl/test_task_queue.cpp
stevenzhu94: changed the unit test to check for alternating queue fetching as requested

stevenzhu94:
the travis bionic builds initially failed the schedule_at_and_schedule unit test, but they passed after I ran them again. The alternating fetch should not affect the test unless the tasks scheduled using schedule_at become expired. I'll increase the delay on those tests and build again.

gavv:
LGTM!

gavv:
> the travis bionic builds initially failed the schedule_at_and_schedule unit test, but they passed after I ran them again. The alternating fetch should not affect the test unless the tasks scheduled using schedule_at become expired. I'll increase the delay on those tests and build again.

Thanks. Ideally in future we should use virtual time in tests, I've created an issue for that: #435.

Issue #435, opened 2020-12-07
Use virtual time in TaskQueue unit tests


gavv:
Currently unit tests for ctl::TaskQueue (`test_task_queue.cpp`) are partially bound to real clock time (`core::timestamp()` function) and due to this they can fail on loaded system, e.g. slow travis runners.

We can make them 100% reliable by using virtual clock instead of real time, i.e. clock that is fully controlled by unit tests.

We already use such approach in unit tests for pipeline::TaskPipeline (`test_task_pipeline.cpp`). TaskPipeline doesn't call `core::timestamp()` directly. Instead, it have a protected virtual method to get time:

```
virtual core::nanoseconds_t timestamp_imp() const
```

In `test_task_pipeline.cpp` we inherit TaskPipeline, override this method to return virtual time, and test this derived class instead of original TaskPipeline.

Now we could use the same approach `test_task_queue.cpp`:

* add virtual protected method timestamp_imp() to ctl::TaskQueue;
* override it in `test_task_queue.cpp`
* adjust unit tests to work with virtual time instead of real time, i.e. don't use sleep and instead set virtual time to desired values

4d696e6820427569:
Hi, can I work on this issue?

gavv:
Sure, you're welcome!

4d696e6820427569:
I have several questions:
 
 1. I have added a protected virtual method timestamp_imp() to ctl::TaskQueue. Since ctl::TestTaskQueue is a derived class, it needs to implement timestamp_imp(). Similar to pipeline::TestPipeLine, I added a public field time_ and set_time() to ctl::TestTaskQueue. Am I in the right direction?
 
 2. ctl::TestHandler has this public method
 
```
void expect_after(core::nanoseconds_t delay)
 {
     expect_after_ = timestamp() + delay;
 }
```

 Initially, I planned to replace timestamp() with a task_.timestamp_imp() but task_ could be null so it doesn't seem viable. For reference, task_ is a public field of TaskHandler with type TaskQueue::Task*
 
 Do I add a public field time_ and set_time() to ctl::TestHandler, similar to pipeline::TestPipeLine? The problem with this approach is we have to manually pass reasonable values to handler.expect_after() in the unit tests.
 
[...]

PR #437, opened 2020-12-13
Use virtual time in TaskQueue unit tests #435

minhdb:
Initial PR for #435 
I mainly want to check for the CI build status.

I built and ran the tests on a MacOS machine and a arch-based linux virtual machine.

Notes: 
ctl::TaskQueue thread relies on core::Timer wakeup_timer_
core::Timer has a core::Semaphore field. 
On MacOS machine, core::Semaphore depends on core::Timer::timestamp(). 

Edit: Failed CI. Investigating.


gavv:
Thanks for PR! I'll take a look at code and your comments/questions in upcoming days.

gavv:
It seems new ctl tests hang on Linux. Also, there are some warnings and the code is not formatted.

BTW, you can run CI builds locally: https://roc-streaming.org/toolkit/docs/development/continuous_integration.html#run-locally

Though I didn't try it on Mac.

gavv:
To format code, run `scons fmt` (you'll need clang-format installed).

gavv:
> ctl::TaskQueue thread relies on core::Timer wakeup_timer_

Stupid me, I forgot this! That's a problem!

core::Timer implicitly uses real clock. We pass absolute timestamps to it (`deadline`), and it uses core::Semaphore to wait until that time comes on real system clock.

So to make tests independent on real clock, it's not enough to replace core::timestamp() call, we should replace core::Timer calls too.

I suggest the following:

* create core::ITimer, an interface with core::Timer methods; make core::Timer implementing it;
* in TaskQueue, use ITimer instead of Timer
* in ControlLoop, pass Timer as ITimer implementation to TaskQueue
* add ITimer::get_time() virtual method and its implementation in Timer that returns core::timestamp()
* use ITimer::get_time() in TaskQueue instead of core::timestamp()
* in tests, create custom implementation of ITimer; it can have set_time() or add_time() method that will
  * update value returned by get_time()
  * unblock blocked wait_deadline() calls if new virtual time is ahead of corresponding deadline
* we can pass reference to our custom ITimer to both TestTaskQueue and TestHandler; TestHandler can use it to check current virtual time when the handler is invoked

4d696e6820427569:
Thanks for the hint! 

My schedule's a bit tight toward the end of the year but I will look into it asap.

4d696e6820427569:
School was a bit crazy in the previous months. If this issue is still open, I still have some questions and would like to discuss possible solution.

Is core::TaskQueue ever instantiated? Or we only instantiate its derived classes (ControlLoop, TestTaskQueue, etc) ?

> in TaskQueue, use ITimer instead of Timer

There's only one Timer field in TaskQueue, which is

`core::Timer wakeup_timer_;`

I have replaced Timer wakeup_timer_ with ITimer wakeup_timer_ in TaskQueue. 

`core::ITimer wakeup_timer_;`

ITimer is an abstract class (or interface) here. wakeup_timer_ is not visible to derived class so I'm planning to make it protected instead of private.

> in ControlLoop, pass Timer as ITimer implementation to TaskQueue

So, does this mean that for each derived class with TaskQueue as base class, I will need to pass a reference of core::ITimer type (core::Timer or ctl::TestTimer) to the derived class's constructor? 

For example, for ControlLoop we will have a constructor
`// Specifically, timer is core::Timer`
`ControlLoop(ITimer& timer)`

Or in TestTaskQueue
`// test_timer is our custom implementation of ITimer in test_task_queue.cpp`
`// Let that be ctl::TestTimer`
`TestTaskQueue(ITimer& test_timer)` 

Thank you!

gavv:
> School was a bit crazy in the previous months. If this issue is still open, I still have some questions and would like to discuss possible solution.

@4d696e6820427569 Good to hear from you! Yep, the issue is still open, you're welcome.

> Is core::TaskQueue ever instantiated? Or we only instantiate its derived classes (ControlLoop, TestTaskQueue, etc) ?

IIRC, it's never instantiated, it's a base class only.

> > in TaskQueue, use ITimer instead of Timer
> 
> There's only one Timer field in TaskQueue, which is
> 
> `core::Timer wakeup_timer_;`
> 
> I have replaced Timer wakeup_timer_ with ITimer wakeup_timer_ in TaskQueue.
> 
> `core::ITimer wakeup_timer_;`

Yep, I meant that timer.

> ITimer is an interface here. wakeup_timer_ is not visible to derived class so I'm planning to make it protected instead of private.
> 
> > in ControlLoop, pass Timer as ITimer implementation to TaskQueue
> 
> So, does this mean that for each derived class with TaskQueue as base class, I will need to pass a reference of core::ITimer type (core::Timer or ctl::TestTimer) to the derived class's constructor?
> 
> For example, for ControlLoop we will have a constructor
> `// Specifically, timer is core::Timer`
> `ControlLoop(ITimer& timer)`
> 
> Or in TestTaskQueue
> `// test_timer is our custom implementation of ITimer in test_task_queue.cpp`
> `// Let that be ctl::TestTimer`
> `TestTaskQueue(ITimer& test_timer)`

Yeah, this should work. Though, probably a bit cleaner solution would be to add pure virtual method to TaskQueue like:

```
virtual ITimer& timer() = 0;
```

...and implement it in derived classes. This way, we'll have only one timer field (in derived class) instead of two timer fields (ITimer in base class + Timer in derived class). Not a big deal really, but IMHO looks a bit better.

4d696e6820427569:
```
virtual ITimer& timer() = 0;
```
> This way, we'll have only one timer field (in derived class) instead of two timer fields (ITimer in base class + Timer in derived class)

I think it sounds cleaner as well with only one `core::ITimer` field. I'm still a bit confused to how I should do this though. 

When we make `wakeup_timer_` to be `core::ITimer` we run into a problem.

In `TaskQueue`, some functions/methods were implemented using `wakeup_timer_`. For instance,

```
void TaskQueue::run() {
    // This is called when a derived class's instantiated as a runnable
    // thread. There are functions/methods in TaskQueue similar to this as well.
    //....
    wakeup_timer_.wait_deadline();
    //...
}
```

In derived classes, TestTaskQueue for instance, this won't compile unless we explicitly state the exact type of `wakeup_timer_`. Because `core::ITimer` is an interface (or abstract class).

Let me know your thoughts. I will continue digging around to see what works. Thank you for your time and guidance!

gavv:
Let me demonstrate what I mean:

```cpp
class TaskQueue {
  //....
protected:
  virtual core::ITimer& timer() = 0;
  //....
  // no wakeup_timer_ field anymore
};

void TaskQueue::run() {
  //....
  // using timer() instead of wakeup_timer_
  // timer() is actually implemented in derived class
  // so it depends on derived class what specific type of timer we use here
  timer().wait_deadline();
  //...
}

void TaskQueue::stop_and_wait() {
  //...
  // same here
  timer().try_set_deadline(0);
 //...
}

class ControlLoop : public TaskQueue {
  //....
protected:
  // implement method defined in base class
  virtual core::ITimer& timer();
  //....
private:
  // wakeup_timer_ field is now here
  core::Timer wakeup_timer_;
};

core::ITimer& ControlLoop::timer() {
  return wakeup_timer_;
}

.....

// in tests:

class TestTaskQueue : public TaskQueue {
  //....
protected:
  // implement method defined in base class
  virtual core::ITimer& timer();
  //....
private:
  // here we use mock timer
  // MockTimer also implements core::ITimer
  MockTimer mock_timer_;
};

core::ITimer& TestTaskQueue::timer() {
  return mock_timer_;
}
```

This pattern is called [template method](https://en.wikipedia.org/wiki/Template_method_pattern) (no connection with C++ templates though).

4d696e6820427569:
Some CI tests from `roc_pipeline` are failing due to calling pure virtual function. From the debug trace it seems to happen in `TaskQueue::run()`. Maybe it's calling pure virtual function in a class's destructor/constructor somewhere.

I'm playing around with this solution

```cpp
class TaskQueue {
  //....
protected:
  // No longer pure virtual.
  virtual core::ITimer& timer();
  core::Timer wakeup_timer_;
};

void TaskQueue::run() {
  //....
  timer().wait_deadline();
  //...
}

core::ITimer TaskQueue() {
  return wakeup_timer_;
}

void TaskQueue::stop_and_wait() {
  //...
  timer().try_set_deadline(0);
 //...
}

class ControlLoop : public TaskQueue {
  // No wakeup_timer_ and timer() since it's inherited directly
  // from TaskQueue.
};
.....

// in tests:

class TestTaskQueue : public TaskQueue {
  //...
public:
  // Making this public since I want to call it
  // in CppUTest
  virtual core::ITimer& timer();
  //....
private:
  MockTimer mock_timer_;
};

core::ITimer& TestTaskQueue::timer() {
  return mock_timer_;
}

```

It's getting timeout for some reasons. I suspect I did not update the mock timer correctly.

Do you have any advice on debugging / setting break points? I usually just invoke this command and put print statement.
```
scons -Q test --enable-werror --enable-debug --sanitizers=all --enable-tests --build3rdparty=openfec
```

Also, can we close this PR and let me create a new PR instead? This old PR is quite behind from remotes/origin/develop.
Thank you so much!

gavv:
Hi!

> Do you have any advice on debugging / setting break points? I usually just invoke this command and put print statement.

See here on how to invoke test binaries manually: https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/building/developer_cookbook.rst#running-tests-manually

You can invoke it under debugger (gdb, cgdb, gdb integration for vim, etc).

> Also, can we close this PR and let me create a new PR instead? This old PR is quite behind from remotes/origin/develop.

Sure, you can close it, or you can just rebase your branch on fresh develop and force-push it. Actually there is no difference - github will detect force-push and update everything correctly.

gavv:
> ERROR: roc_core: semaphore: semaphore_wait(): (os/kern) invalid name

I suggest to google the part of this error `(os/kern) invalid name` and try to investigate when `semaphore_wait()` can report it. Probably we're invoking it after semaphore destruction or before its initialization.

gavv:
> Some CI tests from roc_pipeline are failing (not at the same test) due to calling pure virtual function. From the debug trace it seems to happen in TaskQueue::run(). Maybe it's calling pure virtual function in a class's destructor/constructor somewhere leading to undefined behavior.

Yep, you should ensure that timer() (and other virtual methods) are not called from ctor/dtor. You can achieve this by ensuring that we always call stop_and_wait() before destroying the derived class which implements timer() (e.g. the derived class dtor is calling stop_and_wait() before proceeding to the base class dtor) and also that we never access timer() after stop_and_wait() call.

gavv:
Closing this for now, feel free to reopen if you'll have a chance to continue work.

Issue #86, opened 2017-07-12
Configurable channel set


gavv:
What we need:

- configurable sender input channel set
- configurable receiver output channel set
- configurable network channel set

Features:

- [x] mono, stereo
- [x] surround sound (3.x, 4.x, 5.x, 6.x, 7.x)
- [x] multitrack recording (up to 1024 channels/tracks)
- [x] different orders (smtpe, alsa)
- [x] surround <=> surround mapping (upmixing & downmixing)
- [x] multitrack <=> multitrack mapping
- [x] surround <=> multitrack mapping
- [x] mapping between orders

Internals:

- [x] ChannelLayout (surround, multitrack)
- [x] ChannelOrder (order of surrounf channels in memory; smtpe, alsa)
- [x] ChannelPosition (identifiers of surround channels)
- [x] ChannelSet (multi-word channel mask + layout + order)
- [x] update SampleSpec
- [x] downmixing/upmixing tables
- [x] ChannelMapper
- [x] ChannelMapperReader
- [x] ChannelMapperWriter
- [x] update pipeline
- [x] rtp::FormatMap should provide layout, order, and mask

API:

- [x] mono
- [x] stereo
- [x] multitrack

CLI:

- [x] --io-encoding 
- [x] --packet-encoding

kaosbeat:
is it, at the moment, possible to have a multi-channel setup? 

for example jack-router allows to specify virtual outputs. They show up in my qjack-ctl
![Screen Shot 2021-03-04 at 14 34 58](https://user-images.githubusercontent.com/204628/109971854-f63e2900-7cf6-11eb-8963-9d93d6611df2.png)

How would I connect each one of them to a single roc-send? So my roc-recv on the other side just has to deal with 1 channel. 

When doing `roc-send -L` I get
```
supported device drivers:
  core
```
on OSX

How would I connect to the right device? And pass the right channel to each roc-send instance?

gavv:
Hi!

> is it, at the moment, possible to have a multi-channel setup?

Currently, roc always uses exactly two channels. I'm planning to fix this in 0.2 (upcoming major release).

> How would I connect each one of them to a single roc-send? So my roc-recv on the other side just has to deal with 1 channel.

Currently roc doesn't support JACK. But if you map each output port to a separate ALSA device, you can run an instance of roc-send for each device.

If somebody will add support for JACK in future, it could be possible to connect roc-send directly to a JACK port.

[...]

PR #444, opened 2021-03-12
Introduce SampleSpec lcass

nythoang:
Pull request for issue #378 

Hi Victor, I checked if we could remove `packet::num_channels()` completely and put it into `SampleSpec`. I saw that in some cases, `num_channels()` is being used only with channel mask (e.g. test_pcm_funcs.cpp:79)
-  Should we temporarily construct `SampleSpec` with default sample rate and channel mask there to get `num_channels()`?
- If we can move `num_channels()` into `SampleSpec`, should we move `packet::channel_mask_t` into `sample_spec.h` and rename the usages into `audio::sample_spec_t`?

Let me know so I can make the changes accoridngly!


src/modules/roc_audio/sample_spec.h
```diff
@@ -0,0 +1,75 @@
...
+#include "roc_packet/units.h"
+
+namespace roc {
+namespace audio {
+
+//! Default sample rate, number of samples per second.
+const size_t DefaultSampleRate = 44100;
+
+//! Default channel mask.
+const packet::channel_mask_t DefaultChannelMask = 0x3;
```
gavv, line 26: It would be nice to keep these defaults in one place (in roc_pipeline/config.h) and not duplicate them here. Is there something stopping us from removing them from here?

src/modules/roc_pipeline/config.h
```diff
@@ -168,8 +165,8 @@ struct ReceiverSessionConfig {
     //! Target latency, nanoseconds.
     core::nanoseconds_t target_latency;
 
-    //! Channel mask.
-    packet::channel_mask_t channels;
+    //! Sample spec.
+    audio::SampleSpec sample_spec;
```
gavv, line 169: Could you please take a look at this comment? https://github.com/roc-streaming/roc-toolkit/pull/430#discussion_r536713364

It suggest to get rid of ReceiverSessionConfig::sample_spec at all and only use sample_spec from format and common_config. What do you think, can we do it?

src/modules/roc_sndio/target_sox/roc_sndio/sox_backend.h
```diff
@@ -12,12 +12,12 @@
...
 #define ROC_SNDIO_SOX_BACKEND_H_
 
-#include <sox.h>
-
+#include "roc_audio/sample_spec.h"
 #include "roc_core/mutex.h"
 #include "roc_core/noncopyable.h"
 #include "roc_core/singleton.h"
 #include "roc_sndio/ibackend.h"
+#include <sox.h>
```
gavv, line 20: Nitpick: let's move <sox.h> back to its original position.

src/modules/roc_audio/sample_spec.h
```diff
@@ -0,0 +1,75 @@
...
+
+//! Sample spec.
+class SampleSpec {
+public:
+    //! Default constructor.
+    SampleSpec();
+
+    //! Constructor with sample rate and channel mask.
+    SampleSpec(size_t sample_rate, packet::channel_mask_t channel_mask);
```
gavv, line 35: I have a suggestion regarding constructors and getters:

* default ctor will set sample rate and channel mask to zeros (instead of 44100 and 0x3)
* non-default ctor will panic if provided sample rate or channel mask is zero
* getters will panic if returned sample rate or channel mask is zero

This way, we can earlier catch various errors with fully or partially uninitialized sample specs. We can be almost sure that if tests are not crashing, then no component is using an uninitialized field, since usually components access them early in their ctors.

If there are places where a component needs to check whether sample rate or channel mask is already initialized or not, we can add have_sample_rate() and have_channel_mask() methods for them.

gavv (commented):
Thanks for the PR!

I started reviewing it, here are some preliminary comments.

gavv:
> Hi Victor, I checked if we could remove packet::num_channels() completely and put it into SampleSpec. I saw that in some cases, num_channels() is being used only with channel mask (e.g. test_pcm_funcs.cpp:79)

Regarding these usages.

In packet_writer.h and packet_reader.h I think we can replace channel mask with either sample spec or number of channels. It seems we already have both defined in roc_pipeline tests which are using packet writer and reader.

In test_sox_sink.cpp we can make FrameDuration a field instead of a constant and use sample_spec to compute it. We already have sample_spec in setup() method.

It seems the remaining usages are only two: test_frame_encoder_decoder.cpp and test_pcm_funcs.cpp. Both are in roc_audio tests. So I think we can just copy-paste packet::num_channels() to a static function in these files, or maybe place it to src/tests/roc_audio/test_helpers.

This way we can get rid of packet::num_channels() by the cost of duplicating channel calculation logic in SampleSpec and a couple of tests. The benefit is that we can be sure that, except tests, there is only one way to work with sample rate and channel mask - SampleSpec - and all components have to use it. I think it's worth it.

src/modules/roc_sndio/target_sox/roc_sndio/sox_source.cpp
```diff
@@ -38,15 +38,16 @@ SoxSource::SoxSource(core::IAllocator& allocator, const Config& config)
     }
 
     frame_length_ = config.frame_length;
-    channels_ = config.channels;
+    sample_spec_ = audio::SampleSpec(config.sample_spec.sample_rate(),
+                                     config.sample_spec.channel_mask());
```
gavv, line 42: Can't we change it to just `sample_spec_ = config.sample_spec`?

src/modules/roc_sndio/target_sox/roc_sndio/sox_source.cpp
```diff
@@ -321,6 +322,7 @@ bool SoxSource::open_() {
     }
 
     is_file_ = !(input_->handler.flags & SOX_FILE_DEVICE);
+    sample_spec_.set_sample_rate((unsigned long)input_->signal.rate);
```
gavv, line 325: Could you please move error checking below (`if (input_->signal.channels != n_channels_) {`) before these lines?

src/modules/roc_sndio/target_sox/roc_sndio/sox_source.h
```diff
@@ -89,7 +90,7 @@ class SoxSource : public ISource, private core::NonCopyable<> {
     core::Array<sox_sample_t> buffer_;
     size_t buffer_size_;
     core::nanoseconds_t frame_length_;
-    packet::channel_mask_t channels_;
+    audio::SampleSpec sample_spec_;
```
gavv, line 93: We can also remove `n_channels_` and use `sample_spec_.num_channel()` instead.

src/tests/roc_audio/test_depacketizer.cpp
```diff
@@ -25,7 +25,8 @@ namespace audio {
 
 namespace {
 
-enum { MaxBufSize = 4000, SamplesPerPacket = 200, NumCh = 2, ChMask = 0x3 };
+enum { MaxBufSize = 4000, SamplesPerPacket = 200, SampleRate=100, ChMask = 0x3,
+       SamplesSize = 400 };
```
gavv, line 29: `SamplesSize` is `SamplesPerPacket * <number of channels>`, right? If so, I suggest to explicitly define NumCh and define SamplesSize as `SamplesPerPacket * NumCh`, to make it clear. I think it's OK that we'll have both ChMask and NumCh defined, since they will be close to each other.

gavv (commented):
I've finished the review, here are a few more comments.

nythoang:
Hi Victor, I made the changes according to your review. Let me know if I need to change something else as well!

gavv:
@nythoang Sorry for so long delay a thanks a lot for your work! I finally found time to review the PR carefully. BTW, this PR is quite important because it's the first step to implement #86, which is a must-have. So, thanks!

I've found two minor problems in the new pipeline construction code of ReceiverSession. Since you were waiting for review for so long, I'm going to merge this now and prepare a small follow-up commit by myself. I'll fix the formatting issue as well. Will post it here soon.

gavv:
Here are the follow-up commits:

* 0c3d3adcbaed2b5cb3b0b9543db7a633e1bd0f1c
* 763298ee4f2d7f7e2632630d012c7dd8e7f60b11
* 878f7113c2c5498e93df039bc40541b87d291c16

PR #455, opened 2021-07-15
[docs/building/user_cookbook] Fix missing extension for pulse modules

maximeborges:
I've been following the recipe for the RPi2/3, and saw that it was missing the .so extension when copying the pulseaudio modules.
I guess that it's also necessary for the other embedded targets, so I edited those also.


gavv:
Thanks!

PR #462, opened 2021-10-11
Added rpm spec for building a rpm package

mk0x9:
Hi, I had added rpm spec file to build rpm packages based on the master branch on the [copr](https://copr.fedorainfracloud.org/coprs/mk9/roc-toolkit-dev/).

As fedora switched to pipewire already I plan at a later stage push to the copr pipewire built with roc modules (right now fedora doesn't ship them).


mk0x9:
Sure, I'll try to think of the best way to achieve the result, I had never used fedora copr before. I plan also to add script to build an rpm package with roc modules for the pipewire. Although I'm not sure if this repo would be the best place to host it.

gavv:
Hi!

* Is this PR still relevant?
* The URL in README gives 404: https://copr.fedorainfracloud.org/coprs/mk9/roc-toolkit-dev/package/roc-toolkit/
* The spec defines name as `libroc`, but it also provides command-line tools, not just the library?
* Is it possible to get version from git tag or `.h` header instead of hard-coding it? So that I don't need to update it every release.

gavv:
Follow-up issue: https://github.com/roc-streaming/roc-toolkit/issues/526

PR #475, opened 2022-06-28
Added rpm spec compatible with openSUSE 15.4

twojstaryzdomu:
Note an installed libopenfec-devel package is a requirement.

This package builds without any problems on openSUSE 15.3 & 15.4.

_disable_source_fetch is purposefully undefined to allow downloading the source tarball.


gavv:
I've added an RPM spec to the repo, based on #475 (this PR), #462, and the spec from Fedora. See d3ac0d075644b40fa61c67954a611c17dc2da8c4

Let me know if any changes are needed.

Thanks!

gavv:
Follow-up issue: https://github.com/roc-streaming/roc-toolkit/issues/526

Issue #364, opened 2020-05-24
Add CSPRNG based on OpenSSL


gavv:
## Problem

Earlier we implemented CSPRNG using libuv (#102).

Now we're slowly removing libuv dependency. On the other hand, we will likely employ OpenSSL for DTLS support (#318), and OpenSSL provides CSPRNG as well, which we could use.

## Acceptance criteria

The following solution is suggested:

- [x] Added tests for `fast_random`
- [ ] Added tests for `secure_random`.
- [x] Added OpenSSL dependency to SConstruct and build-3rdparty.py. Added a new target directory `target_openssl`. Added a new scons option `--disable-openssl`, which disables OpenSSL dependency and target directory. By default it should be enabled.
- [x] Added `--build-3rdparty=openssl:1.1.1t` to CI (e.g. to ubuntu 14.04)
- [x] Added a new target directory `target_nocsprng`. Enable it if `--disable-openssl` is given.
- [ ] Added `core::secure_random()` implementation based on OpenSSL to `target_openssl`.
- [ ] Added `core::secure_random()` fallback implementation based on `fast_random()` to `target_nocsprng`.
- [x] Added OpenSSL to https://github.com/roc-streaming/dockerfiles.
- [x] Added OpenSSL to https://github.com/roc-streaming/distfiles.
- [x] Updated https://roc-streaming.org/toolkit/docs/building/dependencies.html
- [x] Updated https://roc-streaming.org/toolkit/docs/building/user_cookbook.html

## Info

About target directories: https://roc-streaming.org/toolkit/docs/internals/code_structure.html#targets

An example of adding a new dependency can be found in #246 and #265.

## Implementation

- #483
- #485
- #489
- #490
- #493
- #524

jsomwaru:
You think I can take this?

gavv:
You're welcome!

jsomwaru:
If openssl is enabled should I remove target_libuv as target directory? Is the current target_libuv not supposed to have any conflicts with the OpenSSL target directory and why you specified to create the new tartget_libuv_csprn?

gavv:
Hi, sorry for late reply.

> If openssl is enabled should I remove target_libuv as target directory?

No, we have a lot of other code using libuv except PRNG implementation.

> Is the current target_libuv not supposed to have any conflicts with the OpenSSL target directory and why you specified to create the new tartget_libuv_csprn?

The only conflict will be secure_random().

[...]

PR #483, opened 2022-12-03
#364 random tests

adrianopol:
#364 added some tests for `secure_random()`

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -742,6 +742,11 @@ subenvs = type('subenvs', (), subenv_attrs)
 env, subenvs = env.SConscript('3rdparty/SConscript',
                        duplicate=0, exports='env subenvs meta')
 
+env.Append(CPPDEFINES=[
+    # for UINT32_MAX and others (https://bugzilla.mozilla.org/show_bug.cgi?id=673556):
+    ('__STDC_LIMIT_MACROS', '1'),
+])
+
 if 'target_posix' in env['ROC_TARGETS'] and meta.platform not in ['darwin', 'unix']:
     env.Append(CPPDEFINES=[('_POSIX_C_SOURCE', '200809')])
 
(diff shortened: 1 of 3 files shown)


src/tests/roc_core/test_random.cpp
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2015 Roc Streaming authors
```
gavv, line 2: 2015 -> 2022

src/tests/roc_core/test_random.cpp
```diff
@@ -0,0 +1,47 @@
...
+};
+
+TEST(random, min_min) {
+    uint32_t lo_hi = 0;
+    uint32_t res;
+    bool ok = secure_random(lo_hi, lo_hi, res);
+
+    CHECK(ok);
+    LONGS_EQUAL(res, lo_hi);
+}
```
gavv, line 27: Could you please add the same check for `fast_random` to each test?

gavv (commented):
Thanks! See comments below

gavv:
CI failed:

* code is not formatted
* build failed for one of arm toolchains: https://github.com/roc-streaming/roc-toolkit/actions/runs/3610087937/jobs/6083675329

If you follow the link, you'll see this command at the beginning of the file:

```
scripts/ci_checks/docker.sh rocstreaming/toolchain-arm-bcm2708hardfp-linux-gnueabi:gcc-4.7 scripts/ci_checks/linux-arm/arm-bcm2708hardfp-linux-gnueabi-gcc-4.7.sh
```

You can run this command locally to reproduce that CI step on your machine. See here: https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/development/continuous_integration.rst

src/tests/roc_core/test_random.cpp
adrianopol: done

src/tests/roc_core/test_random.cpp
adrianopol: done

PR #490, opened 2022-12-11
#364 add openssl library

adrianopol:
#364
should be merged after #489

### DIFF
--- a/3rdparty/SConscript
+++ b/3rdparty/SConscript
@@ -312,15 +312,14 @@ if 'openssl' in autobuild_dependencies:
     env.BuildThirdParty(thirdparty_build_dir, meta.toolchain,
         meta.thirdparty_variant, thirdparty_versions, 'openssl')
 elif 'openssl' in system_dependencies:
-    pkg_config_path = None
-    # set custom PKG_CONFIG_PATH in MacOS
+    ossl_prefix = None
+    # macOS: fallback to use a brew-installed openssl (via PKG_CONFIG_PATH)
     if meta.platform == 'darwin' and not is_crosscompiling:
-        pkg_config_path = env.FindBrewLibrary('openssl')
+        ossl_prefix = env.FindBrewPackage('openssl')
 
     conf = Configure(env, custom_tests=env.CustomTests)
 
-    if not conf.AddPkgConfigDependency(
-            'libssl', '--cflags --libs', config_path=pkg_config_path):
+    if not conf.AddPkgConfigDependency('libssl', '--cflags --libs', prefix=ossl_prefix):
         conf.env.AddPkgConfigLibs(['libssl'])
 
     if not conf.CheckLibWithHeaderExt(
--- a/3rdparty/SConscript
+++ b/3rdparty/SConscript
@@ -312,16 +312,22 @@ if 'openssl' in autobuild_dependencies:
     env.BuildThirdParty(thirdparty_build_dir, meta.toolchain,
         meta.thirdparty_variant, thirdparty_versions, 'openssl')
 elif 'openssl' in system_dependencies:
+    pkg_config_path = None
+    # set custom PKG_CONFIG_PATH in MacOS
+    if meta.platform == 'darwin' and not is_crosscompiling:
+        pkg_config_path = env.FindBrewLibrary('openssl')
+
     conf = Configure(env, custom_tests=env.CustomTests)
 
-    if not conf.AddPkgConfigDependency('libssl', '--cflags --libs'):
+    if not conf.AddPkgConfigDependency(
+            'libssl', '--cflags --libs', config_path=pkg_config_path):
         conf.env.AddPkgConfigLibs(['libssl'])
 
     if not conf.CheckLibWithHeaderExt(
             'ssl', 'openssl/rand.h', 'C', run=not is_crosscompiling):
         env.Die("OpenSSL not found (see 'config.log' for details)")
 
-    env= conf.Finish()
+    env = conf.Finish()
 
 # dep: ragel
 if 'ragel' in autobuild_dependencies:
--- a/3rdparty/SConscript
+++ b/3rdparty/SConscript
@@ -319,11 +319,10 @@ elif 'openssl' in system_dependencies:
 
     conf = Configure(env, custom_tests=env.CustomTests)
 
-    if not conf.AddPkgConfigDependency('libssl', '--cflags --libs', prefix=ossl_prefix):
+    if not conf.AddPkgConfigDependency('libssl', '--cflags --libs', add_prefix=ossl_prefix):
         conf.env.AddPkgConfigLibs(['libssl'])
 
-    if not conf.CheckLibWithHeaderExt(
-            'ssl', 'openssl/rand.h', 'C', run=not is_crosscompiling):
+    if not conf.CheckLibWithHeaderExt('ssl', 'openssl/rand.h', 'C', run=not is_crosscompiling):
         env.Die("OpenSSL not found (see 'config.log' for details)")
 
     env = conf.Finish()
--- a/3rdparty/SConscript
+++ b/3rdparty/SConscript
@@ -12,6 +12,7 @@ thirdparty_versions = {
     'libuv':            '1.35.0',
     'ltdl':             '2.4.6',
     'openfec':          '1.4.2.4',
+    'openssl':          '3.0.7',
     'pulseaudio':       '12.2',
     'sndfile':          '1.0.28',
     'sox':              '14.4.2',
@@ -305,6 +306,22 @@ elif 'sox' in system_dependencies:
             env.Die("libsox not found (see 'config.log' for details)")
 
     env = conf.Finish()
+
+# dep: openssl
+if 'openssl' in autobuild_dependencies:
+    env.BuildThirdParty(thirdparty_build_dir, meta.toolchain,
+        meta.thirdparty_variant, thirdparty_versions, 'openssl')
+elif 'openssl' in system_dependencies:
+    conf = Configure(env, custom_tests=env.CustomTests)
+
+    if not conf.AddPkgConfigDependency('libssl', '--cflags --libs'):
+        conf.env.AddPkgConfigLibs(['libssl'])
+
+    if not conf.CheckLibWithHeaderExt(
+            'ssl', 'openssl/rand.h', 'C', run=not is_crosscompiling):
+        env.Die("OpenSSL not found (see 'config.log' for details)")
+
+    env= conf.Finish()
 
 # dep: ragel
 if 'ragel' in autobuild_dependencies:
--- a/SConstruct
+++ b/SConstruct
@@ -191,6 +191,11 @@ AddOption('--disable-sox',
           action='store_true',
           help='disable SoX support in tools')
 
+AddOption('--disable-openssl',
+          dest='disable_openssl',
+          action='store_true',
+          help='disable OpenSSL support required for DTLS and SRTP')
+
 AddOption('--disable-libunwind',
           dest='disable_libunwind',
           action='store_true',
@@ -720,6 +725,11 @@ else:
     if not GetOption('disable_openfec'):
         env.Append(ROC_TARGETS=[
             'target_openfec',
+        ])
+
+    if not GetOption('disable_openssl'):
+        env.Append(ROC_TARGETS=[
+            'target_openssl',
         ])
 
     if not GetOption('disable_speexdsp'):
--- a/scripts/ci_checks/macos/macos.sh
+++ b/scripts/ci_checks/macos/macos.sh
@@ -7,10 +7,6 @@ brew install \
      libuv speexdsp sox openssl@3 \
      cpputest google-benchmark
 
-# > openssl@3 is keg-only, which means it was not symlinked into /usr/local,
-# > because macOS provides LibreSSL.
-export PKG_CONFIG_PATH="/usr/local/opt/openssl@3/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
-
 scons -Q \
       --enable-werror \
       --enable-debug \
--- a/scripts/ci_checks/macos/macos.sh
+++ b/scripts/ci_checks/macos/macos.sh
@@ -3,9 +3,13 @@
 set -euxo pipefail
 
 brew install \
-     automake scons ragel gengetopt\
-     libuv speexdsp sox \
+     automake scons ragel gengetopt \
+     libuv speexdsp sox openssl@3 \
      cpputest google-benchmark
+
+# > openssl@3 is keg-only, which means it was not symlinked into /usr/local,
+# > because macOS provides LibreSSL.
+export PKG_CONFIG_PATH="/usr/local/opt/openssl@3/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
 
 scons -Q \
       --enable-werror \
--- a/scripts/scons_helpers/build-3rdparty.py
+++ b/scripts/scons_helpers/build-3rdparty.py
@@ -679,6 +679,8 @@ def openssl_get_platform(toolchain):
     'linux-generic32'
     >>> openssl_get_platform('arm-bcm2708hardfp-linux-gnueabi')
     'linux-generic32'
+    >>> openssl_get_platform('armv7a-linux-androideabi33')
+    'android-arm'
     >>> openssl_get_platform('mips64-linux-musl')
     'linux64-mips64'
     >>> openssl_get_platform('arm-linux-androideabi')
@@ -695,10 +697,10 @@ def openssl_get_platform(toolchain):
     'linux-x86'
 
     # other edge cases
-    >>> openssl_get_platform('mingw')
-    'mingw'
-    >>> openssl_get_platform('cygwin')
-    'Cygwin'
+    >>> openssl_get_platform('mingw-w64-x86_64')
+    'mingw64'
+    >>> openssl_get_platform('x86_64-pc-cygwin')
+    'Cygwin-x86_64'
     >>> openssl_get_platform('')
     'cc'
     """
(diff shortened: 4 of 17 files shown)


3rdparty/SConscript
```diff
@@ -12,6 +12,7 @@ thirdparty_versions = {
     'libuv':            '1.35.0',
     'ltdl':             '2.4.6',
     'openfec':          '1.4.2.4',
+    'openssl':          '3.0.7', # minimal supported version is 1.1.1
```
gavv, line 15: We maintain the list of minimal versions [here](https://roc-streaming.org/toolkit/docs/building/dependencies.html), let's remove this comment and instead add openssl to that page.

SConstruct
```diff
@@ -704,6 +704,7 @@ else:
 
     env.Append(ROC_TARGETS=[
         'target_libuv',
+        'target_openssl',
```
gavv, line 707: We should add new option `--disable-openssl` and add this target only if the option is not set.

scripts/ci_checks/macos/macos.sh
```diff
@@ -3,10 +3,14 @@
...
 brew install \
-     automake scons ragel gengetopt\
-     libuv speexdsp sox \
+     automake scons ragel gengetopt \
+     libuv speexdsp sox openssl@3 \
      cpputest google-benchmark
 
+# > openssl@3 is keg-only, which means it was not symlinked into /usr/local,
+# > because macOS provides LibreSSL.
+export PKG_CONFIG_PATH="/usr/local/opt/openssl@3/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
```
gavv, line 12: As discussed in char, we should remove this and adjust build system to work with openssl from brew out of the box.

In `3rdparty/SConscript`, we should check that:

* we're running on macos and are not cross-compiling
* `brew` is in PATH (see env.Which)
* `brew --prefix` returns existing directory
* openssl is installed at `<PREFIX>/Cellar/openssl@<SLOT>/<VERSION>`

Where:

* PREFIX is output of `brew --prefix`
* SLOT is short version, e.g. `1.1` or `3`
* VERSION is full version, e.g. `1.1.1s`

I think we should not hard-code SLOT and VERSION, but instead see what is present and use the latest one.

If all these check succeeded, we can append `<PREFIX>/Cellar/openssl@<SLOT>/<VERSION>/lib/pkgconfig` to `PKG_CONFIG_PATH` and try to find openssl.

What we'll have in result:

* if user did not specify any option, we'll automatically use openssl 1.1 or 3 from brew
* user still can install openssl manually and set PKG_CONFIG_PATH to use it

Here is some output on one of my machines.

```
$ brew --prefix
/usr/local
```

```
$ tree -L 2 /usr/local/Cellar/openssl@1.1/1.1.1s
/usr/local/Cellar/openssl@1.1/1.1.1s
├── AUTHORS
├── CHANGES
├── INSTALL_RECEIPT.json
├── LICENSE
├── NEWS
├── README
├── bin
│   ├── c_rehash
│   └── openssl
├── include
│   └── openssl
├── lib
│   ├── engines-1.1
│   ├── libcrypto.1.1.dylib
│   ├── libcrypto.a
│   ├── libcrypto.dylib -> libcrypto.1.1.dylib
│   ├── libssl.1.1.dylib
│   ├── libssl.a
│   ├── libssl.dylib -> libssl.1.1.dylib
│   └── pkgconfig
└── share
    ├── doc
    └── man
```

```
$ ls -1 /usr/local/Cellar/openssl@1.1/1.1.1s/lib/pkgconfig
libcrypto.pc
libssl.pc
openssl.pc
```

gavv (commented):
Thanks for PR!!

scripts/ci_checks/macos/macos.sh
adrianopol: agreed to ignore SLOT and VERSION and instead use `$(bre --prefix)/opt/openssl` which is a link to the particular version, e.g. `$(brew --prefix)/opt/openssl@1.1/1.1.1s/`

3rdparty/SConscript
adrianopol: done

SConstruct
adrianopol: done

scripts/ci_checks/macos/macos.sh
adrianopol: done

3rdparty/SConscript
```diff
@@ -306,6 +307,28 @@ elif 'sox' in system_dependencies:
...
 
+# dep: openssl
+if 'openssl' in autobuild_dependencies:
+    env.BuildThirdParty(thirdparty_build_dir, meta.toolchain,
+        meta.thirdparty_variant, thirdparty_versions, 'openssl')
+elif 'openssl' in system_dependencies:
+    pkg_config_path = None
+    # set custom PKG_CONFIG_PATH in MacOS
+    if meta.platform == 'darwin' and not is_crosscompiling:
+        pkg_config_path = env.FindBrewLibrary('openssl')
```
gavv, line 318: A few minor suggestions:

1. Let's make FindBrewLibrary to be method of config, and invoke conf.FindBrewLibrary instead of env.FindBrewLibrary. The implementation will be moved to config.py in scons plugin. It collects methods for finding system libs, dirs, etc.

2. Let's rename it to FindBrewPackage.

3. Let's make FindBrewPackage a bit more generic: it will return package path, instead of pkgconfig path instead of the package, i.e. it will exclude `'/lib/pkgconfig` from returned path.

4. Accordingly, let's AddPkgConfigDependency handle this more generic version seamlessly. Instead of config_path argument, it will have an argument like prepend_prefix or additional_prefix. AddPkgConfigDependency will treat it as additional installation prefix that should be searched for `/lib/pkgconfig` subdir.

With this changes, FindBrewPackage won't be limited to only libs and to only packages that have pkgconfig files.

3rdparty/SConscript
```diff
@@ -306,6 +307,28 @@ elif 'sox' in system_dependencies:
 
     env = conf.Finish()
 
+# dep: openssl
+if 'openssl' in autobuild_dependencies:
+    env.BuildThirdParty(thirdparty_build_dir, meta.toolchain,
+        meta.thirdparty_variant, thirdparty_versions, 'openssl')
+elif 'openssl' in system_dependencies:
+    pkg_config_path = None
+    # set custom PKG_CONFIG_PATH in MacOS
```
gavv, line 316: nit: MacOS => macOS

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,164 @@ def splitenv(envlist):
...
 
+# Guess platform argument for OpenSSL's Configure script basing on toolchain string.
+# (see `./Configure LIST` for a list of all platforms and their format)
+# See also openssl/Configurations/10-main.conf
+def openssl_get_platform(toolchain):
+    """
+    >>> openssl_get_platform('x86_64-pc-linux-gnu')
+    'linux-x86_64'
+    >>> openssl_get_platform('aarch64-linux-gnu')
+    'linux-aarch64'
```
gavv, line 675: Shouldn't it be

```
linux64-x86_64
linux64-aarch64
```

?

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,164 @@ def splitenv(envlist):
...
+
+    # other edge cases
+    >>> openssl_get_platform('android64')
+    'android64'
+    >>> openssl_get_platform('')
+    'cc'
+    """
+
+    # we assume that:
+    # - 'android' has more priority than 'linux'
```
gavv, line 703: nit: more => higher

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,164 @@ def splitenv(envlist):
...
+        replace('bcm2708hardfp-linux', 'linux').\
+        replace('openwrt-linux', 'linux').\
+        replace('pc-linux', 'linux')
+    # normalize arch: no suffix means 32 bit
+    toolchain = toolchain.replace('mips-', 'mips32-')
+
+    ### corner cases
+
+    # 64 bit => explicitly prefer linux64 or android64 (concider libc and other libs
+    # to be 64 bit)
```
gavv, line 720: nit: concider => consider

but better use "assume"

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,164 @@ def splitenv(envlist):
...
+    if bitness == 64:
+        if 'linux-' in toolchain:
```
gavv, line 722: I suggest to replace `linux-` with `linux`. I can't imagine false positives for this, and it looks a safer heuristic.

scripts/scons_plugin/config.py
```diff
@@ -457,7 +457,10 @@ def AddPkgConfigDependency(context, package, flags):
         context.Result('pkg-config not available')
         return False
 
-    cmd = '%s %s --silence-errors %s' % (quote(pkg_config), package, flags)
+    cmd = ''
+    if config_path and os.path.isdir(config_path):
+        cmd = 'env PKG_CONFIG_PATH={} '.format(quote(config_path))
+    cmd += '{} {} --silence-errors {}'.format(quote(pkg_config), package, flags)
```
gavv, line 463: nit: we mostly use %s formatting in scons, could you use it here too for consistency?

scripts/scons_plugin/pkgconfig.py
```diff
@@ -45,6 +45,32 @@ def write_file(target, source, env):
...
+# library installed by brew.
+#
+# When `required` = True error is raised unless the search was successful. Otherwise in
+# case of any error None is returned.
+def FindBrewLibrary(env, lib_name, required=True):
+    if not env.Which('brew'):
+        if required:
+            env.Die('brew not found')
+        else:
+            return None
```
gavv, line 58: We should not fail if the user does not have brew. Using brew is just a convenient shortcut, but user is free to install openssl by any other means.

We can pass required=False, but I think we can just remove required flag at all, for now.

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,164 @@ def splitenv(envlist):
...
+
+    bitness = 64 if '64' in toolchain else 32
+
+    # normalize os: make toolchain to have at most 3 components
+    toolchain = toolchain.\
+        replace('bcm2708hardfp-linux', 'linux').\
+        replace('openwrt-linux', 'linux').\
+        replace('pc-linux', 'linux')
+    # normalize arch: no suffix means 32 bit
+    toolchain = toolchain.replace('mips-', 'mips32-')
```
gavv, line 715: The presence of this block indicates that heuristics in this function are not generic enough.

Ideally, the function should work without *knowing* if these four specific cases; these strings should be present only in tests.

* `bcm2708` is the name of random chip; we can't know and list names of all chips, and vendors can include chip names in toolchain;
* `openwrt` is the name of random distro; again, we can't list names of all of them;
* `pc-linux` is rather generic, but there are some variations, e.g. on some systems it's `uknown-linux`
* instead of `mips-`, it can be `mipsle-`, for example

It seems that all these problems are caused by relying on presence of specific tokens in specific places, like `linux` or `mips` before hyphen.

It seems all of them can be easily worked around by switching to a more relaxed matching. We can just check presence of `linux` and `mips` in toolchain string, and check presence or absence of other keywords like `64` or `armeabi`.

The lesser we assume about toolchain structure, the better...

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,164 @@ def splitenv(envlist):
...
+
+    ### corner cases
+
+    # 64 bit => explicitly prefer linux64 or android64 (concider libc and other libs
+    # to be 64 bit)
+    if bitness == 64:
+        if 'linux-' in toolchain:
+            toolchain = re.sub(r'\blinux\b', 'linux64', toolchain)
+        if 'android' in toolchain and 'android64' not in toolchain:
+            toolchain = re.sub(r'\bandroid\b', 'android64', toolchain)
```
gavv, line 725: As I mentioned above, relying on word boundaries here (`\b`) looks not enough generic

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,164 @@ def splitenv(envlist):
...
+                if len(ac_arr) > 1:
+                    compiler = ac_arr[1]
+                    if compiler in toolchain:
+                        return '-'.join([os, arch, compiler])
+                #~print('2: ', os, arch)
+                return '-'.join([os, arch])
+            if os in ['mingw', 'mingw64', 'Cygwin', 'UEFI', 'UWIN', 'android64']:
+                return os
+    if 'gcc' in toolchain:
+        return 'gcc'
```
gavv, line 794: I think this can never happen. CMIIAW.

But you can use `checkfamily(env, toolchain, 'gcc')`.

3rdparty/SConscript
adrianopol:
```
[[ paq@iota ~/work/roc/roc-toolkit ]] rg MacOS | wc -l
118
[[ paq@iota ~/work/roc/roc-toolkit ]] rg macOS | wc -l
101
```

scripts/scons_helpers/build-3rdparty.py
adrianopol: no, there are no such targets

scripts/scons_helpers/build-3rdparty.py
adrianopol: fixed

scripts/scons_helpers/build-3rdparty.py
adrianopol: fixed

3rdparty/SConscript
adrianopol: agreed: `macOS`

scripts/scons_plugin/pkgconfig.py
adrianopol: TODO: if brew is installed and contains openssl then append path to openssl's pkgconfig path to `env['PKG_CONFIG_PATH']`.

3rdparty/SConscript
adrianopol: done

3rdparty/SConscript
adrianopol: fixed

scripts/scons_plugin/config.py
adrianopol: `format()` is always better and safer; see
- https://docs.python.org/2/library/string.html#format-examples
- https://stackoverflow.com/a/5082482

scripts/scons_plugin/pkgconfig.py
adrianopol: fixed

scripts/scons_helpers/build-3rdparty.py
adrianopol: fair enough; rewritten most of this

scripts/scons_helpers/build-3rdparty.py
adrianopol: done

3rdparty/SConscript
```diff
@@ -306,6 +307,27 @@ elif 'sox' in system_dependencies:
...
+elif 'openssl' in system_dependencies:
+    ossl_prefix = None
+    # macOS: fallback to use a brew-installed openssl (via PKG_CONFIG_PATH)
+    if meta.platform == 'darwin' and not is_crosscompiling:
+        ossl_prefix = env.FindBrewPackage('openssl')
+
+    conf = Configure(env, custom_tests=env.CustomTests)
+
+    if not conf.AddPkgConfigDependency('libssl', '--cflags --libs', prefix=ossl_prefix):
+        conf.env.AddPkgConfigLibs(['libssl'])
```
gavv, line 323: nit: I suggest to rename `prefix` arg to something like `append_prefix` or `additional_prefix`, to make it clear that this only adds prefix to the search path, not replaces path.

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,152 @@ def splitenv(envlist):
...
+    >>> openssl_get_platform('x86_64-linux-android')
+    'android-x86_64'
+    >>> openssl_get_platform('i686-pc-linux-gnu')
+    'linux-x86'
+
+    # other edge cases
+    >>> openssl_get_platform('mingw')
+    'mingw'
+    >>> openssl_get_platform('cygwin')
+    'Cygwin'
```
gavv, line 701: nit: let's replace these inputs in test with real examples:

```
mingw-w64-x86_64
x86_64-pc-cygwin
```

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,152 @@ def splitenv(envlist):
...
+    >>> openssl_get_platform('arm-linux-androideabi')
+    'android-arm'
+    >>> openssl_get_platform('aarch64-linux-android')
+    'android-arm64'
+    >>> openssl_get_platform('aarch64-linux-android29')
+    'android-arm64'
+    >>> openssl_get_platform('i686-linux-android')
+    'android-x86'
+    >>> openssl_get_platform('x86_64-linux-android')
+    'android-x86_64'
```
gavv, line 693: nit: let's add one more test: `armv7a-linux-androideabi33` => `android-arm`

to be sure that `armv7a` works same as `arm`

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,152 @@ def splitenv(envlist):
...
+        ('mingw64',      []),
+        ('mingw',        []),
+    ]
+    arch_aliases = {
+        'armv4':     ['arm32', 'arm'],
+        'generic32': ['generic'],
+        'mips32':    ['mips'],
+        'x86':       ['i386', 'i486', 'i586', 'i686'],
+        'arm64':     ['aarch64'], # fix for absent platform dirs in android NDK
+    }
```
gavv, line 763: 1. IIRC, we should never pass `linux-armv4` to OpenSSL, right? Because compilation will fail. Instead, we should use `linux-generic32`, which we handle above. In this case, I guess we should **remove `armv4`** from `platforms` and `arch_aliases`.

2. Since in OpenSSL, `android-armeabi` is an alias for `android-arm` (you mentioned it in chat), then I suggest to **remove `armeabi`** from `android` platform.

3. Also maybe **remove** commented out **`android64`** block? Since `android64` has no real value.

4. I think we should **remove `generic64` and `generic32`** from `BSD` and `linux`, and from `arch_aliases`. They will never occur in toolchain string.

gavv: I suggest to use `generic` as a fallback for BSD and linux OSes.

For example, if we detected that `os` is `linux`, and can't detect `arch`, then we should use `linux-generic32` or `linux-generic64` depending on bitness.

This way, this special check:

```
    if 'linux' in toolchain and 'arm' in toolchain and bitness == 32:
        return 'linux-generic32'
```

will be naturally not needed, because we will automatically employ `linux-generic32`.

gavv: It seems that this special check:

```
    if 'android' in toolchain and 'arm' in toolchain and 'eabi' in toolchain:
        return 'android-arm'
```

is not needed now. If `android` and `arm` are in toolchain, we will use `android-arm`, regardless of `eabi`, which is correct.

scripts/scons_plugin/config.py
```diff
@@ -457,7 +457,19 @@ def AddPkgConfigDependency(context, package, flags):
...
+        pkg_config_path = os.path.join(prefix, 'lib', 'pkgconfig')
+        global_pkg_config_path = env.get('PKG_CONFIG_PATH', None)
+
+        # set user-defined $(prefix)/lib/pkgconfig path for pkg-config if needed (lowest
+        # priority); otherwise PKG_CONFIG_PATH is propagated implicitly
+        if os.path.isdir(pkg_config_path):
+            if global_pkg_config_path:
+                pkg_config_path = global_pkg_config_path + ':' + pkg_config_path
+            cmd = 'env PKG_CONFIG_PATH={} '.format(quote(pkg_config_path))
+    cmd += '{} {} --silence-errors {}'.format(quote(pkg_config), package, flags)
```
gavv, line 472: I just realized that we had a bug (before this PR): we did not propagate PKG_CONFIG_PATH to env.

If user set PKG_CONFIG_PATH through env var, it's not needed, but if user did it via scons arguments, like this:

```
scons PKG_CONFIG_PATH=/path
```

then we wont pass it to pkg-config, but should.

Could you rework this block to also pass PKG_CONFIG_PATH to cmd when `prefix` is not set, but `global_pkg_config_path` is set?

scripts/scons_plugin/osutils.py
```diff
@@ -116,10 +116,24 @@ def _which(env, prog, mode, searchpath):
...
+# Find and return path to the package (its install prefix directory) of the latest version
+# of `pkg_name` package installed by brew. If not found, None is returned.
+def FindBrewPackage(env, pkg_name, required=True):
+    if not env.Which('brew'):
+        return None
+
+    # $(brew --prefix)/opt/some_pkg should be a link to the latest version of some_pkg,
+    # e.g. ../Cellar/some_pkg@1.2/1.2.3/
+    prefix = os.path.join(env.GetCommandOutput('brew --prefix'), 'opt', pkg_name)
+    if not os.path.isdir(prefix):
```
gavv, line 128: `GetCommandOutput` can return none (if brew fails, e.g. problem with permissions). We should handle it, otherwise os.path.join will raise exception.

3rdparty/SConscript
adrianopol: done

scripts/scons_helpers/build-3rdparty.py
adrianopol: done

scripts/scons_helpers/build-3rdparty.py
adrianopol: done

scripts/scons_helpers/build-3rdparty.py
adrianopol: Yes, there are problems with armv4. On Android, only the following 4 archs are acceptable (see component after `arch-` string): `arm`, `arm64`, `x86`, and `x86_64`:
````
# docker run --rm -ti --cap-add SYS_PTRACE -u 0 -v .../roc/roc-toolkit:/opt/roc -w /opt/roc -e CI= rocstreaming/toolchain-linux-android:ndk21 /bin/bash
root@eb758900a4b5:/opt/roc# ls -l /opt/android-ndk-r21/platforms/android-29/     
total 16
drwxr-xr-x 3 root root 4096 Jan  6  2020 arch-arm
drwxr-xr-x 3 root root 4096 Jan  6  2020 arch-arm64
drwxr-xr-x 3 root root 4096 Jan  6  2020 arch-x86
drwxr-xr-x 3 root root 4096 Jan  6  2020 arch-x86_64
root@eb758900a4b5:/opt/roc# 
```
Removed `armv4` and others.

adrianopol: Special check
```
if 'android' in toolchain and 'arm' in toolchain and 'eabi' in toolchain:
```
is still required.

adrianopol: > I suggest to use generic as a fallback for BSD and linux OSes.

done

scripts/scons_plugin/config.py
adrianopol: done

scripts/scons_plugin/osutils.py
adrianopol: done

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -664,18 +664,152 @@ def splitenv(envlist):
...
+        ('mingw64',  []),
+        ('mingw',    []),
+    ]
+    arch_aliases = {
+        'arm':       ['armv4', 'armv7a'], # just in case
+        'arm64':     ['aarch64'], # fix for absent platform dirs in android NDK
+        'generic32': ['generic'],
+        'mips32':    ['mips'],
+        'x86':       ['i386', 'i486', 'i586', 'i686'],
+    }
```
gavv, line 760: Let's remove `armv4` and `armv7a` - `arm` is subset of both of them, so should work without alias.

Let's also remove `generic` - it never appears in toolchain.

Let's also remove comment `# fix for absent platform dirs in android NDK` - android is not the only place where aarch64 is used instead of arm64, this name is widely used.

I checked, after removing all them tests still pass.

scripts/scons_plugin/config.py
```diff
@@ -52,6 +56,38 @@ def linux_dirs():
...
+def _compose_pkg_config_path(env, add_prefix, is_doc_test=False):
+    """
+    >>> _compose_pkg_config_path({'PKG_CONFIG_PATH': '/a/b'}, '/usr/local/qqq', True)
+    '/a/b:/usr/local/qqq/lib/pkgconfig'
+    >>> _compose_pkg_config_path({}, '/usr/local/qqq', True)
+    '/usr/local/qqq/lib/pkgconfig'
+    >>> _compose_pkg_config_path({'PKG_CONFIG_PATH': '/a/b/'}, None, True)
+    '/a/b/'
+    >>> _compose_pkg_config_path({}, None, True)
+    ''
```
gavv, line 70: I suggest removing these tests to avoid maintaining two versions of code (test vs non-test). Discussion in chat.

scripts/scons_plugin/config.py
```diff
@@ -445,8 +481,8 @@ def select_path():
     context.Result(env['PKG_CONFIG_PATH'])
     return True
 
-def AddPkgConfigDependency(context, package, flags):
-    context.Message("Searching pkg-config package %s..." % package)
+def AddPkgConfigDependency(context, package, flags, add_prefix=None):
```
gavv, line 484: From chat:

```python
def AddPkgConfigDependency(context, package, flags, add_prefix=None):
    context.Message("Searching pkg-config package {} ...".format(package))

    env = context.env
    if 'PKG_CONFIG_DEPS' not in env.Dictionary():
        env['PKG_CONFIG_DEPS'] = []

    pkg_config = env.get('PKG_CONFIG', None)
    if not pkg_config:
        context.Result('pkg-config not available')
        return False

    cmd = []
    pkg_config_path = _compose_pkg_config_path(env, add_prefix)
    if pkg_config_path:
        cmd = ['env', 'PKG_CONFIG_PATH='+pkg_config_path]

    cmd += [pkg_config, package, '--silence-errors'] + flags.split()
    try:
        env.ParseConfig(cmd)
        env.AppendUnique(PKG_CONFIG_DEPS=[package])
    except:
        context.Result('not found')
        return False

    context.Result('yes')
    return True
```

gavv:
Thanks!!!

PR #492, opened 2022-12-12
Error handling for invalid compiler version

june34876:
Disabling return none for ParseCompilerVersion function in parsers.py


gavv:
1. We're usually not throwing exceptions from scons helpers, returning None is a common way to report failure.
2. Just throwing exception without handling it makes no sense. As you can see, this patch broke many CI steps.
3. Not being able to determine compiler version is a possible situation. We should not fail, we should just handle it correctly in SConstruct.

PR #510, opened 2023-02-22
Fix work without pkg-config on mac

gavv:
This PR fixes OpenSSL search on macos when pkg-config is not installed, which happens randomly on some (but not all) github action workers.

Without this PR, OpenSSL can be used without pkg-config on linux, but not on mac, because on mac OpenSSL is inside brew prefix.

Changes:

* rename pkgconfig.py to systemdeps.py
* move FindBrewPackage to systemdeps.py
* replace AddPkgConfigLibs with AddManualDependency, which can handle not only libs, but also custom prefix
* use AddManualDependency with ossl_prefix if OpenSSL was not found using pkg-config


scripts/scons_plugin/systemdeps.py
```diff
@@ -0,0 +1,100 @@
...
+
+Name: %(name)s
+Requires: %(deps)s
+Version: %(version)s
+Description: %(desc)s
+URL: %(url)s
+
+Libs: %(libdirs)s %(libs)s
+Cflags: %(incdirs)s
+""" % { 'prefix': prefix,
```
adrianopol, line 31: May I ask you to use `format()` instead of `%` operator in all new functions please? I almost broke my head trying to understand what does `s` suffix mean.

> format(*args, **kwargs)
> This method of string formatting is the new standard in Python 3, and should be preferred to the % formatting described in String Formatting Operations in new code.

https://docs.python.org/2/library/stdtypes.html#str.format

```
Python 2.7.18.6 (default, Jan  1 1980, 00:00:00) 
[GCC 11.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> """ xxx={xxx} """.format(xxx=123)
' xxx=123 '

Python 3.10.9 (main, Dec  6 2022, 18:44:57) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> """ xxx={xxx} """.format(xxx=123)
' xxx=123 '
```

scripts/scons_plugin/systemdeps.py
```diff
@@ -0,0 +1,100 @@
...
+        'name': name,
+        'desc': desc,
+        'url': url,
+        'version': version,
+        'deps': ' '.join(dep_list),
+        'libdirs': ' '.join(['-L'+s for s in libdir_list]),
+        'incdirs': ' '.join(['-I'+s for s in incdir_list]),
+        'libs': ' '.join(['-l'+s for s in lib_list]),
```
adrianopol, line 39: are these strings already escaped?

scripts/scons_plugin/systemdeps.py
```diff
@@ -0,0 +1,100 @@
...
+# those subdirs are added to the headers and library search paths.
+# This function also automatically informs GeneratePkgConfig that correspondings
+# paths and libs should be added to .pc file.
+def AddManualDependency(env, libs=[], prefix=None):
+    # append values to env[<name>] and env[PKG_CONFIG_<name>]
+    def append_var(name, values):
+        for var in [name, 'PKG_CONFIG_'+name]:
+            if var not in env.Dictionary():
+                env[var] = []
+            env.AppendUnique(**{var: values})
```
adrianopol, line 66: is it different from `env.AppendUnique(var=values)`?

scripts/scons_plugin/systemdeps.py
```diff
@@ -0,0 +1,100 @@
+import os
+
+def GeneratePkgConfig(env, build_dir, filename, prefix, libdir, name, desc, url, version):
+    target = os.path.join(build_dir, filename)
+
+    dep_list = env.get('PKG_CONFIG_DEPS', [])
+
+    lib_list = \
+      ['roc'] + \
+      env.get('PKG_CONFIG_LIBS', [])
```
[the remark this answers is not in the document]
adrianopol: So, if someone calls `GeneratePkgConfig()` before/without calling `AddManualDependency()`, it won't work as expected. If this is a desired behaviour, then OK. But I would at least write about this dependency in the function's comment.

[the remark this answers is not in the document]
gavv: I've renamed the vars to:

* _DEPS_PCFILES
* _DEPS_LIBS
* _DEPS_LIBPATH
* _DEPS_CPPPATH

to make it clear that they hold info about our dependencies. Leading underscore indicates that this is not standard scons stuff.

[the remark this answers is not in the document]
gavv: > So, if someone calls GeneratePkgConfig() before/without calling AddManualDependency(), it won't work as expected.

Yep.

> If this is a desired behaviour, then OK. But I would at least write about this dependency in the function's comment.

I'll add a comment.

scripts/scons_plugin/systemdeps.py
gavv: Done.

scripts/scons_plugin/systemdeps.py
gavv: No, I forgot to escape them. Fixed.

scripts/scons_plugin/systemdeps.py
gavv: In

```
env.AppendUnique(var=values)
```

variable name is string `"var"`

And in

```
env.AppendUnique(**{var: values})
```

variable name is value of variable var.

In this regard, AppendUnique works similar to format().

scripts/scons_plugin/systemdeps.py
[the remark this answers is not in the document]
gavv: Added comment.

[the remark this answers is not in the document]
gavv: I also removed loop `for var in [name, 'PKG_CONFIG_'+name]` and refactored it to two calls:

```
        append_var('LIBS', libs)
        append_var('_DEPS_LIBS', libs)
```

This way variable names are greppable.

adrianopol (approved):

scripts/scons_plugin/systemdeps.py
```diff
@@ -0,0 +1,116 @@
...
+
+Name: {name}
+Description: {desc}
+URL: {url}
+Version: {version}
+Requires: {deps}
+
+Libs: {libdirs} {libs}
+Cflags: {incdirs}
+""".lstrip().format(
```
adrianopol, line 45: FYI: https://docs.python.org/2.7/library/textwrap.html#textwrap.dedent

gavv: Great idea, applied

PR #514, opened 2023-03-06
Remove unnecessary OpenSSL options

gavv:
I consulted OpenSSL docs, and it seems that some of the options we pass only enable builtin features, without adding any external dependencies. So they do not harm and it's better to enable them to make sure we don't miss any important functionality.

I kept disabled tests and similar stuff.

The initial motivation for doing this was attempt to use `--build-3rdparty=openssl:1.1.1t`. It failed to build because some of the options are not supported by older OpenSSL. Now it works.


adrianopol (approved):

adrianopol:
LGTM

PR #517, opened 2023-03-06
GH-512 python: % -> format(); other small refactoring (part 2)

adrianopol:

PR #534, opened 2023-04-30
Update help2rst.py

Sanjipan:
I have changed normal print to .format() prints.


gavv:
This change is already covered by #517.

Also, PRs should be targeted to develop branch, see [here](https://roc-streaming.org/toolkit/docs/development/version_control.html).

PR #536, opened 2023-04-30
Update device_type.cpp

Sanjipan:


gavv:
https://roc-streaming.org/toolkit/docs/development/coding_guidelines.html

PR #544, opened 2023-08-27
Capture timestamps in frames and rtp packets

baranovmv:
Add unit timestamp fileds in nanoseconds to audio::Frame and  packet::RTP.

Following components were modified to forward timestamps:

-    ChannelMapperReader
-    ChannelMapperWriter
-    Depacketizer
-    Packetizer
-    ResamplerReader
-    ResamplerWriter

Added `CaptureTsFiller` class (packet reader) that gets a pair of a unix timestamp and a correspondent rtp timestamp and fills  `capture_timestamp` field in rtp packet.


src/internal_modules/roc_audio/depacketizer.cpp
```diff
@@ -107,21 +112,32 @@ Depacketizer::read_samples_(sample_t* buff_ptr, sample_t* buff_end, FrameInfo& i
...
         if (buff_ptr < buff_end) {
             sample_t* new_buff_ptr = read_packet_samples_(buff_ptr, buff_end);
+            const size_t nsamples = size_t(new_buff_ptr - buff_ptr);
 
-            info.n_decoded_samples += size_t(new_buff_ptr - buff_ptr);
-
+            info.n_decoded_samples += nsamples;
+            if (nsamples && !info.ts) {
+                info.ts = next_capture_ts_;
+            }
```
gavv, line 131: nit: For clarity, I'd change it to:

```
if (nsamples && !info.ts && valid_capture_ts_) {
```

src/internal_modules/roc_audio/depacketizer.cpp
```diff
@@ -107,21 +112,32 @@ Depacketizer::read_samples_(sample_t* buff_ptr, sample_t* buff_end, FrameInfo& i
...
+            next_capture_ts_ += sample_spec_.samples_overall_2_ns(nsamples);
```
gavv, line 132: I guess we should do it only if valid_capture_ts_ is true.

BTW I think we need a test for depacketizer when timestamps in packets are zero for a while and then become non-zero, to catch this gotchas.

src/internal_modules/roc_audio/depacketizer.cpp
```diff
@@ -207,6 +223,9 @@ void Depacketizer::update_packet_(FrameInfo& info) {
         return;
     }
 
+    next_capture_ts_ = packet_->rtp()->capture_timestamp;
+    valid_capture_ts_ = !!next_capture_ts_;
```
gavv, line 227: I think if valid_capture_ts_ became true, it should never become false again, otherwise it will break logic above, what do you think?

gavv: Also: we have rtp::Validator, that terminates session if there are unexpected seqnum or timestamp jumps. I suggest to modify it to detect capture_ts jumps as well. It will handle the case when timestamp suddenly became zero after being non-zero.

src/internal_modules/roc_audio/depacketizer.h
```diff
@@ -63,9 +63,13 @@ class Depacketizer : public IFrameReader, public core::NonCopyable<> {
         // Number of packets dropped during frame construction.
         size_t n_dropped_packets;
 
+        // This frame first sample timestamp.
+        core::nanoseconds_t ts;
```
gavv, line 67: nit: let's rename it to capture_ts for disambiguity.

src/internal_modules/roc_packet/rtp.h
```diff
@@ -45,6 +46,15 @@ struct RTP {
...
     timestamp_t duration;
 
+    //! Timestamp of the first sample at the moment it was captured from an interface.
+    //! @remarks
+    //!  In an ideal case the meaning of this value should be the same on a sender
+    //!  and a receiver, particularly it should store the moment in time the first sample
+    //!  of a packet came into existence. This could be achieved by outbound signaling
+    //!  such as RTCP (with XR extension). In case RTCP is not fully functional
+    //!  this value could store the moment a packet arrived.
+    core::nanoseconds_t capture_timestamp;
```
gavv, line 56: I think capture_timestamp is not really RTP-specific, unlike other fields in this struct. I can imagine that we will somehow fill it for other protocols too. Also its format and usage are protocol-independent, and it does not correspond to anything in RTP header. So I suggest to move it to Packet class.

src/internal_modules/roc_audio/sample_spec.h
```diff
@@ -113,6 +116,9 @@ class SampleSpec {
     //!  Same as samples_per_chan_2_ns(), but supports negative deltas.
     core::nanoseconds_t rtp_timestamp_2_ns(packet::timestamp_diff_t rtp_delta) const;
 
+    //! Convert number of samples (possibly non-integer) to nanoseconds.
+    core::nanoseconds_t n_samples_2_ns(const float n) const;
```
gavv, line 120: For consistency with other methods, let's remove it and instead add `fract_samples_overall_2_ns` and move right after `samples_overall_2_ns()`. So we will have:

```cpp
    //! Convert nanoseconds duration to number of samples per channel.
    size_t ns_2_samples_per_chan(core::nanoseconds_t ns_duration) const;

    //! Convert number of samples per channel to nanoseconds duration.
    core::nanoseconds_t samples_per_chan_2_ns(size_t n_samples) const;

    //! Convert (possibly fractional) number samples per channel to nanoseconds duration.
    core::nanoseconds_t fract_samples_per_chan_2_ns(const float n_samples) const;

    //! Convert nanoseconds duration to number of samples for all channels.
    size_t ns_2_samples_overall(core::nanoseconds_t ns_duration) const;

    //! Convert number of samples for all channels to nanoseconds duration.
    core::nanoseconds_t samples_overall_2_ns(size_t n_samples) const;

    //! Convert (possibly fractional) number samples for all channels to nanoseconds duration.
    core::nanoseconds_t fract_samples_overall_2_ns(const float n_samples) const;
```

gavv (commented):
:heart_eyes:

gavv:
Summary from matrix:

- [ ] add sender counterpart for CaptureTsFiller
- [ ] add pipeline::ReceiverStats (maybe to roc_pipeline/stats.h) with `end_to_end_latency` and `local_latency`
- [ ] add pipeline::ReceiverSource::stats()
- [ ] implement pipeline::ReceiverSource::latency() to return same value as `local_latency` in stats

src/internal_modules/roc_audio/frame.h
```diff
@@ -24,7 +27,7 @@ class Frame : public core::NonCopyable<> {
     //! Construct frame from samples.
     //! @remarks
     //!  The pointer is saved in the frame, no copying is performed.
-    Frame(sample_t* samples, size_t num_samples);
+    Frame(sample_t* samples, size_t num_samples, core::nanoseconds_t ts = 0);
```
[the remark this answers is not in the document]
baranovmv: > 1. Let's rename ts to capture_ts or capture_timestamp

Ok

>  2. Are frames without timestamp legit know? If yes, could you provide examples? If no, I think we should remove default value and instead fix all places were such invalid frames are constructed.

They are not valid, but they get timestamp after instantiation, for example here timestamp is being set by some reader from upstream:

```
Frame temp_frame(temp_data, size);
if (!rp->read(temp_frame)) {
    continue;
}
```

src/internal_modules/roc_audio/channel_mapper.cpp
```diff
@@ -59,6 +59,7 @@ void ChannelMapper::map(const Frame& in_frame, Frame& out_frame) {
                   (unsigned long)out_frame.num_samples());
     }
 
+    out_frame.capture_timestamp() = in_frame.capture_timestamp();
```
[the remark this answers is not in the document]
baranovmv: > ChannelMapper does not preserve flags, I think it should either preserve both flags and timestamps or none

This only is enough convincing for me.

src/internal_modules/roc_packet/rtp.h
```diff
@@ -45,6 +46,15 @@ struct RTP {
     //!  Duration is measured in the same units as timestamp.
     timestamp_t duration;
 
+    //! Timestamp of the first sample at the moment it was captured from an interface.
+    //! @remarks
+    //!  In an ideal case the meaning of this value should be the same on a sender
+    //!  and a receiver, particularly it should store the moment in time the first sample
+    //!  of a packet came into existence. This could be achieved by outbound signaling
+    //!  such as RTCP (with XR extension). In case RTCP is not fully functional
+    //!  this value could store the moment a packet arrived.
```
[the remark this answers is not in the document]
baranovmv: The wording here is intentionally vague, as I don't pretend to fully understand all possible situation/use-cases

src/internal_modules/roc_audio/depacketizer.cpp
```diff
@@ -89,6 +91,9 @@ void Depacketizer::read_frame_(Frame& frame) {
     }
 
     roc_panic_if(buff_ptr != buff_end);
+    if (valid_capture_ts_) {
+        frame.capture_timestamp() = info.ts;
+    }
 
     set_frame_flags_(frame, info);
```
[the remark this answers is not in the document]
baranovmv: Are you saying that there must not be rtp packets in the pipeline without proper capture timestamp? Should we assert this  the code?

src/internal_modules/roc_audio/depacketizer.cpp
```diff
@@ -107,21 +112,32 @@ Depacketizer::read_samples_(sample_t* buff_ptr, sample_t* buff_end, FrameInfo& i
...
             const size_t max_samples = (size_t)(buff_end - buff_ptr);
+            const size_t nsamples = std::min(mis_samples, max_samples);
 
-            buff_ptr = read_missing_samples_(
-                buff_ptr, buff_ptr + std::min(mis_samples, max_samples));
+            buff_ptr = read_missing_samples_(buff_ptr, buff_ptr + nsamples);
+            if (!info.ts) {
+                info.ts =
+                    next_capture_ts_ - sample_spec_.samples_overall_2_ns(mis_samples);
+            }
```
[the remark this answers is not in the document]
baranovmv:
```
info.ts = next_capture_ts_ >
                        sample_spec_.samples_overall_2_ns(mis_samples) ?
                    next_capture_ts_ - sample_spec_.samples_overall_2_ns(mis_samples) : 0;
```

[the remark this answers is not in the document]
baranovmv: As I changed depacketizer in such way to keep `incrementing next_capture_ts_` starting from 0 until we have proper capture_ts.

src/internal_modules/roc_packet/rtp.h
baranovmv: Every time we interact with capture timestamp in the code we rely on number of samples in the packet. So far `capture_timestamp` is strictly bounded to RTP (with an audio stream).

src/internal_modules/roc_audio/depacketizer.cpp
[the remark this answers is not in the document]
gavv: No, I don't think we can require it. For example RTCP can be non functioning.

In this discussion I just suggested to move this piece:

```
    if (valid_capture_ts_) {
        frame.capture_timestamp() = info.ts;
    }
```

into set_frame_flags_() function.

src/internal_modules/roc_audio/sample_spec.cpp
```diff
@@ -84,7 +84,14 @@ core::nanoseconds_t SampleSpec::samples_per_chan_2_ns(const size_t n_samples) co
...
 
-    return rtp_timestamp_2_ns((packet::timestamp_diff_t)n_samples);
+    return fract_samples_overall_2_ns((float)n_samples);
+}
+
+core::nanoseconds_t SampleSpec::fract_samples_per_chan_2_ns(const float samples) const {
+    roc_panic_if_msg(!is_valid(), "sample spec: attempt to use invalid spec: %s",
+                     sample_spec_to_str(*this).c_str());
+
+    return fract_samples_overall_2_ns(samples);
```
gavv, line 94: Looks like a bug.

src/internal_modules/roc_audio/depacketizer.cpp
```diff
@@ -107,21 +108,36 @@ Depacketizer::read_samples_(sample_t* buff_ptr, sample_t* buff_end, FrameInfo& i
...
             buff_ptr = new_buff_ptr;
         }
 
         return buff_ptr;
     } else {
+        const size_t nsamples = size_t(buff_end - buff_ptr);
+        if (valid_capture_ts_) {
+            info.capture_ts = next_capture_ts_;
+            next_capture_ts_ += sample_spec_.samples_overall_2_ns(nsamples);
+        }
```
gavv, line 140: Don't we need:

```
if (!info.capture_ts) {
    info.capture_ts = next_capture_ts_;
}
```

?

src/internal_modules/roc_rtp/validator.cpp
```diff
@@ -66,6 +66,21 @@ bool Validator::check_(const packet::RTP& prev, const packet::RTP& next) const {
         sn_dist = -sn_dist;
     }
 
+    if (!!next.capture_timestamp && !!prev.capture_timestamp) {
+        const bool valid_capt_ts =
+            core::ns_equal(next.capture_timestamp, prev.capture_timestamp,
+                           config_.max_ts_jump)
+            && next.capture_timestamp > prev.capture_timestamp;
```
gavv, line 73: 1. max_ts_jump is in other units. I guess we need a new config field.
2. Validator is covered with tests. Please add a test for this new feature.
3. Let's move capture_ts check to the end? I think timestamp and seqnum checks are more informative in case of error (because they are in packet fields directly, not estmaited as capture ts) and it'd be better to make them first.

src/internal_modules/roc_rtp/validator.cpp
```diff
@@ -66,6 +66,21 @@ bool Validator::check_(const packet::RTP& prev, const packet::RTP& next) const {
...
+        if (!valid_capt_ts) {
+            roc_log(LogDebug,
+                    "rtp validator: too long capture timestamp jump:"
+                    " prev=%lu next=%lu dist=%lu",
+                    (unsigned long)prev.capture_timestamp,
+                    (unsigned long)next.capture_timestamp, (unsigned long)sn_dist);
```
gavv, line 79: Instead of sn_dist, we should print distance between capture timestamps.

src/internal_modules/roc_rtp/validator.cpp
```diff
@@ -66,6 +66,21 @@ bool Validator::check_(const packet::RTP& prev, const packet::RTP& next) const {
         sn_dist = -sn_dist;
     }
 
+    if (!!next.capture_timestamp && !!prev.capture_timestamp) {
+        const bool valid_capt_ts =
+            core::ns_equal(next.capture_timestamp, prev.capture_timestamp,
+                           config_.max_ts_jump)
```
gavv, line 72: This call to ns_equal here looks weird.

Maybe rename ns_equal to ns_in_delta?

src/internal_modules/roc_audio/EndToEndLatencyMonitor.cpp
```diff
@@ -0,0 +1,49 @@
...
+//
+// Created by misha on 31.08.23.
+//
+
+#include "EndToEndLatencyMonitor.h"
+
+namespace roc {
+namespace audio {
+
+EndToEndLatencyMonitor::EndToEndLatencyMonitor(IFrameReader& reader)
```
gavv, line 18: I suggest to merge this functionality into LatencyMonitor.

Later LatencyMonitor will anyway need it when we will implement clock sync mode based on RTCP. So there is no need to have two monitors.

src/internal_modules/roc_audio/latency_monitor.cpp
```diff
@@ -250,13 +253,16 @@ void LatencyMonitor::report_latency_(packet::timestamp_diff_t latency) {
...
+            (long)latency, (double)latency_ / core::Millisecond,
             (unsigned long)target_latency_,
             (double)input_sample_spec_.rtp_timestamp_2_ns(
                 (packet::timestamp_diff_t)target_latency_)
                 / core::Millisecond);
 }
 
+core::nanoseconds_t LatencyMonitor::latency() const {
+    return latency_;
+}
```
gavv, line 265: Giving it second thought... This local latency fields is very arguable. It will oscillate a lot because of network jitter, so this is probably not something that the user would expect from "latency of pipeline".

I suggest to remove local latency for now, and add only end to end latency in this PR.

src/internal_modules/roc_pipeline/receiver_stats.h
```diff
@@ -0,0 +1,41 @@
...
+#ifndef ROC_PIPELINE_RECEIVER_STATS_H_
+#define ROC_PIPELINE_RECEIVER_STATS_H_
+
+#include "roc_core/time.h"
+
+namespace roc {
+namespace pipeline {
+
+//! Per participant statistics.
+struct Stats {
```
gavv, line 21: Let's rename it to SessionStats. And remove local_latency as described above.

src/internal_modules/roc_pipeline/receiver_session.h
```diff
@@ -92,6 +94,11 @@ class ReceiverSession : public core::RefCounted<ReceiverSession, core::ArenaAllo
     //! Handle estimated link metrics.
     void add_link_metrics(const rtcp::LinkMetrics& metrics);
 
+    //! Gets recent latency statistics.
+    //! @returns
+    //! True if statistics is valid.
+    bool stats(Stats& stats) const;
```
gavv, line 100: Let's simplify it:

```
Stats stats() const;
```

I think there is no need to return bool. Instead, if some of the stats are not available for whatever reason (e.g. there is no latency info), they should be set to zero.

src/internal_modules/roc_audio/depacketizer.cpp
baranovmv: No, assertion is enough:

```
const size_t nsamples = size_t(buff_end - buff_ptr);
if (valid_capture_ts_) {
    roc_panic_if(info.capture_ts);
    info.capture_ts = next_capture_ts_;
    next_capture_ts_ += sample_spec_.samples_overall_2_ns(nsamples);
}
return read_missing_samples_(buff_ptr, buff_end);
```

As we fill all the requested samples, the outer while loop will never run the second iteration.

src/internal_modules/roc_audio/latency_monitor.cpp
baranovmv: May be we could get smoothed local latency from `FrequencyEstimator`, but it needs some experiments.

src/internal_modules/roc_audio/depacketizer.cpp
baranovmv: My bad, assertion fails. We have to check this!

baranovmv: @gavv  Augmented `TEST(depacketizer, frame_flags_incompltete_blank)` to cover this edge-case, and stumbled upon probable bug there, pls take a look

gavv:
Follow-up commits:

* 28983a578396e584ff3937c2cb2036b453d74aae
* e883429939afab8c4f14fe80dd9e5ad8fd5880c1
* 219c6d92757c7cca6b805b0e34caa99f4c1fe394

PR #545, opened 2023-09-01
Added version check to CI checks

dheerajd5:
This PR adds a version-check to CI checks. It takes the version fetched from git and from the file `version.h` and compares it.
https://github.com/roc-streaming/roc-toolkit/issues/497

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,6 +105,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+          
 
       - name: Build and test
         run: |
@@ -152,6 +153,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+          fetch-tags: true
 
       - name: Build and test
         run: |
@@ -310,6 +312,34 @@ jobs:
       - name: Build and test
         run: |
           scripts/ci_checks/macos/${{ matrix.script }}.sh
+
+      - name: Handle failure
+        if: failure()
+        run: |
+          scripts/ci_checks/failure.sh
+
+  release-checks:
+    runs-on: ubuntu-latest
+
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - script: linux-checks/check-version
+
+    name: ${{ matrix.script }}
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          fetch-tags: true
+
+      - name: Build and test
+        run: |
+          scripts/ci_checks/${{ matrix.script }}.sh
 
       - name: Handle failure
         if: failure()
--- a/scripts/ci_checks/linux-checks/check-version.sh
+++ b/scripts/ci_checks/linux-checks/check-version.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+
+tagname="$(git tag -l | tail -n1)"
+
+header_tagname="v$(python3 scripts/scons_helpers/parse-version.py)"
+
+if [[ "$tagname" != "$header_tagname" ]]; then
+  echo >&2 
+  echo >&2"version check FAILED"
+  echo >&2"the git tag name and version declared in version.h is not the same"
+  echo >&2"${tagname} git version tag"
+  echo >&2"${header_tagname} header version tag"
+  echo >&2
+  exit 1
+fi
+
+echo "version check SUCCEEDED"
+echo 
+


scripts/ci_checks/linux-checks/check-version.sh
```diff
@@ -0,0 +1,21 @@
...
+
+header_tagname="v$(python3 scripts/scons_helpers/parse-version.py)"
+
+if [[ "$tagname" != "$header_tagname" ]]; then
+  echo
+  echo "version check FAILED" >&2
+  echo "the git tag name and version declared in version.h is not the same" >&2
+  echo "${tagname} git version tag" >&2
+  echo "${header_tagname} header version tag" >&2
+  echo -e "\n"
```
adrianopol, line 16: This variant is more readable (and also prevents you from forgetting to do stream redirection as you did in lines 11 and 16):

```suggestion
  echo >&2
  echo >&2 "version check FAILED"
  echo >&2 "the git tag name and version declared in version.h is not the same"
  echo >&2 "${tagname} git version tag"
  echo >&2 "${header_tagname} header version tag"
  echo >&2 
```

Also `echo` without arguments prints a new line. `echo -e "\n"` prints two, but I suppose one is sufficient here.

scripts/ci_checks/linux-checks/check-version.sh
```diff
@@ -0,0 +1,21 @@
...
+  echo "version check FAILED" >&2
+  echo "the git tag name and version declared in version.h is not the same" >&2
+  echo "${tagname} git version tag" >&2
+  echo "${header_tagname} header version tag" >&2
+  echo -e "\n" 
+  exit 1
+fi
+
+echo "version check SUCCEEDED"
+echo -e "\n"
```
adrianopol, line 21: add a new line in the end of the file please

.github/workflows/build.yml
```diff
@@ -314,4 +322,4 @@ jobs:
       - name: Handle failure
         if: failure()
         run: |
-          scripts/ci_checks/failure.sh
+          scripts/ci_checks/failure.sh
```
adrianopol, line 325: add a new line in the end of the file please

scripts/ci_checks/linux-checks/check-version.sh
adrianopol, line 21: Just a plain `echo` will be enough here (if you want to leave a single new line).

P.S.: Also, I started to think about is it OK to write a failure-text to stderr and a succeed-text to stdout. I suppose this script is not going to be used in a pipe where its output would be parsed by some other program. If so, it's completely fine to use different streams here (as you did in this pull request).

adrianopol (changes requested):

gavv:
Two small comments:

* we shouldn't run this check unless we're building a tag naned as v* - because we usually update header before creating a tag

* ideally we shouldn't checkout full history unless we're doing this check, to avoid slowing down regular builds

adrianopol (approved):
LGTM; also pay attention to gavv's comments please.
Thank you!

gavv:
@dheerajd5 As you mentioned in the issue, android failure is not related to your issue. I restarted that job.

dheerajd5:
> Two small comments:
> 
> * we shouldn't run this check unless we're building a tag naned as v* - because we usually update header before creating a tag
> * ideally we shouldn't checkout full history unless we're doing this check, to avoid slowing down regular builds

Hey, So I looked into conditional execution in Actions, to check if we're building a tag with v*, we can probably check the latest commit message, and see if the string matches a v* format. 
And for the second point, we can pass an if condition to see what kind of check we're going to run and based of that if it is necessary to pull the full history. I will try to implement the solution to the second point. 

One more doubt, there's few commits that have been commited to the `roc-streaming:develop` branch after I put a PR, so when I pull these changes locally and then push with the changes I've added it adds all these commits as co-authored by me, how do I make it so these changes are not co authored by me.

dheerajd5:
Hey so I thought if I reset my branch to the `roc-streaming:develop` push those changes and then add my version-check script and then push those changes, it would push the commits with the correct authors, but it just closed the PR, do I need to issue another PR or will reopening this one be enough?

gavv:
I think we should extract this from `linux-checks` into separate job. Only that job will do deep checkout and fetch tags, and other jobs won't be affected. E.g. we can call it `release-checks`.

The new job then can be enabled conditionally only for properly named tags. Here is how you can do it: https://github.com/roc-streaming/roc-go/blob/6212890ef56528624350571ce13f308e0a3ae11c/.github/workflows/build.yaml#L219C53-L219C53

dheerajd5:
There are also ways to conditionally execute steps in a job, that's what I've done. Although the solution is a bit hard coded, the workflow only executes the step, fetch tag, if the script that is going to be run is version-check.

But in this case, there was no way that I was aware of where I could conditionally run it based on the tag name.

I'll go ahead with the solution you suggested.

dheerajd5:
Hey, so I implemented the changes, but it's not passing a few unrelated test. I don't really have an idea why, Namely, the `macos13` series, Here's the workflow run https://github.com/dheerajd5/roc-toolkit/actions/runs/6157211700 and here's the commit page, (https://github.com/dheerajd5/roc-toolkit/commit/64f2fd63291d9b06a4870e4b0d7b27d5465297e3)

gavv:
> Hey, so I implemented the changes, but it's not passing a few unrelated tests. I don't really have an idea why, Namely, the `macos13` series

Could you please fetch fresh develop and rebase on it? Should be fixed now.

dheerajd5:
Hey so I did what you told, it has passed all the checks now. It hasn't run the check I've written cause I didn't upload this with a version tag.

gavv:
Thanks. Tested it in a fork, there were problems.

Follow-up commit with fixes: 0266785edcb87d7349a4329b7de026c4c7c15c18

Changes:

* renamed "package" workflow to "release" and moved release checks to that workflow
* replaced "fetch-tags" with "fetch-depth" because of this issue: https://github.com/actions/checkout/issues/1467
* removed ">&2" because errors were not displaying in github actions log (@adrianopol FYI)

dheerajd5:
Ahh, sorry I couldn't submit a more complete code, but enjoyed working, thanks guys @gavv @adrianopol

gavv:
You're welcome!

Issue #539, opened 2023-07-20
Frame timestamps


gavv:
- [x] Add capture_timestamp field to:

   - [x] audio::Frame
   - [x] packet::RTP

   (It should have type of nanoseconds_t and hold capture/origin timestamp of the frame in local system clock domain, in nanoseconds since unix epoch)

- [x] Teach components to forward capture_timestamp:

   - [x] ChannelMapperReader
   - [x] ChannelMapperWriter
   - [x] Depacketizer
   - [x] Packetizer
   - [x] ResamplerReader
   - [x] ResamplerWriter

  (And update corresponding tests).

- [x] Update printer

   - [x] Packet::print() should include capture timestamp
   - [x] Frame::print() should include capture timestamp

- [x] Tests

   - [x] cover CTS in tests for ChannelMapperReader and ChannelMapperWriter
   - [x] cover CTS in tests for ResamplerReader and ResamplerWriter
   - [x] cover CTS in tests for Packetizer and Depacketizer
   - [x] cover CTS in pipeline tests
   - [x] ResamplerReader, ResamplerWriter: tests for zero CTS
   - [x] ResamplerReader, ResamplerWriter: tests for small CTS
   - [x] Packetizer: test for zero CTS
   - [x] Depacketizer: test for small CTS
   - [x] Depacketizer: test for zero CTS in first packet of frame and non-zero in second
   - [x] fix timestamp_mapping_remixing in receiver_source
   - [x] fix timestamp_mapping_remixing in sender_sink

- [x] Bug-fixes

   - [x] Depacketizer, ResamplerReader, ResamplerWriter: handling of very small CTS
   - [x] Depacketizer: when setting CTS, take into account number of samples already written to frame
   - [x] LatencyMonitor: reported e2e latency is lower than niq latency when using PulseAudio
   - [x] crash in roc-send when --repair endpoint is not provided

- [x] Refactoring

   - [x] rename timestamp_t to stream_timestamp_t (and corresponding helper methods)
   - [x] rename SampleSpec.ns_2_rtp_timestamp and rtp_timestamp_2_ns (rtp_timestamp => stream_timestamp)
   - [x] move ntp_timestamp_t from roc_packet/units.h to roc_packet/ntp.h

- [x] Documentation

   - [x] Add a page to internals that briefly describes stream timestamp and wallclock timestamp, difference between them, and usage.

gavv:
[...]

PR #549, opened 2023-09-12
Fixes timestamp mapping remixing tests

baranovmv:
#539

In `receiver_source` test `recv_timestamp_mapping_remixing` the failure was caused by the issue in **Depacketizer**. As you suggested (after my localization): [here](https://github.com/roc-streaming/roc-toolkit/issues/539#issuecomment-1712284093):

> We should check if Depacketizer correctly handles case when first packet for frame had zero CTS, and second packet for the same frame had non-zero CTS.
> 
> I think it's not covered right now.
> 
> We need a test for this and maybe a fix.

In `sender_sink` tests the issue is simple: Resampler backend throw away first n samples (N==latency), which was not properly reflected in `test::PacketReader`

### DIFF
--- a/src/tests/roc_pipeline/test_helpers/packet_reader.h
+++ b/src/tests/roc_pipeline/test_helpers/packet_reader.h
@@ -69,9 +69,9 @@ public:
         abs_offset_ += samples_per_packet;
     }
 
-    void read_nonzero_packet(size_t samples_per_packet,
-                             const audio::SampleSpec& sample_spec,
-                             core::nanoseconds_t base_capture_ts = -1) {
+    packet::PacketPtr read_nonzero_packet(size_t samples_per_packet,
+                                          const audio::SampleSpec& sample_spec,
+                                          core::nanoseconds_t base_capture_ts = -1) {
         packet::PacketPtr pp = read_packet_();
 
         audio::sample_t samples[MaxSamples] = {};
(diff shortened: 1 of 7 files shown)


src/tests/roc_pipeline/test_helpers/utils.h
```diff
@@ -24,7 +25,8 @@ namespace {
 
 const double SampleEpsilon = 0.00001;
 
-const core::nanoseconds_t TimestampEpsilon = core::Microsecond;
+const core::nanoseconds_t TimestampEpsilon =
+    (core::nanoseconds_t)(1. / DefaultSampleRate * core::Second);
```
gavv, line 29: Different tests may use different sample rates.

We can rework expect_capture_timestamp() so that it accepts sample_spec and uses it to calculate epsilon. Or hard-code rate-independent epsilon that is just large enough for all rates.

src/tests/roc_pipeline/test_helpers/packet_reader.h
```diff
@@ -165,9 +167,9 @@ class PacketReader : public core::NonCopyable<> {
                           core::nanoseconds_t base_ts) {
         CHECK(pkt.rtp());
 
-        if (base_ts < 0) {
+        if (base_ts == 0) {
             LONGS_EQUAL(0, pkt.rtp()->capture_timestamp);
-        } else {
+        } else if (base_ts > 0) {
```
gavv, line 172: This rules became quite a bit complicated.

Suggestion: revert changes in PacketReader and just don't use it in TEST(sender_sink, timestamp_mapping_remixing_packet_reader), instead read packets directly from queue and check their CTS.

gavv (commented):
Thanks for PR!

PR #550, opened 2023-09-19
gh-546: Support NTP 2036 year base

### LINKED ISSUE #546: Support NTP 2036 year base
Currently we don't implement this bit from RFC:

```
      As the NTP timestamp format has been in use for the last 17 years,
      it remains a possibility that it will be in use 40 years from now
      when the seconds field overflows. As it is probably inappropriate
      to archive NTP timestamps before bit 0 was set in 1968, a
      convenient way to extend the useful life of NTP timestamps is the
      following convention: If bit 0 is set, the UTC time is in the
      range 1968-2036 and UTC time is reckoned from 0h 0m 0s UTC on 1
      January 1900. If bit 0 is not set, the time is in the range 2036-
      2104 and UTC time is reckoned from 6h 28m 16s UTC on 7 February
      2036. Note that when calculating the correspondence, 2000 is not a
      leap year. Note also that leap seconds are not counted in the
      reckoning.
```

https://datatracker.ietf.org/doc/html/rfc2030

Affected functions:

* ntp_2_unix()
* unix_2_ntp()
* ntp_equal_delta()

gavv:
Fixes #546.

unix_2_ntp() and ntp_2_unix() are updated according to RFC 2030.

ntp_equal_delta() is removed because it was never used and proper implementation is not trivial.

### DIFF
--- a/src/internal_modules/roc_packet/ntp.cpp
+++ b/src/internal_modules/roc_packet/ntp.cpp
@@ -16,32 +16,47 @@ namespace packet {
 
 namespace {
 
-// Number of seconds fron NTP epoch (1900-01-01) to Unix epoch (1970-01-01).
-// Equals to 70 years + 17 leap days.
-static uint64_t unix_epoch = uint64_t(70 * 365 + 17) * (24 * 60 * 60);
+// Number of seconds from NTP epoch (1900-01-01) to Unix epoch (1970-01-01).
+// Equals to 70 years with 17 leap days.
+static uint64_t UnixEpoch = uint64_t(70 * 365 + 17) * (24 * 3600);
+
+// Number of seconds from Unix epoch to NTP Era 1 (7 Feb 2036 6:28:16).
+// Equals to 66 years with 16 leap days, 37 days, 6 hours, 28 minutes and 16 seconds.
+// See RFC 2030.
+static uint64_t Era1 =
+    uint64_t(66 * 365 + 16 + 37) * (24 * 3600) + (6 * 3600) + (28 * 60) + 16;
 
 } // namespace
 
-bool ntp_equal_delta(ntp_timestamp_t a, ntp_timestamp_t b, ntp_timestamp_t delta) {
-    ntp_timestamp_t abs_error = std::max(a, b) - std::min(a, b);
-    return abs_error <= delta;
-}
-
 ntp_timestamp_t unix_2_ntp(core::nanoseconds_t unix_time) {
-    const core::nanoseconds_t ntp_time =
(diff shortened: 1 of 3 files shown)


dshil:
Hi @gavv, this comment isn't directly related to this PR, just some thoughts. it was a little bit tricky for me to find out where was the actual conversation of the fraction part of timestamps. I was playing a little bit with a code, and tried to make it more understandable, at least for me.

```cpp
namespace {

uint64_t bits_pack(uint32_t hi, uint32_t lo) {
    return ((uint64_t)hi << 32) | (lo & 0x00000000FFFFFFFFULL);
}

void bits_unpack(uint64_t value, uint32_t& hi, uint32_t& lo) {
    hi = (value & 0xFFFFFFFF00000000ULL) >> 32;
    lo = value & 0xFFFFFFFFULL;
}

} // namespace

ntp_timestamp_t nanoseconds_2_ntp(core::nanoseconds_t ns_delta) {
    roc_panic_if_msg(ns_delta < 0, "ntp: can not convert negative delta to ntp");

    const uint64_t seconds = (uint64_t)(ns_delta / core::Second);
    const uint64_t fraction = (uint64_t)ns_delta - seconds * core::Second;
    const uint64_t nans = (fraction << 32) / core::Second;

    return bits_pack(seconds, nans);
}

core::nanoseconds_t ntp_2_nanoseconds(ntp_timestamp_t ntp_delta) {
    uint32_t hi = 0;
    uint32_t lo = 0;
    bits_unpack(ntp_delta, hi, lo);

    const core::nanoseconds_t seconds = (core::nanoseconds_t)hi;
    const core::nanoseconds_t nans = (core::nanoseconds_t)(lo * core::Second) >> 32;

    return seconds * core::Second + nans;
}
```

Now, it seems clear how the fractional part should be computed:
- NTP -> UNIX, the fractional part should be multiplied by the factor of `core::Second / (1 << 32)`
- UNIX -> NTP, the fractional part should be multiplied by the factor of `(1 << 32) / core::Second`

---

Also, I want to mention the following resources that were very helpful with understanding NTP and UNIX timestamps:

- [A Very Short Introduction to NTP Timestamps](https://tickelton.gitlab.io/articles/ntp-timestamps/#:~:text=To%20convert%20from%20NTP%20to,the%20result%20by%202%5E32.)
- [RFC 2030 | NTP Timestamp Format](https://datatracker.ietf.org/doc/html/rfc2030#section-3), which was already mentioned in the issue
- [The NTP Era and Era Numbering](https://www.eecis.udel.edu/~mills/y2k.html)

dshil (approved):

PR #585, opened 2023-10-01
Slab pool boundary guard

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/582

# What

* Used sizeof(AlignMax) as boundary size.
* Updated PoisonOps.
* Updated PoolImpl.

# Testing

* Wrote tests can be un-ignored to cause a panic.


gavv:
Thanks for PR, will review in upcoming days! Android timeout is a known issue, I've restarted CI.

nolan-veed:
Currently, I put a boundary before and after user the data area. But, it just occurred to me that if the user data area is already padded (max-alignment) - should we also check that for buffer overruns.

src/tests/roc_core/test_pool.cpp
```diff
@@ -479,5 +479,58 @@ TEST(pool, embedded_capacity_reuse) {
...
+    CHECK(pointer);
+
+    char* data = (char*)pointer;
+    data--;
+    *data = 0x00;
+
+    pool.deallocate(pointer);
+}
+
+IGNORE_TEST(pool, panics_on_boundary_guard_after_non_embedded_object_violation) {
```
gavv, line 516: The name is a bit too long and hard to read, maybe use something shorter like `guard_before_object` and `guard_after_object`?

gavv (commented):
Thanks again for the patch, looks fine. Here are some comments.

src/internal_modules/roc_core/poison_ops.h
```diff
@@ -28,14 +28,25 @@ class PoisonOps {
     //! Helps catching use after free bugs.
     static void after_use(void* data, size_t size);
 
+    //! Poison memory that is a boundary guard.
+    //! Helps catching buffer overflow/underflow bugs.
+    static void prepare_boundary_guard(void* data, size_t size);
+
+    //! Checks memory that is a boundary guard and panics if not.
+    static void check_boundary_guard(void* data, size_t size);
```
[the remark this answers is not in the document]
nolan-veed: > Also, I think we could then merge AlignOps into MemoryOps. If you wish, you're welcome to do it, or I can handle it later.

I'll leave this for later. It'll touch a lot of other files.

src/internal_modules/roc_core/poison_ops.cpp
```diff
@@ -36,5 +36,33 @@ void PoisonOps::after_use(void* data, size_t size) {
...
+    }
+
+    if (size == 0) {
+        return;
+    }
+    char* data_begin = (char*)data;
+    char* data_end = (char*)data + size;
+    while (data_begin < data_end) {
+        if (*data_begin != Pattern_BoundaryGuard)
+            roc_panic("poisoner: data is not boundary guard");
```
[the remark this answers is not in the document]
nolan-veed: I did wonder about 2 as well. Will do.

src/internal_modules/roc_core/poison_ops.h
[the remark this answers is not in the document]
nolan-veed: > Also, I checked [wiki](https://en.wikipedia.org/wiki/Buffer_overflow_protection), and there is a nice term for these kinds of guards: "canary". I suggest to use it.
> 
TIL... canary.

src/internal_modules/roc_core/pool_impl.cpp
```diff
@@ -40,7 +40,8 @@ PoolImpl::PoolImpl(const char* name,
     , n_used_slots_(0)
     , slab_min_bytes_(clamp(min_alloc_bytes, preallocated_size, max_alloc_bytes))
     , slab_max_bytes_(max_alloc_bytes)
-    , slot_size_(AlignOps::align_max(std::max(sizeof(Slot), object_size)))
+    , slot_size_no_boundary_(AlignOps::align_max(std::max(sizeof(Slot), object_size)))
+    , slot_size_(BoundarySize + slot_size_no_boundary_ + BoundarySize)
```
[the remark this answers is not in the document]
nolan-veed: I've considered the above.

Slot size is no max of sizeof Slot and object size (including padding, if present and 2 boundaries).

So, the take_slot_from_user_() now checks for the guard before the object, and after the object (including any padding, if present).

src/internal_modules/roc_core/pool_impl.cpp
```diff
@@ -104,20 +110,40 @@ void PoolImpl::deallocate(void* memory) {
...
+    bool canary_ok = MemoryOps::check_canary(canary_before, BoundarySize);
+    MemoryOps::poison_after_use(memory, object_size_);
+    canary_ok &=
+        MemoryOps::check_canary(canary_after, object_size_padding_ + BoundarySize);
+
+    if (!canary_ok) {
+        num_buffer_overflows_++;
+        roc_log(LogError, "pool: buffer overflow detected");
+        roc_panic_if_not(flags_ & PoolFlags_DisableOverflowPanic);
+    }
```
gavv, line 144: I suggest to panic before poisoning memory.

This way, if we catch this panic in debugger, we can inspect original memory before poisoning.

src/internal_modules/roc_core/pool.h
```diff
@@ -23,6 +23,11 @@
 namespace roc {
 namespace core {
 
+//! Memory pool flags.
+enum PoolFlags {
+    PoolFlags_DisableOverflowPanic = 0x01, //! Disables overflow panic and logs instead.
+};
```
gavv, line 29: A few nitpicks:

* Let's rename it from `PoolFlags_DisableOverflowPanic` to `PoolFlag_DisableOverflowPanic`, because it's a name for single flag.

* Usually we use `(1 << N)` instead of `0xM` to define flags.

* Negations are a bit harder to handle when reading code. So I'd suggest to add something like `PoolFlag_PanicOnOverflow` or `PoolFlags_OverflowPanics` and use it in default ctor argument.

* If doxygen comment is after declaration, IIRC it should be `//!<` instead of `//!`.

src/internal_modules/roc_core/pool_impl.cpp
nolan-veed: Nice one! I hadn't thought about that use case. Will do.

src/internal_modules/roc_core/pool.h
nolan-veed: Sorry. Still adapting to the project.

src/internal_modules/roc_core/pool_impl.cpp
nolan-veed: I've separated the ok variables, so that you can see at a glance, which check failed...
image

gavv:
Thanks for update!

Here is a small follow-up commit: c25b331b2028918b4dd2c66caec722c88c20e463

It adds name to other panics, and fixes small bug:

```cpp
if ((flags_ & PoolFlag_PanicOnOverflow) == 1) {
```

will work only if PoolFlag_PanicOnOverflow is equal to 1, so it's replaced with:

```cpp
if ((flags_ & PoolFlag_PanicOnOverflow) != 0) {
```

Issue #183, opened 2019-05-15
Add support for runtime pipeline errors


gavv:
Currently, pipeline readers and writers can report errors on initialization, but can't report error when read or write is called.

There are at least two places where we need to report such errors:

* FEC writer and reader (they may perform reallocations);
* PulseAudio sink (it may lost connection with the server).

We need to make read and write to return bool and ensure that all pipeline components carefully forward those errors. We also need to add tests for that.

Then we will be able to report errors from the components listed above.

gavv:
Update components:

- [ ] `audio::Depacketizer`:
   - on `StatusNoData`, break reading loop
   - on `StatusAgain`, retry packet read
   - on other error, forward error from frame read()
- [ ] `fec::Reader` - return `StatusDead` when after `alive_` becomes false
- [ ] `audio::Watchdog`, `audio::LatencyMonitor` - return `StatusDead` when after `alive_` becomes false
- [ ] `rtp::Validator` - return `StatusAgain` instead of `StatusNoData` when it drops packet; Depacketizer should retry read in this case
- [ ] `node::SenderEncoder`, `node::ReceiverDecoder`, `node::Sender`, `node::Receiver` - return `StatusNotFound` when attempting to access slot or interface that is not available

Update adapters:

- [ ] adapters should return status instead of bool
- [ ] add adapter to convert status to API error

Update API:

- [ ] add error codes to C API
- [ ] forward errors from adapters to user
- [ ] forward errors from internals to user

gavv:
We also should migrate roc_netio TCP classes to roc_status codes. SocketError enum should be hidden implementation detail.

PR #591, opened 2023-10-03
Add support for status codes in packet writer

dshil:
GH-303


src/internal_modules/roc_audio/packetizer.cpp
```diff
@@ -131,7 +132,13 @@ void Packetizer::end_packet_() {
         pad_packet_();
     }
 
-    writer_.write(packet_);
+    const status::StatusCode code = writer_.write(packet_);
+    if (code != status::StatusOK) {
+        // TODO: handle returned status code (gh-183).
+        roc_panic_if_msg(code != status::StatusNoData,
+                         "packetizer: failed to write packet: status=%s",
+                         status::code_to_str(code));
```
gavv, line 140: Here and in similar places: I think writer can never return StatusNoData (what would it mean?). So here we need unconditional panic (for now, later we will unconditionally forward code I think).

src/internal_modules/roc_fec/writer.cpp
```diff
@@ -92,12 +93,13 @@ bool Writer::resize(size_t sblen, size_t rblen) {
...
 
-void Writer::write(const packet::PacketPtr& pp) {
+status::StatusCode Writer::write(const packet::PacketPtr& pp) {
     roc_panic_if_not(is_valid());
     roc_panic_if_not(pp);
 
     if (!alive_) {
-        return;
+        // TODO: return StatusDead (gh-183)
+        return status::StatusNoData;
```
gavv, line 102: Since StatusNoData is not appropriate for writers, and upper code is not ready yet for errors for writers, before 183 I think we should return StatusOK in cases where we don't fail currently.

src/internal_modules/roc_fec/writer.h
```diff
@@ -77,7 +77,7 @@ class Writer : public packet::IWriter, public core::NonCopyable<> {
     //! @remarks
     //!  - writes the given source packet to the output writer
     //!  - generates repair packets and also writes them to the output writer
-    virtual void write(const packet::PacketPtr&);
+    virtual ROC_ATTR_NODISCARD status::StatusCode write(const packet::PacketPtr&);
```
gavv, line 80: Do we need the attribute in derived classes, or only in the interface?

src/internal_modules/roc_node/receiver_decoder.cpp
```diff
@@ -146,11 +148,11 @@ bool ReceiverDecoder::write(address::Interface iface, const packet::PacketPtr& p
                 "receiver decoder node:"
                 " can't write to %s interface: interface not activated",
                 address::interface_to_str(iface));
-        return false;
+        // TODO: return StatusNotFound (gh-183)
+        return status::StatusNoData;
```
gavv, line 152: I guess for now we should return Unknown

src/internal_modules/roc_packet/shipper.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+status::StatusCode Shipper::write(const PacketPtr& packet) {
+    if (address_.has_host_port()) {
+        packet->add_flags(Packet::FlagUDP);
+        packet->udp()->dst_addr = address_;
+    }
+
+    if ((packet->flags() & Packet::FlagComposed) == 0) {
+        if (!composer_.compose(*packet)) {
+            roc_panic("shipper: can't compose packet");
```
gavv, line 28: Let's add TODO for returning status instead of panic.

src/internal_modules/roc_packet/shipper.h
```diff
@@ -0,0 +1,45 @@
...
+namespace packet {
+
+//! Prepare a packet for being sent.
+class Shipper : public IWriter, public core::NonCopyable<> {
+public:
+    //! Initialize.
+    //!  - @p address - address that is assigned to a packet.
+    //!  - @p composer to compose a packet if necessary.
+    //!  - @p writer to write outgoing packet.
+    Shipper(const address::SocketAddr& address, IComposer& composer, IWriter& writer);
```
gavv, line 30: nit: better rename address to dst_address or something like this, to avoid confusion, because packet has two addresses: src and dst.

src/internal_modules/roc_packet/sorted_queue.cpp
```diff
@@ -38,7 +38,8 @@ void SortedQueue::write(const PacketPtr& packet) {
                 "sorted queue: queue is full, dropping packet:"
                 " max_size=%u",
                 (unsigned)max_size_);
-        return;
+        // TODO: return StatusAgain (gh-183)
+        return status::StatusNoData;
```
gavv, line 42: It shouldn't be StatusAgain. In this case upper code will immediately retry writing this packet. But we don't want it, if queue is full, we really want to drop packet. So we should drop and return OK.

src/internal_modules/roc_packet/sorted_queue.cpp
```diff
@@ -56,7 +57,8 @@ void SortedQueue::write(const PacketPtr& packet) {
 
         if (cmp == 0) {
             roc_log(LogDebug, "sorted queue: dropping duplicate packet");
-            return;
+            // TODO: return StatusBadArg (gh-183)
+            return status::StatusNoData;
```
gavv, line 61: Same: dropping duplicate packets is a legit case.

src/internal_modules/roc_pipeline/sender_endpoint.cpp
```diff
@@ -99,10 +99,16 @@ SenderEndpoint::SenderEndpoint(address::Protocol proto,
     }
 
     composer_ = composer;
+    if (!composer_) {
+        return;
+    }
+
+    writer_.reset(new (arena) packet::Shipper(dest_address, *composer_, dest_writer),
+                  arena);
```
gavv, line 107: Instead of `core::ScopedPtr<packet::IWriter>`, let's use `core::Optional<packet::Shipper>` to avoid unneeded allocation.

src/internal_modules/roc_rtcp/session.h
```diff
@@ -44,7 +44,7 @@ class Session {
 
     //! Parse and process incoming packet.
     //! Invokes session hooks methods during processing.
-    void process_packet(const packet::PacketPtr& packet);
+    virtual ROC_ATTR_NODISCARD status::StatusCode write(const packet::PacketPtr& packet);
```
gavv, line 47: I'd keep it `status::StatusCode process_packet()` and don't convert rtcp::Session to packet writer.

The semantics is quite different. Packets writers are chained to form sender pipeline that handles transport packets. rtpc::Session is not part of that pipeline. It handles control packets, it's not chained. Instead it invokes various hooks bases on what it found in the packet being processed.

I think it makes sense to keep its interface distinguishable from other writers.

gavv: BTW, thanks for noticing that we need statuses in rtcp::Session, I didn't think about it.

Later we'll need to forward these errors too. I've added it to #614.

gavv (commented):
Thanks for PR!

src/internal_modules/roc_fec/writer.h
dshil: In derived classes too, otherwise nodisard error won't be triggered.

Nodisard is triggered:

```cpp
void foo(const packet::PacketPtr& pp, packet::IWriter& writer) {
   writer.write(pp);
}
```

Nodisard isn't triggered:

```cpp
struct FooWriter : public packet::IWriter {
   virtual status::StatusCode write(const packet::PacketPtr& pp) {
   }
};

void foo(const packet::PacketPtr& pp, FooWriter& writer) {
   writer.write(pp);
}
```

src/internal_modules/roc_rtcp/session.h
dshil: I think also about packet::IComposer, probably it's worth to replace `bool` via `status::StatusCode` too.

gavv:
That was huge!

I've rebased PR and fixed several minor issues (mostly mechanical), now going to merge it.

What I changed: (75641bf065e360a5fc58c2db4039cb9e8b14a0cd)

* Added more TODO comments and reformatted them as `TODO(gh-183): text`.

* Reverted change in ReceiverEndpoint: use ReceiverSessionGroup instead of IWriter as before. On one hand, polymorphism is not really needed there (except tests, but see below). On the other hand, pipeline has quite complicated layout, and it's very convenient to use static typing to be able to see call graph from code.

* Pruned SenderEndpoint and ReceiverEndpoint tests a bit:

   * First, I think it's not the best idea to add endpoint-level tests for how packets are going (I mean tests write_read_packet, pull_packets_failed). We already test packet handling in GREAT detail in source/sink-level tests, which cover endpoints as well (route_packets, pull_packets are covered). In source/sink tests, we already have a lot of facilities to generate mock input and to validate output (e.g. to produce valid and corrupt RTP packets), and I'd like to avoid duplicating it in endpoint-tests (which will happen if we'll start covering SenderEndpoint and ReceiverEndpoint at endpoint level). So I kept endpoint tests for construction, but dropped tests for packet flow.

   * Second, I also simplified no-memory test. I think there is no need to check that some protocols can be used even when arena fails (I don't consider it a feature, just a coincidence),  so I excluded them from tests and tests became much cleaner.
 
* In test::Proxy from public_api tests, there is no need to forward statuses, we just need to check that they're always OK - so I've simplified it.

* Renamed some long test names (I'm a bit allergic).

PR #594, opened 2023-10-05
Slab pool ownership guard

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/582

# What

* Chose to use a struct for the layout of the slot when supplying memory to user.
  * Owner is first member, then canary guard, then zero-length array data.
    * The actual data follows, then another canary guard.
* Added new flag.
* Updated logic to set/check owner.
* Some compile time enums/structs for sizeof AlignMax and aligning size based on that.
  * This is so that embedded storage size is easily calculated, with the size per slot in the impl class.

# Testing

* Added tests.

### DIFF
--- a/src/internal_modules/roc_core/pool.h
+++ b/src/internal_modules/roc_core/pool.h
@@ -116,7 +116,7 @@ public:
     }
 
 private:
-    AlignedStorage<EmbeddedCapacity * PoolImpl::CalculateSlotSize<sizeof(T)>::value>
+    AlignedStorage<EmbeddedCapacity * PoolImpl::SlotSize<sizeof(T)>::value>
         embedded_data_;
     PoolImpl impl_;
 };
--- a/src/internal_modules/roc_core/pool.h
+++ b/src/internal_modules/roc_core/pool.h
@@ -115,8 +115,12 @@ public:
     }
 
 private:
-    AlignedStorage<EmbeddedCapacity * PoolImpl::SlotSize<sizeof(T)>::value>
-        embedded_data_;
+    enum {
+        SlotSize = (sizeof(PoolImpl::SlotHeader) + sizeof(PoolImpl::SlotCanary)
+                    + sizeof(T) + sizeof(PoolImpl::SlotCanary) + sizeof(AlignMax) - 1)
+            / sizeof(AlignMax) * sizeof(AlignMax)
+    };
+    AlignedStorage<EmbeddedCapacity * SlotSize> embedded_data_;
     PoolImpl impl_;
 };
 
--- a/src/internal_modules/roc_core/pool.h
+++ b/src/internal_modules/roc_core/pool.h
@@ -109,8 +109,7 @@ public:
     }
 
 private:
-    AlignedStorage<EmbeddedCapacity*(sizeof(T) + PoolImpl::CanarySize
-                                     + PoolImpl::CanarySize)>
+    AlignedStorage<EmbeddedCapacity * PoolImpl::CalculateSlotSize<sizeof(T)>::value>
         embedded_data_;
     PoolImpl impl_;
 };
--- a/src/internal_modules/roc_core/pool.h
+++ b/src/internal_modules/roc_core/pool.h
@@ -25,13 +25,12 @@ namespace core {
 
 //! Memory pool flags.
 enum PoolFlags {
-    PoolFlag_PanicOnOverflow = (1 << 0),         //!< Panic when buffer overflow detected.
-    PoolFlag_PanicOnInvalidOwnership = (1 << 1), //!< Panic when invalid ownership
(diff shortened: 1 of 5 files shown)


src/internal_modules/roc_core/pool_impl.h
```diff
@@ -30,10 +30,39 @@ namespace core {
...
+            slot_size_ = MaxAlignedSize<sizeof(Slot)>::value,
+            user_slot_size_ = MaxAlignedSize<sizeof(UserSlot)>::value
+                + MaxAlignedSize<ObjectSize>::value + PoolImpl::CanarySize,
+        };
+
+    public:
+        enum {
+            value = user_slot_size_ > slot_size_ ? user_slot_size_ : slot_size_,
+        };
+    };
```
gavv, line 64: Seems these helpers (along with MaxAlignedSize and AlignMaxSize) are getting out of control and become a bit too complicated. I'd prefer to keep things simple here.

What do you think about the following approach?

In PoolImpl, define two types:

```cpp
class PoolImpl {
public:
    ...

    struct SlotHeader {
        PoolImpl* owner;
        AlignMax data[];
    };

    typedef AlignMax SlotCanary;
};
```

In PoolImpl methods, use SlotHeader struct for ownership check, and compute pointers to SlotCanary (before and after) manually, as you did it before this PR.

Then in Pool class, you can use relatively simple code:

```cpp
enum {
    // header + canary + payalod + canary, maximum-aligned
    SlotSize = (sizeof(PoolImpl::SlotHeader) + sizeof(PoolImpl::SlotCanary)*2 + sizeof(T)
                + sizeof(AlignMax) - 1)
        / sizeof(AlignMax) * sizeof(AlignMax)
};

AlignedStorage<EmbeddedCapacity * SlotSize> embedded_data_;
```

This way, MaxAlignedSize<T>, AlignMaxSize, and SlotSize<T> can be removed.

And also before- and after- canaries will be handled uniformly.

src/internal_modules/roc_core/pool_impl.h
```diff
@@ -30,10 +30,39 @@ namespace core {
...
+    struct Slot : ListNode {};
+
+    // Slot layout when supplying data to user.
+    struct UserSlot {
+        PoolImpl* owner;
+        AlignMax canary_before;
+        AlignMax data[];
+        // Another canary immediately after actual data. Any padding for data is
+        // considered part of canary.
+    };
```
gavv, line 44: nitpick: private definitions should go after public.

I'm always forgetting to add it to code style doc, but the order is usually:

* public
  * types & constants
  * methods
* private
  * types & constants
  * methods
  * fields 

This is not super-strict, but preferred for consistency.

gavv (commented):
Thanks for the patch!

src/internal_modules/roc_core/pool.h
```diff
@@ -25,11 +25,13 @@ namespace core {
 
 //! Memory pool flags.
 enum PoolFlags {
-    PoolFlag_PanicOnOverflow = (1 << 0), //!< Panic when buffer overflow detected.
+    PoolFlag_PanicOnOverflow = (1 << 0),         //!< Panic when buffer overflow detected.
+    PoolFlag_PanicOnInvalidOwnership = (1 << 1), //!< Panic when invalid ownership
```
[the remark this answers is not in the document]
nolan-veed: OK. Will do. Makes sense.

src/internal_modules/roc_core/pool_impl.h
nolan-veed: Will fix. It was because I needed the definitions before the complicated SlotSize.

src/internal_modules/roc_core/pool_impl.h
nolan-veed: Yes. It is looking complicated now. Will go for new approach.

gavv:
CI is failing (doxygen). BTW if you install doxygen and pass --enable-doxygen to scons, you'll be able to catch these issues locally. See https://roc-streaming.org/toolkit/docs/building/developer_cookbook.html

nolan-veed:
Sorry about that.

gavv:
LGTM, thanks!

gavv:
Here is a small follow-up commit: 25e4d98307d43b28e7e28e8f364bf742ef1e0609

* improved comments
* fixed calculation of slot_size_: moved max() call inside align_max() call (needed in potential case when Slot is not max-aligned)
* fixed calculation of object_size_padding_: take into account possible alignment for SlotHeader (needed in potential case when SlotHeader os not max-aligned)

I think in practice both Slot and SlotHeader will be max-aligned, but since it's not guaranteed and explicit, I fixed the code to handle it.

Issue #426, opened 2020-09-21
Make fast_random() lock-free


gavv:
_Last revised: Oct 2023_

Currently, [core::fast_random()](https://github.com/roc-streaming/roc-toolkit/blob/da0383488b087e632d6fbcf9bd99be00670173c0/src/modules/roc_core/target_posix/roc_core/fast_random.cpp#L38) is implemented using pthread_once + pthread_mutex. This approach works, but creates an additional contention point between real-time threads, which is better to be avoided.

Another problem with current implementation is that it uses nrand48(), which is not completely thread-safe (see comments in the source code).

To improve this, we can adopt a simple atomic-based lock-free PRNG algorithm. As a bonus, it will be fully portable, instead of POSIX-specific (given that we already have portable atomics in roc_core).

The implementation can be very simple. For example, we can use `mulberry32` from [this SO answer](https://stackoverflow.com/questions/17035441/looking-for-decent-quality-prng-with-only-32-bits-of-state/52056161#52056161) and replace `state += 0x6D2B79F5` with an atomic add-and-fetch. Here is a real-world example of lock-free PRNG: [lfds710_prng.h](https://github.com/liblfds/liblfds/blob/master/liblfds/liblfds7.1.0/liblfds710/inc/liblfds710/lfds710_prng.h).

See #420 for background.

GoldenBug:
In your issue description you mention that the current implementation uses both pthread_once + pthread_mutex, however in your linked file I only see your usage of pthread_mutex.

Is that the correct branch/commit/file?
Or do I misunderstand something when you say the current implementation uses pthread_once.

gavv:
@GoldenBug You're right, my bad. We use only mutex; pthread_once was employed in #420, but it wasn't merged.

roopmukherjee:
Will a distribution work? Perhaps similar to https://stackoverflow.com/a/21238187.

stevenzhu94:
Hi all. @gavv the current implementation returns uniformly distributed pseudo random numbers, so the solution mentioned by @roopmukherjee would be a good replacement, no?

gavv:
@roopmukherjee @stevenzhu94

Using C++11 random would require to have separate implementation for old non-C++11 compilers (which we still support), and I'd like to avoid this. Also, I'm not sure if C++11 thread_local is lock-free, are you aware of that?

[...]

Issue #579, opened 2023-09-29
Extract HashmapImpl class from Hashmap class


gavv:
[core::Hashmap](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/hashmap.h) implements intrusive hashmap with incremental rehashing. It is rather big template class.

To reduce code size and compilation times, it would be great to extract most of its internals to a non-template implementation class. All existing code will continue using Hashmap<T>, however its methods will mostly just invoke similar methods of HashmapImpl and make some type casts. This way, the template part will be small, and non-template part will be compiled once and reused.

We use the same approach with [core::Pool](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/pool.h) and [core::PoolImpl](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/pool_impl.h).

gavv:
Thanks!

PR #596, opened 2023-10-06
Implement lock-free fast_random function

ForeverASilver:
Commit for #426

- nrand48() is replaced with splitmix32()
- obtaining the seed is now an atomic fetch and add
- replaced pthread_mutex_lock and unlock with single pthread_once() call to an state initialization function for the seed

### DIFF
--- a/src/internal_modules/roc_audio/packetizer.cpp
+++ b/src/internal_modules/roc_audio/packetizer.cpp
@@ -34,10 +34,11 @@ Packetizer::Packetizer(packet::IWriter& writer,
     , payload_size_(payload_encoder.encoded_byte_count(samples_per_packet_))
     , packet_pos_(0)
     , valid_(false) {
-    source_ = (packet::stream_source_t)core::fast_random(0, packet::stream_source_t(-1));
-    seqnum_ = (packet::seqnum_t)core::fast_random(0, packet::seqnum_t(-1));
-    stream_ts_ =
-        (packet::stream_timestamp_t)core::fast_random(0, packet::stream_timestamp_t(-1));
+    source_ =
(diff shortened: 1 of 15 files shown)


gavv (commented):
Thanks for PR! Here are a few comments.

gavv:
Not sure if it's ready for review, but I checked new commits, the new version looks good!

I think this CI job failed because you have different version of clang-format. https://github.com/roc-streaming/roc-toolkit/actions/runs/6436815832/job/17480913363?pr=596

Which one do you use? I'm using 14.0.6. I think you can rollback changes in hashmap.h and check-formatting job will pass.

Other CI jobs failed because fast_random() is missing doxygen comment, please add it.

Please ping me when CI is fixed and PR is ready for review.

ForeverASilver:
Weirdly enough formatting is still somehow an issue and i also have clang-format 14.0.6

gavv:
I've pushed commit to this PR's branch that rolls back formatting: 671b26d8305af7d9d74cf8a854f9dd58d1bda84d

Here is how you can run the same version of clang-format that is used on CI:

```
docker run -t --rm -u "${UID}" -v "${PWD}:${PWD}" -w "${PWD}" rocstreaming/env-ubuntu scons -Q fmt
```

it should produce the same results that are expected on CI.

Ideally we need to investigate what's the problem and maybe adjust hashmap code to be unaffected by it, but we have ongoing issue that also touches hashmap (#579) and I'd like to avoid creating conflicts.

So for now, if you will work on something else in roc, you can use the docker command above. We can return to this problem later if needed.

---

Thanks for the patch! I'll merge it after CI passes.

PR #597, opened 2023-10-06
Heap arena canary guards

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/583

# What

* Changed allocate/deallocate to have canary around memory returned to user.
  * Canary after memory includes padding for memory.
* Added flag to enable testing.
* Added test.

# Testing

Added tests.

### DIFF
--- a/src/internal_modules/roc_core/heap_arena.cpp
+++ b/src/internal_modules/roc_core/heap_arena.cpp
@@ -47,12 +47,10 @@ size_t HeapArena::num_guard_failures() const {
 void* HeapArena::allocate(size_t size) {
     num_allocations_++;
 
-    size_t size_aligned = AlignOps::align_max(size);
-    size_t total_size =
-        sizeof(ChunkHeader) + sizeof(ChunkCanary) + size_aligned + sizeof(ChunkCanary);
-    size_t size_padding = size_aligned - size;
+    size_t chunk_size =
+        sizeof(ChunkHeader) + sizeof(ChunkCanary) + size + sizeof(ChunkCanary);
 
-    ChunkHeader* chunk = (ChunkHeader*)malloc(total_size);
+    ChunkHeader* chunk = (ChunkHeader*)malloc(chunk_size);
 
     char* canary_before = (char*)chunk->data;
     char* memory = (char*)chunk->data + sizeof(ChunkCanary);
@@ -60,7 +58,7 @@ void* HeapArena::allocate(size_t size) {
 
     MemoryOps::prepare_canary(canary_before, sizeof(ChunkCanary));
     MemoryOps::poison_before_use(memory, size);
-    MemoryOps::prepare_canary(canary_after, size_padding + sizeof(ChunkCanary));
+    MemoryOps::prepare_canary(canary_after, sizeof(ChunkCanary));
 
     chunk->size = size;
 
@@ -82,8 +80,6 @@ void HeapArena::deallocate(void* ptr) {
         ROC_CONTAINER_OF((char*)ptr - sizeof(ChunkCanary), ChunkHeader, data);
 
     size_t size = chunk->size;
-    size_t size_aligned = AlignOps::align_max(size);
-    size_t size_padding = size_aligned - size;
 
     char* canary_before = (char*)chunk->data;
     char* memory = (char*)chunk->data + sizeof(ChunkCanary);
@@ -92,7 +88,8 @@ void HeapArena::deallocate(void* ptr) {
     const bool canary_before_ok =
         MemoryOps::check_canary(canary_before, sizeof(ChunkCanary));
     const bool canary_after_ok =
-        MemoryOps::check_canary(canary_after, size_padding + sizeof(ChunkCanary));
+        MemoryOps::check_canary(canary_after, sizeof(ChunkCanary));
+
     if (!canary_before_ok || !canary_after_ok) {
         num_guard_failures_++;
         if (AtomicOps::load_seq_cst(flags_) & HeapArenaFlag_EnableGuards) {
--- a/src/internal_modules/roc_core/heap_arena.cpp
+++ b/src/internal_modules/roc_core/heap_arena.cpp
@@ -42,13 +42,24 @@ size_t HeapArena::num_allocations() const {
 void* HeapArena::allocate(size_t size) {
     num_allocations_++;
 
-    Chunk* chunk = (Chunk*)malloc(sizeof(Chunk) + size);
+    size_t total_size = (sizeof(Chunk) + sizeof(ChunkCanary) + size + sizeof(ChunkCanary)
+                         + sizeof(AlignMax) - 1)
+        / sizeof(AlignMax) * sizeof(AlignMax);
+    size_t size_padding = AlignOps::align_max(size) - size;
+
+    Chunk* chunk = (Chunk*)malloc(total_size);
+
+    char* canary_before = (char*)chunk->data;
+    char* memory = (char*)chunk->data + sizeof(ChunkCanary);
+    char* canary_after = (char*)chunk->data + sizeof(ChunkCanary) + size;
+
+    MemoryOps::prepare_canary(canary_before, sizeof(ChunkCanary));
+    MemoryOps::poison_before_use(memory, size);
+    MemoryOps::prepare_canary(canary_after, size_padding + sizeof(ChunkCanary));
 
     chunk->size = size;
 
-    MemoryOps::poison_before_use(chunk->data, size);
-
-    return chunk->data;
+    return memory;
 }
 
 void HeapArena::deallocate(void* ptr) {
@@ -62,9 +73,23 @@ void HeapArena::deallocate(void* ptr) {
         roc_panic("heap arena: unpaired deallocate");
     }
 
-    Chunk* chunk = ROC_CONTAINER_OF(ptr, Chunk, data);
+    Chunk* chunk = ROC_CONTAINER_OF((char*)ptr - sizeof(ChunkCanary), Chunk, data);
 
-    MemoryOps::poison_after_use(chunk->data, chunk->size);
+    size_t size = chunk->size;
+    size_t size_padding = AlignOps::align_max(size) - size;
+
+    char* canary_before = (char*)chunk->data;
+    char* memory = (char*)chunk->data + sizeof(ChunkCanary);
+    char* canary_after = (char*)chunk->data + sizeof(ChunkCanary) + size;
+
+    bool canary_before_ok = MemoryOps::check_canary(canary_before, sizeof(ChunkCanary));
+    bool canary_after_ok =
+        MemoryOps::check_canary(canary_after, size_padding + sizeof(ChunkCanary));
+    if (!canary_before_ok || !canary_after_ok) {
+        roc_panic("heap arena: buffer overflow detected");
+    }
+
+    MemoryOps::poison_after_use(memory, chunk->size);
 
     free(chunk);
 }
--- a/src/internal_modules/roc_core/heap_arena.cpp
+++ b/src/internal_modules/roc_core/heap_arena.cpp
@@ -18,8 +18,10 @@ namespace core {
 
 int HeapArena::enable_leak_detection_ = false;
 
-HeapArena::HeapArena()
-    : num_allocations_(0) {
+HeapArena::HeapArena(size_t flags)
+    : num_allocations_(0)
+    , flags_(flags)
+    , num_guard_failures_(0) {
 }
 
 HeapArena::~HeapArena() {
@@ -86,7 +88,10 @@ void HeapArena::deallocate(void* ptr) {
     bool canary_after_ok =
         MemoryOps::check_canary(canary_after, size_padding + sizeof(ChunkCanary));
     if (!canary_before_ok || !canary_after_ok) {
-        roc_panic("heap arena: buffer overflow detected");
+        num_guard_failures_++;
+        if ((flags_ & HeapArenaFlag_EnableGuards) != 0) {
+            roc_panic("heap arena: buffer overflow detected");
+        }
     }
 
     MemoryOps::poison_after_use(memory, chunk->size);
@@ -94,5 +99,8 @@ void HeapArena::deallocate(void* ptr) {
     free(chunk);
 }
 
+size_t HeapArena::num_guard_failures() const {
+    return num_guard_failures_;
+}
 } // namespace core
 } // namespace roc
--- a/src/internal_modules/roc_core/heap_arena.cpp
+++ b/src/internal_modules/roc_core/heap_arena.cpp
@@ -40,6 +40,10 @@ size_t HeapArena::num_allocations() const {
     return (size_t)num_allocations_;
 }
 
+size_t HeapArena::num_guard_failures() const {
+    return num_guard_failures_;
+}
+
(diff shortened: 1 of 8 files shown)


src/internal_modules/roc_core/heap_arena.cpp
```diff
@@ -42,13 +44,24 @@ size_t HeapArena::num_allocations() const {
 void* HeapArena::allocate(size_t size) {
     num_allocations_++;
 
-    Chunk* chunk = (Chunk*)malloc(sizeof(Chunk) + size);
+    size_t total_size = (sizeof(Chunk) + sizeof(ChunkCanary) + size + sizeof(ChunkCanary)
+                         + sizeof(AlignMax) - 1)
+        / sizeof(AlignMax) * sizeof(AlignMax);
+    size_t size_padding = AlignOps::align_max(size) - size;
```
gavv, line 50: There is no need to apply alignment or padding here.

In pool, single slab was containing multiple slots placed continuously. If the size of a slot is not max-aligned, next slot will start from non-aligned address, and in result its data will be non-aligned.

In heap arena, we don't have this problem. Each chunk does its own call to malloc(). malloc() always returns max-aligned memory, and Chunk::data is of type AlignMax; because of this, data is guaranteed to be aligned.

src/internal_modules/roc_core/heap_arena.h
```diff
@@ -45,15 +58,22 @@ class HeapArena : public IArena, public NonCopyable<> {
     //! Deallocate previously allocated memory.
     virtual void deallocate(void*);
 
+    //! Get number of guard failures.
+    size_t num_guard_failures() const;
+
 private:
     struct Chunk {
```
gavv, line 65: I suggest to rename it to `ChunkHeader` for consistency with pool.

src/internal_modules/roc_core/heap_arena.cpp
```diff
@@ -62,12 +75,32 @@ void HeapArena::deallocate(void* ptr) {
...
 
-    MemoryOps::poison_after_use(chunk->data, chunk->size);
+    char* canary_before = (char*)chunk->data;
+    char* memory = (char*)chunk->data + sizeof(ChunkCanary);
+    char* canary_after = (char*)chunk->data + sizeof(ChunkCanary) + size;
+
+    bool canary_before_ok = MemoryOps::check_canary(canary_before, sizeof(ChunkCanary));
+    bool canary_after_ok =
+        MemoryOps::check_canary(canary_after, size_padding + sizeof(ChunkCanary));
```
gavv, line 89:
```suggestion
    const bool canary_before_ok = MemoryOps::check_canary(canary_before, sizeof(ChunkCanary));
    const bool canary_after_ok =
        MemoryOps::check_canary(canary_after, size_padding + sizeof(ChunkCanary));
```

src/internal_modules/roc_core/heap_arena.cpp
```diff
@@ -62,12 +75,32 @@ void HeapArena::deallocate(void* ptr) {
...
+    if (!canary_before_ok || !canary_after_ok) {
+        num_guard_failures_++;
+        if ((flags_ & HeapArenaFlag_EnableGuards) != 0) {
+            roc_panic("heap arena: buffer overflow detected");
+        }
```
gavv, line 94:
```suggestion
        if (flags_ & HeapArenaFlag_EnableGuards) {
            roc_panic("heap arena: detected memory violation: ok_before=%d ok_after=%d",
                      (int)canary_before_ok, (int)canary_after_ok);
        }
```

(same as here: https://github.com/roc-streaming/roc-toolkit/pull/594#issuecomment-1750704681)

src/internal_modules/roc_core/heap_arena.h
```diff
@@ -30,7 +39,11 @@ namespace core {
 //! The memory is always maximum aligned. Thread-safe.
 class HeapArena : public IArena, public NonCopyable<> {
 public:
-    HeapArena();
+    //! Initialize.
+    //!
+    //! @b Parameters
+    //! - @p flags defines options to modify behaviour as indicated in HeapArenaFlags
+    HeapArena(size_t flags = DefaultHeapArenaFlags);
```
gavv, line 46: In HeapArena, we have static method enable_leak_detection() which enabled panic on leak for all arena instances. (This method is used to enable panics globally in tools and tests, but to disable it in C library.)

With this constructor, we now have two source of configuration for arena: one via static method, another via ctor flags.

To keep things simpler, I'd suggest to keep only one:

* remove flags from ctor
* replace static method enable_leak_detection() with static method set_flags() or configure() or something like this, which accepts flags
* define two flags: EnableGuards and EnableLeakDetection
* initially flags are set to EnableGuards
* in places where we called enable_leak_detection(), we will now set flags EnableGuards | EnableLeakDetection
* in tests we can temporary change flags and then change them back

gavv (commented):
Thanks for PR!

src/internal_modules/roc_core/heap_arena.h
nolan-veed: Along with the above... I've kept the DefaultHeapArenaFlags, and used that all over as well. Let me know.

src/internal_modules/roc_core/heap_arena.cpp
```diff
@@ -16,10 +16,13 @@
 namespace roc {
 namespace core {
 
-int HeapArena::enable_leak_detection_ = false;
+int HeapArena::enable_leak_detection_ =
+    (DefaultHeapArenaFlags & HeapArenaFlag_EnableLeakDetection) != 0;
+int HeapArena::enable_guards_ = (DefaultHeapArenaFlags & HeapArenaFlag_EnableGuards) != 0;
```
nolan-veed, line 21: Can these be bools? Is there any build problems when using AtomicOps on bools?

gavv: > Can these be bools?

Probably not on all platforms where we use libatomic_ops instead of C11 __atomic funcs.

In this case it's better to use a single atomic that will hold all flags, instead of seperate atomic for each flag.

src/internal_modules/roc_core/heap_arena.h
gavv: LGTM

src/internal_modules/roc_core/heap_arena.cpp
gavv: So the new code is:

```cpp
    size_t size_aligned = AlignOps::align_max(size);
    size_t total_size =
        sizeof(ChunkHeader) + sizeof(ChunkCanary) + size_aligned + sizeof(ChunkCanary);
    size_t size_padding = size_aligned - size;
```

But why do you think we need align_max() and size_padding?

My understanding is that we need just:

```cpp
size_t total_size =
        sizeof(ChunkHeader) + sizeof(ChunkCanary) + size + sizeof(ChunkCanary);
```

src/internal_modules/roc_core/heap_arena.h
```diff
@@ -20,6 +20,17 @@
 namespace roc {
 namespace core {
 
+//! Heap arena flags.
+enum HeapArenaFlags {
+    //! Enable leak detection, etc.
+    HeapArenaFlag_EnableLeakDetection = (1 << 0),
+    //! Enable guards such as canary, etc.
+    HeapArenaFlag_EnableGuards = (1 << 1),
```
gavv, line 28:
```suggestion
    //! Enable panic if leaks detected in arena destructor.
    HeapArenaFlag_EnableLeakDetection = (1 << 0),
    //! Enable panic if memory violation detected when deallocating chunk.
    HeapArenaFlag_EnableGuards = (1 << 1),
```

src/tests/roc_core/test_heap_arena.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "roc_core/memory_ops.h"
+
+namespace roc {
+namespace core {
+
+TEST_GROUP(heap_arena) {};
+
+TEST(heap_arena, guard_object) {
+    core::HeapArena::set_flags(core::DefaultHeapArenaFlags
+                               & ~core::HeapArenaFlag_EnableGuards);
```
gavv, line 21: nitpick: we can move setting and restoring flags to setup() and teardown() methods of TEST_GROUP. This way we can be sure that flags are restored no matter if test succeeded or failed.

src/internal_modules/roc_core/heap_arena.h
```diff
@@ -45,15 +60,22 @@ class HeapArena : public IArena, public NonCopyable<> {
     //! Deallocate previously allocated memory.
     virtual void deallocate(void*);
 
+    //! Get number of guard failures.
+    size_t num_guard_failures() const;
```
gavv, line 64: nitpick: let's move this method right after num_allocations(), another counter getter used in tests.

src/internal_modules/roc_core/heap_arena.cpp
nolan-veed: My thinking...
Since we need size_aligned, in order to calculate the size_padding, why not use that in the total_size.

Also, I'm not sure, but some malloc libraries might actually track this to see if we are accessing outside the size requested?

gavv: Why do we need size_aligned and size_padding?

My point is that padding is not needed in HeapArena at all.

Instead of:

```cpp
MemoryOps::prepare_canary(canary_after, size_padding + sizeof(ChunkCanary));
```

we can use:

```
MemoryOps::prepare_canary(canary_after, size + sizeof(ChunkCanary));
```

gavv: > Also, I'm not sure, but some malloc libraries might actually track this to see if we are accessing outside the size requested?

That's true. There are a lot of different tools. E.g. on MSVC you can enable guard pages around each allocations. On glibc there is MALLOC_CHECK_, and in clang and gcc there are sanitizers (which we support via --sanitizers scons option).

These checks are toolchain-dependent and enabled conditionally. Often they should be enabled at build time and can either affect performance and/or require dependencies to be built with same flags.

In this issue we implement minimalistic portable version of such check that will be always enabled, regardless of platform, build flags, etc, and hopefully will help to detect certain class of bugs.

And we will also continue using more serious checks as well. For example, on CI we have jobs that run tests under sanitizers and valgrind.

gavv: > Instead of: [...]

Sorry, the code I provided is wrong. Here is what I mean:

```cpp
    const size_t chunk_size =
        sizeof(ChunkHeader) + sizeof(ChunkCanary) + size + sizeof(ChunkCanary);

    ChunkHeader* chunk = (ChunkHeader*)malloc(chunk_size);

    char* canary_before = (char*)chunk->data;
    char* memory = (char*)chunk->data + sizeof(ChunkCanary);
    char* canary_after = (char*)chunk->data + sizeof(ChunkCanary) + size;

    MemoryOps::prepare_canary(canary_before, sizeof(ChunkCanary));
    MemoryOps::poison_before_use(memory, size);
    MemoryOps::prepare_canary(canary_after, sizeof(ChunkCanary));
```

i.e., there is no need to pad `size` to make it max-aligned. Because of reasons I mentioned here: https://github.com/roc-streaming/roc-toolkit/pull/597#discussion_r1348795259

src/tests/roc_core/test_heap_arena.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace core {
+
+TEST_GROUP(heap_arena) {
+    void setup() { core::HeapArena::set_flags(core::DefaultHeapArenaFlags
+                                              & ~core::HeapArenaFlag_EnableGuards);
+}
+void teardown() {
+    core::HeapArena::set_flags(core::DefaultHeapArenaFlags
+                               | core::HeapArenaFlag_EnableGuards);
+}
```
gavv, line 24: Seems clang-format broke formatting.

Please use

```
// clang-format off
// clang-format on
```

as we do it in `src/tests/roc_core/test_printer.cpp`

src/tests/roc_core/test_heap_arena.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace roc {
+namespace core {
+
+TEST_GROUP(heap_arena) {
+    void setup() { core::HeapArena::set_flags(core::DefaultHeapArenaFlags
+                                              & ~core::HeapArenaFlag_EnableGuards);
+}
+void teardown() {
+    core::HeapArena::set_flags(core::DefaultHeapArenaFlags
+                               | core::HeapArenaFlag_EnableGuards);
```
gavv, line 23: In teardown() it can be just DefaultHeapArenaFlags I guess.

src/internal_modules/roc_core/heap_arena.cpp
nolan-veed: In PoolImpl, after the data, we use prepare_canary() and check_canary() with padding bytes + canary-size. So, for non-max-aligned data, we'll always check a few more byets. I was going for something similar here.

But, let me know. I'll do what you suggested.

gavv: Yeah,

In pool, we need to add padding to object to ensure that **slot size** of multiple of max-align bytes.

If beginning of first slot is max-aligned, and size of each slot is multiple of max-align bytes, it is then guaranteed that beginning of **each** slot is max-aligned.

In other words, we need to pad objects because we split continuous chunk into multiple slots and need to align each of them.

However, in arena, there are no multiple slots. We need to ensure that beginning of user data is max-aligned. But we don't care if size of user data is multiple of max-align bytes, because we don't place anything after user data.

Well, we place canary guard there, but we don't care about its alignment.

We need max-aligning for user data because we don't know its structure. Maybe user data is begins with uint64_t, which requires to be 8-byte aligned on some CPUs, and otherwise attempt to access it will cause SIGBUS.

But our canary guard is accessed via MemoryOps; it interprets canary as bytes and does not place any requirements on alignment of canary.

Hence, in arena we can place trailing canary right after user data and don't care about padding. This will simplify implementation a bit and we'll avoid unnecessary calculations.

src/tests/roc_core/test_heap_arena.cpp
```diff
@@ -0,0 +1,75 @@
...
+    void setup() {
+        core::HeapArena::set_flags(core::DefaultHeapArenaFlags
+                                    & ~core::HeapArenaFlag_EnableGuards);
+    }
+    void teardown() {
+        core::HeapArena::set_flags(core::DefaultHeapArenaFlags
+                                    | core::HeapArenaFlag_EnableGuards);
+    }
+// clang-format on
+}; // namespace roc
```
gavv, line 28:
```suggestion
// clang-format off
TEST_GROUP(heap_arena) {
    void setup() {
        core::HeapArena::set_flags(core::DefaultHeapArenaFlags
                                    & ~core::HeapArenaFlag_EnableGuards);
    }
    void teardown() {
        core::HeapArena::set_flags(core::DefaultHeapArenaFlags);
    }
};
// clang-format on
```

gavv: * Formatting is still broken. Here is a fix.
* In teardown(), we can use just DefaultHeapArenaFlags, without `| core::HeapArenaFlag_EnableGuards` part.

nolan-veed: Yes. I should have done that.

gavv:
LGTM!

PR #604, opened 2023-10-09
Fix in SpeexResampler CTS calculations, also relaxed its tests.

baranovmv:
Only the changes we discussed today in Matrix


gavv:
LGTM!

PR #611, opened 2023-10-13
HashMap implementation extract

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/579

# What

* Carefully pulled out the code from templated class to implementation class header and cpp file.
* Implementation class has no idea/concept of ownership, etc. - so that lives in the main class.
* Implementation class has no idea about key equality comparison - so provided that as callback to main class.

# Testing

Existing tests should suffice.

### DIFF
--- a/src/internal_modules/roc_core/hashmap.h
+++ b/src/internal_modules/roc_core/hashmap.h
@@ -103,7 +103,14 @@ public:
 
     //! Release ownership of all elements.
     ~Hashmap() {
-        impl_.release_all(&Hashmap<T, EmbeddedCapacity, OwnershipPolicy>::node_release);
+        HashmapNode::HashmapNodeData* node = impl_.front();
+
+        while (node != NULL) {
+            impl_.remove(node);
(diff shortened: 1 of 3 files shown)


nolan-veed:
I'll fix the builds... wanted some early feedback.

gavv:
Thanks for PR!

I'll do full review in upcoming days, but overall approach looks fine.

I think we can avoid node_release_callback: instead of iterating over elements in HashmapImpl::release_bucket_array_ using buckets, we can iterate them in template class destructor using front/back/nextof and release ownership in template class. Then implementation class will just deallocate buckets. After releasing ownership, and before deallocating buckets, buckets will contains dangling pointers to released elements, but that's not a problem because we're anyway in destructor.

nolan-veed:
> 
> I think we can avoid node_release_callback: ...

I'll do that... it's better.

gavv:
Small follow-up: de830bd2aef0433edbd760e08dc9c3f44f05228f

PR #612, opened 2023-10-14
add script for generating enums in bindings repo (java, go)

### LINKED ISSUE #574: Create Python script that generates Go and Java files for enum declarations
We have a [C library](https://roc-streaming.org/toolkit/docs/api/reference.html) and [Go](https://github.com/roc-streaming/roc-go) and [Java](https://github.com/roc-streaming/roc-java/) bindings for it.

Bindings are created manually. This is mostly fine and intended.

However, there is one place where we would like to use code generation instead. C library defines quite a lot of enums, and these enums are replicated one-to-one in bindings.

**It would be nice to implement a script that automatically parses enums in C library and generates corresponding Go and Java files.**

The easiest way to parse C declarations is to run doxygen build, which will produce XML file with all C API declarations. Then we can find enum declarations in that file.

Enums are declared in these headers:

* https://github.com/roc-streaming/roc-toolkit/blob/develop/src/public_api/include/roc/config.h
* https://github.com/roc-streaming/roc-toolkit/blob/develop/src/public_api/include/roc/log.h

After building documentation, as [described here](https://roc-streaming.org/toolkit/docs/building/developer_cookbook.html#building-documentation), Doxygen-generated XML files will be located in `build/docs/public_api/xml`.

For example, there is `roc_format` enum with `ROC_FORMAT_PCM_FLOAT32` value, and here is corresponding nodes in XML:

```xml
       <memberdef kind="enum" id="config_8h_1ac24a6637fefa48d61bb8631c25bde919" prot="public" static="n     o" strong="no">
         <type></type>
         <name>roc_format</name>
         <enumvalue id="config_8h_1ac24a6637fefa48d61bb8631c25bde919aa851fa1a679b2df37b113fefe53b98e1"      prot="public">
           <name>ROC_FORMAT_PCM_FLOAT32</name>
           <initializer>= 1</initializer>
           <briefdescription>
 <para>PCM floats. </para>
           </briefdescription>
           <detaileddescription>
 <para>Uncompressed samples coded as 32-bit native-endian floats in range [-1; 1]. Channels are interle     aved, e.g. two channels are encoded as &quot;L R L R ...&quot;. </para>
           </detaileddescription>
         </enumvalue>
```

And here is what we should generate (kind of):

* https://github.com/roc-streaming/roc-go/blob/main/roc/config.go
* https://github.com/roc-streaming/roc-java/blob/main/src/main/java/org/rocstreaming/roctoolkit/FecEncoding.java

In go, all enums are currently in same file (`config.go`), and in java, each enum is in its own file. Currently they're updated manually. And now we want to replace this hand-written definitions with generated code.

Generator should produce something similar to what exists currently. It should:

* follow same naming conventions (e.g. snake_case it converted to CamelCase for enum names), 
* preserve enum values equal to value in C library
* preserve comments

Since one class per file is preferred for Java, for simplicity we can use same approach for Go as well.

We should keep in mind that in future more bindings may be added, and it should be feasible to extend generator to support more output languages.

We use Python for scripting, so let's use it for generator as well. The script should be added here: https://github.com/roc-streaming/roc-toolkit/tree/develop/scripts

We'll need 3 pull requests:

* PR to `roc-toolkit` repo that adds generator script
* PRs to `roc-go` and `roc-java` repos that replace hand-written enum declarations with generated

ortex:
fix #574 

generation results:
https://github.com/roc-streaming/roc-java/pull/127
https://github.com/roc-streaming/roc-go/pull/117 (splitted into 2 commits for better diff: move enums to separated files, then run script)


scripts/generate_enums.py
```diff
@@ -0,0 +1,356 @@
...
+ROC_JAVA_PACKAGE = "org.rocstreaming.roctoolkit"
+
+ODD_PREFIXES = {'roc_clock_source': 'ROC_CLOCK_', 'roc_protocol': 'ROC_PROTO_'}
+
+
+def to_camel_case(name):
+    return ''.join(x.capitalize() for x in name.split('_'))
+
+
+class DocComment:
```
gavv, line 22: nit: consider using namedtuple or dataclass for such types

scripts/generate_enums.py
```diff
@@ -0,0 +1,356 @@
...
+class EnumDefinition:
+    def __init__(self, name: string, values: list[EnumValue], doc: DocComment):
+        self.name: string = name
+        self.values: list[EnumValue] = values
+        self.doc: DocComment = doc
+
+
+class EnumGenerator:
+    def generate_enum(self, enum_definition: EnumDefinition):
+        pass
```
gavv, line 43: nit: convention is to raise NotImplementedError

scripts/generate_enums.py
```diff
@@ -0,0 +1,356 @@
...
+
+
+def parse_doc_elem(elem: ElementTree.Element):
+    items = []
+    tag = elem.tag
+    parse_children = True
+    text = strip_text(elem.text)
+    if tag == "para":
+        if text:
+            items.append({"type": "text", "text": text})
```
gavv, line 293: nit: consider using namedtuple or dataclass for items

scripts/generate_enums.py
```diff
@@ -0,0 +1,356 @@
...
+
+        return to_camel_case(ref_value.lower().removeprefix('roc_'))
+
+
+def parse_config_xml():
+    try:
+        tree = ElementTree.parse(CONFIG_FILE_PATH)
+        return tree.getroot()
+    except FileNotFoundError:
+        print(f"File not found: {CONFIG_FILE_PATH}")
```
gavv, line 240: nit: convention is to print errors and warnings to stderr

scripts/generate_enums.py
```diff
@@ -0,0 +1,356 @@
...
+    args = parser.parse_args()
+
+    xml_doc = parse_config_xml()
+    enum_definitions = get_enums(xml_doc)
+
+    name_prefixes = {}
+    for enum_definition in enum_definitions:
+        name = enum_definition.name
+        prefix = ODD_PREFIXES.get(name, name.upper() + "_")
+        name_prefixes[name] = prefix
```
gavv, line 346: nit: I'd move this into a function to keep main() high-level

scripts/generate_enums.py
```diff
@@ -0,0 +1,356 @@
...
+
+    if args.type == "all" or args.type == "java":
```
gavv, line 348: Before this code, I'd add checks that source file / destination dirs exist, and print error if not.

scripts/generate_enums.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def generate_enum(self, enum_definition: EnumDefinition):
+        java_name = self.get_java_enum_name(enum_definition.name)
+        enum_values = enum_definition.values
+
+        enum_file_path = (self.base_path + "/src/main/java/"
+                          + ROC_JAVA_PACKAGE.replace(".", "/") + "/" + java_name + ".java")
+        enum_file = open(enum_file_path, "w")
+
+        enum_file.write("// Code generated by generate_enums script from roc-toolkit; DO NOT EDIT!\n\n")
```
gavv, line 60: I suggest to add roc-toolkit commit SHA and git tag, for which these enums were generated.

scripts/generate_enums.py
ortex: dataclass 👍

scripts/generate_enums.py
ortex: done

scripts/generate_enums.py
ortex: done

scripts/generate_enums.py
ortex: done

scripts/generate_enums.py
ortex: done

scripts/generate_enums.py
ortex: done

scripts/generate_enums.py
ortex: done

PR #613, opened 2023-10-14
Heap arena ownership guard

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/583

# What

* Added owner pointer field to ChunkHeader.
* Set it in allocate, check in deallocate.
* Moved the num_allocations checking in deallocate, before freeing.
* Wrote test.

# Testing

There are tests.


gavv:
Thanks!

gavv:
Small fix: f407368a6362cd04c0737b62b6cb4f23f258f5d8

PR #616, opened 2023-10-14
AlignOps EndianOps cpp files

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/601

# What

* Moved non-template methods to cpp files.
* Removed inline as a result.

# Testing

There are tests.

### DIFF
--- a/src/internal_modules/roc_core/align_ops.h
+++ b/src/internal_modules/roc_core/align_ops.h
@@ -7,7 +7,7 @@
  */
 
 //! @file roc_core/align_ops.h
-//! @brief TODO.
+//! @brief Alignment operations.
 
 #ifndef ROC_CORE_ALIGN_OPS_H_
 #define ROC_CORE_ALIGN_OPS_H_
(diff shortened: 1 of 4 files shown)


src/internal_modules/roc_core/align_ops.h
```diff
@@ -7,7 +7,7 @@
  */
 
 //! @file roc_core/align_ops.h
-//! @brief Alignment operations.
+//! @brief TODO.
```
gavv, line 10: Accidental change, I guess

gavv (commented):
Thanks for PR! One small comment

src/internal_modules/roc_core/align_ops.h
gavv: Fixed.

nolan-veed: It wasn't. But, in hindsight, I shouldn't have updated that.

PR #621, opened 2023-10-20
Refactor ChannelMapper

dshil:
gh-86

Based on patch suggested in https://github.com/roc-streaming/roc-toolkit/pull/607.

### DIFF
--- a/src/internal_modules/roc_audio/channel_mapper.cpp
+++ b/src/internal_modules/roc_audio/channel_mapper.cpp
@@ -7,9 +7,6 @@
  */
 
 #include "roc_audio/channel_mapper.h"
-#include "roc_audio/channel_mapper_table.h"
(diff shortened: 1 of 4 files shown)


gavv:
Thanks for PR.

On one hand, it's nice that ChannelMapper::setup_map_matrix_() was split. On the other hand, the new approach looks over-complicated. IMHO there is now too much entities to keep track of when reading the code, and they're not just data but each one is new abstraction with its own API.

I'd prefer to keep it simpler: keep all state inside ChannelMapper, but split setup_map_matrix_() into multiple functions. I incorporated some of you refactorings (notably for building index maps) and implemented this approach here: https://github.com/roc-streaming/roc-toolkit/pull/626

Could you take a look?

gavv:
New version is better! I've pushed follow-up refactoring to #626 (I dropped my old version, but have redone some renames I did there).

PR #623, opened 2023-10-31
Follow up for #549 

baranovmv:
#539 

From #549:

In receiver_source test recv_timestamp_mapping_remixing the failure was caused by the issue in Depacketizer. As you suggested (after my localization): https://github.com/roc-streaming/roc-toolkit/issues/539#issuecomment-1712284093:
 
>   We should check if Depacketizer correctly handles case when first packet for frame had zero CTS, and second packet for the same frame had non-zero CTS.
 
>   I think it's not covered right now.

> We need a test for this and maybe a fix.

In sender_sink tests the issue is simple: Resampler backend throw away first n samples (N==latency), which was not properly reflected in test::PacketReader

### DIFF
--- a/src/tests/roc_pipeline/test_helpers/packet_reader.h
+++ b/src/tests/roc_pipeline/test_helpers/packet_reader.h
@@ -186,7 +186,7 @@ private:
                 base_ts + sample_spec.samples_per_chan_2_ns(abs_offset_);
 
             expect_capture_timestamp(capture_ts, pkt.rtp()->capture_timestamp,
-                                     TimestampEpsilon);
+                                     sample_spec, TimestampEpsilonSmpls);
         }
     }
 
--- a/src/tests/roc_pipeline/test_helpers/packet_reader.h
+++ b/src/tests/roc_pipeline/test_helpers/packet_reader.h
@@ -51,7 +51,7 @@ public:
 
     void read_packet(size_t samples_per_packet,
                      const audio::SampleSpec& sample_spec,
-                     core::nanoseconds_t base_capture_ts = 0) {
+                     core::nanoseconds_t base_capture_ts = -1) {
         packet::PacketPtr pp = read_packet_();
 
         audio::sample_t samples[MaxSamples] = {};
@@ -69,9 +69,9 @@ public:
         abs_offset_ += samples_per_packet;
     }
 
-    packet::PacketPtr read_nonzero_packet(size_t samples_per_packet,
-                                          const audio::SampleSpec& sample_spec,
-                                          core::nanoseconds_t base_capture_ts = 0) {
+    void read_nonzero_packet(size_t samples_per_packet,
+                             const audio::SampleSpec& sample_spec,
+                             core::nanoseconds_t base_capture_ts = -1) {
         packet::PacketPtr pp = read_packet_();
 
         audio::sample_t samples[MaxSamples] = {};
@@ -86,13 +86,11 @@ public:
         }
         CHECK(non_zero > 0);
         abs_offset_ += samples_per_packet;
-
-        return pp;
     }
 
     void read_zero_packet(size_t samples_per_packet,
                           const audio::SampleSpec& sample_spec,
-                          core::nanoseconds_t base_capture_ts = 0) {
+                          core::nanoseconds_t base_capture_ts = -1) {
         packet::PacketPtr pp = read_packet_();
 
         audio::sample_t samples[MaxSamples] = {};
@@ -179,7 +177,7 @@ private:
                                   core::nanoseconds_t base_ts) {
         CHECK(pkt.rtp());
 
-        if (base_ts == 0) {
+        if (base_ts < 0) {
             LONGS_EQUAL(0, pkt.rtp()->capture_timestamp);
         } else {
             const core::nanoseconds_t capture_ts =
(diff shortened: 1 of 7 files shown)


src/tests/roc_pipeline/test_sender_sink.cpp
```diff
@@ -361,7 +361,7 @@ TEST(sender_sink, timestamp_mapping) {
     packet_reader.read_eof();
 }
 
-IGNORE_TEST(sender_sink, timestamp_mapping_remixing) {
+TEST(sender_sink, timestamp_mapping_remixing_packet_reader) {
```
gavv, line 364:
```suggestion
TEST(sender_sink, timestamp_mapping_remixing) {
```

src/tests/roc_pipeline/test_sender_sink.cpp
```diff
@@ -399,9 +399,21 @@ IGNORE_TEST(sender_sink, timestamp_mapping_remixing) {
...
     for (size_t np = 0; np < ManyFrames / FramesPerPacket - 5; np++) {
-        packet_reader.read_nonzero_packet(SamplesPerPacket, packet_sample_spec,
-                                          unix_base);
+        packet::PacketPtr pp;
+        UNSIGNED_LONGS_EQUAL(status::StatusOK, queue.read(pp));
+        CHECK(pp);
+
+        if (np == 0) {
+            cts = pp->rtp()->capture_timestamp;
+            CHECK(cts >= unix_base);
```
gavv, line 410:
```suggestion
            CHECK(cts >= unix_base);
            CHECK(cts < unix_base + core::Millisecond);
```

(I tested, the check passes)

src/tests/roc_pipeline/test_helpers/packet_reader.h
gavv, line 1: I think all changes in this file, except use of TimestampEpsilonSmpls, should be reverted?

* Old version allowed to use `base_capture_ts = 0` to assert for zero CTS and `base_capture_ts = -1` to skip checking CTS. New version does not allow to assert for zero CTS.

* New signature of read_nonzero_packet() is not consistent with other methods and also it's never actually used.

gavv (commented):
Thanks for PR!

src/tests/roc_pipeline/test_helpers/packet_reader.h
baranovmv: > Old version allowed to use base_capture_ts = 0 to assert for zero CTS and base_capture_ts = -1 to skip checking CTS. New version does not allow to assert for zero CTS.

This change in `PacketReader` is needed to keep the tests green (e.g. `TEST(sender_sink, sample_rate_mapping)`). It handles the corner-case in Depacketizer:

> We should check if Depacketizer correctly handles case when first packet for frame had zero CTS, and second packet for the same frame had non-zero CTS.

gavv: I believe it's not needed now, I tried to return old behavior and all tests passed.

gavv: With this patch (on top of your branch) tests pass for me:

```diff
diff --git a/src/tests/roc_pipeline/test_helpers/packet_reader.h b/src/tests/roc_pipeline/test_helpers/packet_reader.h
index a9524394..9b3b741f 100644
--- a/src/tests/roc_pipeline/test_helpers/packet_reader.h
+++ b/src/tests/roc_pipeline/test_helpers/packet_reader.h
@@ -51,7 +51,7 @@ public:
 
     void read_packet(size_t samples_per_packet,
                      const audio::SampleSpec& sample_spec,
-                     core::nanoseconds_t base_capture_ts = 0) {
+                     core::nanoseconds_t base_capture_ts = -1) {
         packet::PacketPtr pp = read_packet_();
 
         audio::sample_t samples[MaxSamples] = {};
@@ -71,7 +71,7 @@ public:
 
     packet::PacketPtr read_nonzero_packet(size_t samples_per_packet,
                                           const audio::SampleSpec& sample_spec,
-                                          core::nanoseconds_t base_capture_ts = 0) {
+                                          core::nanoseconds_t base_capture_ts = -1) {
         packet::PacketPtr pp = read_packet_();
 
         audio::sample_t samples[MaxSamples] = {};
@@ -92,7 +92,7 @@ public:
 
     void read_zero_packet(size_t samples_per_packet,
                           const audio::SampleSpec& sample_spec,
-                          core::nanoseconds_t base_capture_ts = 0) {
+                          core::nanoseconds_t base_capture_ts = -1) {
         packet::PacketPtr pp = read_packet_();
 
         audio::sample_t samples[MaxSamples] = {};
@@ -179,7 +179,7 @@ private:
                                   core::nanoseconds_t base_ts) {
         CHECK(pkt.rtp());
 
-        if (base_ts == 0) {
+        if (base_ts < 0) {
             LONGS_EQUAL(0, pkt.rtp()->capture_timestamp);
         } else {
             const core::nanoseconds_t capture_ts =
```

Actually it seems that this patch makes no practical difference - all tests either use default value (which means "check that cts is zero") on custom positive value for cts base. None of them use "0" or "-1" explicitly.

I'd prefer to keep default value -1 though, because the same is used in test::FrameReader and test::FrameWriter. Or we can update all of them to use zero as default. I'm fine with both approaches, just would like to keep them consistent.

gavv:
I'm preparing 0.3.0 release and have rebased develop on master, please rebase PR.

I also cherry-picked depacketizer fix from this PR to develop (without tests) to make it part of the release: 55821da629f43d729928c5444d74d3deda73afb9

gavv:
Thanks!

Issue #617, opened 2023-10-16
Optimize StringList iteration


gavv:
[StringList](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/string_list.h) implements dynamically extended list of C strings.

Internally, it's just a continuous array. When you append a string to the list, it is copied to the end of array.

It supports iteration via nextof() method (similar to List::nextof and Hashmap::nextof).

Currently nextof() uses strlen() to find position of next string. I**t would be nice to avoid strlen() call for faster iteration**. We can do it by prepending string size before each string stored in list. When nextof() is called, it can lookup this size and just add it to the string pointer.

**In addition, it would be nice to add prevof() method**, that will return element previous to given one, or NULL for the first element.

ForeverASilver:
@gavv 
I can work on this task

gavv:
You're welcome!

gavv:
Merged!

PR #625, opened 2023-11-01
Add lengths to StringList

ForeverASilver:
pull request for #617 

string lengths are stored as size_t and written to the buffer just before the string.

changed tests to create new expected allocation counts to account for different size_t sizes.

any suggested changes are welcome and appreciated

### DIFF
--- a/src/internal_modules/roc_core/string_list.cpp
+++ b/src/internal_modules/roc_core/string_list.cpp
@@ -7,6 +7,7 @@
  */
 
 #include "roc_core/string_list.h"
+#include "roc_core/log.h"
 #include "roc_core/panic.h"
 
 namespace roc {
@@ -14,6 +15,7 @@ namespace core {
 
 StringList::StringList(IArena& arena)
     : data_(arena)
+    , front_(NULL)
     , back_(NULL)
     , size_(0) {
 }
@@ -24,7 +26,7 @@ size_t StringList::size() const {
 
 const char* StringList::front() const {
     if (size_) {
-        return &data_[0];
+        return front_->str;
     } else {
         return NULL;
     }
@@ -32,7 +34,7 @@ const char* StringList::front() const {
 
 const char* StringList::back() const {
     if (size_) {
-        return back_;
+        return back_->str;
     } else {
         return NULL;
     }
@@ -54,14 +56,44 @@ const char* StringList::nextof(const char* str) const {
         roc_panic("stringlist: string doesn't belong to the list");
     }
 
-    const char* ptr = str + strlen(str) + 1;
-    roc_panic_if(ptr > end);
+    const Header* str_header = ROC_CONTAINER_OF(const_cast<char*>(str), Header, str);
 
-    if (ptr == end) {
+    if (str_header == back_) {
         return NULL;
     }
 
-    return ptr;
+    const Header* next_header =
+        (const Header*)((const char*)str_header + str_header->len);
+    return next_header->str;
+}
+
+const char* StringList::prevof(const char* str) const {
+    if (str == NULL) {
+        roc_panic("stringlist: string is null");
+    }
+
+    if (size_ == 0) {
+        roc_panic("stringlist: list is empty");
+    }
+
+    const char* begin = &data_[0];
+    const char* end = &data_[0] + data_.size();
+
+    if (str < begin || str >= end) {
+        roc_panic("stringlist: string doesn't belong to the list");
+    }
+
+    const Header* str_header = ROC_CONTAINER_OF(const_cast<char*>(str), Header, str);
+
+    if (str_header == front_) {
+        return NULL;
+    }
+
+    const Footer* prev_footer = (const Footer*)((const char*)str_header - sizeof(Footer));
+    const Header* prev_header =
+        (const Header*)((const char*)str_header - prev_footer->len);
+
+    return prev_header->str;
 }
 
 void StringList::clear() {
@@ -84,7 +116,9 @@ bool StringList::push_back(const char* begin, const char* end) {
     }
 
     const size_t cur_sz = data_.size();
-    const size_t add_sz = (size_t)(end - begin) + 1;
+    const size_t str_sz = (size_t)(end - begin) + 1;
+    const size_t add_sz =
+        sizeof(Header) + AlignOps::align_as(str_sz, sizeof(Header)) + sizeof(Footer);
 
     if (!grow_(cur_sz + add_sz)) {
         return false;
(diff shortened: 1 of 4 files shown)


gavv:
Thanks for PR!

A few suggestions.

**1.** I think the could be a bit cleaner if we introduce structs for header and footer:

```cpp
class StringList {
...
private:
    struct Header {
        size_t len;
        char str[];
    };

    struct Footer {
        size_t len;
    };
```

Then we do something like this:

```cpp
Header* header = (Header*)&data_[cur_sz];
Footer* footer = (Footer*)(header->str + header->len);
```

and then we can use `header->len`, `footer->len`, and `header->str` instead of char pointers.

To get pointer to Header from pointer to string `x`, we can use `ROC_CONTAINER_OF(x, Header, str)`.

**2.** Instead of `(8 - ((add_sz + len_sz) % 8))` we can use `AlignOps::align_as()`

**3.** Do we really need to store padding size? I think we can store unaligned length and calculate aligned length using `AlignOps::align_as()`.

**4.** To keep tests simpler and to reduce memory usage a bit, I think it's OK to use uint32_t instead of size_t to store length.

ForeverASilver:
1. I agree adding structs would probably make this cleaner, currently working on it.
for points 2, 3, and 4, i was not aware of AlignOps, hence the alignment to keep UBSan happy about accessing size_t on 8 byte boundaries, and the different expected allocations on 4 byte and 8 byte size_t in tests, will change this though to utilize AlignOps as you mentioned

gavv:
@ForeverASilver please ping me when it's ready for review :)

ForeverASilver:
@gavv ready for review!

implemented the Header and Footer structs as per your comments, though due to errors within ROC_CONTAINER_OF, had to do some a couple of const_cast.

src/internal_modules/roc_core/string_list.cpp
```diff
@@ -24,15 +25,19 @@ size_t StringList::size() const {
...
 const char* StringList::front() const {
     if (size_) {
-        return &data_[0];
+        const Header* first_header =
+            ROC_CONTAINER_OF(const_cast<char*>(&data_[0]), Header, len);
+        return first_header->str_;
     } else {
         return NULL;
     }
 }
```
gavv, line 34: We can make thing more clean if we'll maintain two class fields:

```cpp
    Header* front_;
    Header* back_;
```

Then front(), for example, will look as:

```cpp
const char* StringList::front() const {
    if (front_) {
        return front_->str;
    } else {
        return NULL;
    }
}
```

src/internal_modules/roc_core/string_list.cpp
```diff
@@ -54,14 +59,51 @@ const char* StringList::nextof(const char* str) const {
         roc_panic("stringlist: string doesn't belong to the list");
     }
 
-    const char* ptr = str + strlen(str) + 1;
+    const Header* str_header =
+        (const Header*)(ROC_CONTAINER_OF(const_cast<char*>(str), Header, str_));
+    const size_t align = AlignOps::align_as(sizeof(Header) + str_header->len + 1, 8);
+    const char* ptr = (const char*)str_header + align + sizeof(Footer);
     roc_panic_if(ptr > end);
```
gavv, line 66: A few ideas how to cleanup nextof() and prevof()

gavv: 1. To avoid const casts in user code, we can fix definition of ROC_CONTAINER_OF as follows:

```cpp
#define ROC_CONTAINER_OF(ptr, type, member)                                              \
    (reinterpret_cast<type*>(                                                            \
        const_cast<char*>((const char*)(ptr)-offsetof(type, member))))
```

Then we can use it like this:

```cpp
const Header* str_header = ROC_CONTAINER_OF(str, const Header, str);
```

gavv: 2. Let's extract this piece:

```cpp
    const char* begin = &data_[0];
    const char* end = &data_[0] + data_.size();
    // check_pointer_(str);

    if (str < begin || str >= end) {
        roc_panic("stringlist: string doesn't belong to the list");
    }
```

into a private method, say, `check_is_member_(str)` (we have similar in core::List).

gavv: 3. I played with the code a bit, and it appears to me that the code will be much simpler if `header->len` and `footer->len` would contain the length of the whole block (header + string + alignment + footer). What do you think?

Here is an example of how nextof() and prevof() could look like:

```cpp
    const Header* str_header = ROC_CONTAINER_OF(str, const Header, str);
    if (str_header == back_) {
        return NULL;
    }

    const Header* next_header =
        (const Header*)((const char*)str_header + str_header->len);

    return next_header->str;
```

```cpp
    const Header* str_header = ROC_CONTAINER_OF(str, const Header, str);
    if (str_header == front_) {
        return NULL;
    }

    const Footer* prev_footer = (const Footer*)((const char*)str_header - sizeof(Footer));
    const Header* prev_header =
        (const Header*)((const char*)str_header - prev_footer->len);

    return prev_header->str;
```

(I didn't test it)

src/internal_modules/roc_core/string_list.cpp
```diff
@@ -86,17 +128,26 @@ bool StringList::push_back(const char* begin, const char* end) {
...
         return false;
     }
 
     back_ = &data_[cur_sz];
-    memcpy(&data_[cur_sz], begin, add_sz - 1);
-    data_[cur_sz + add_sz - 1] = '\0';
+    memcpy(&data_[cur_sz], &str_header, sizeof(Header));              // copy header
+    memcpy(&data_[cur_sz + sizeof(Header)], begin, add_sz - 1);       // copy string
+    data_[cur_sz + sizeof(Header) + add_sz - 1] = '\0';               // add null
+    memcpy(&data_[cur_sz + align_head], &str_footer, sizeof(Footer)); // copy footer
```
gavv, line 150: I think we can simplify this code a bit if we'll create pointers to header and footer and then write directly into them, instead of memcpy-ing header/footer. Like this:

```cpp
    const size_t cur_sz = data_.size();
    const size_t add_sz = sizeof(Header)
        + AlignOps::align_as(size_t(end - begin), sizeof(Header)) + sizeof(Footer);

    // resize ...

    Header* str_header = (Header*)&data_[cur_sz];
    Footer* str_footer = (Footer*)&data_[cur_sz + add_sz] - sizeof(Footer);

    // fill fields of str_header and str_footer ...
```

gavv (commented):
Thanks for update

src/internal_modules/roc_core/string_list.h
```diff
@@ -87,6 +95,15 @@ class StringList : public NonCopyable<> {
     core::Array<char> data_;
     const char* back_;
     size_t size_;
+
+    struct Header {
+        uint32_t len;
+        char str_[];
```
gavv, line 101: nit: `str_` => `str`

src/internal_modules/roc_core/string_list.cpp
ForeverASilver: push_unique was the only reason why i had the keep the length to the actual string otherwise i couldnt get the string length to do the comparison in there, but i could remove it if you’d like me to

gavv: I wanted to remove push_unique() and instead add `bool find(const char*)`, and use:

```cpp
if (!list.find("foo")) {
    list.push_back("foo");
}
```

IIRC there were only 2 usages of push_unique actually.

You can do this if you wish, or you can keep push_unique but don't bother making it efficient/clean, and I'll remove it later. In both ways I guess you can store whole block size in `len`.

Also I think `push_back(begin, end)` overload may be removed for simplicity, as it's not used IIRC.

ForeverASilver: ill replace push_unique() with `find()`
as for `push_back(begin, end)` one usage i could possibly think of is maybe wanting push a string with a null character within it, like if you ran it through `strtok` for example, but im find with removing it if you still think it wont be useful

gavv: Thanks. Makes sense, we indeed have non-null-terminated strings inside some ragel parsers, push_back(begin, end) can be useful here, let's keep it.

ForeverASilver:
@gavv ready for review!

gavv:
Thanks!

Small follow-up commit: 46491cdb6e49a1b0755b78d73199d21234912835 (cosmetic fixes)

Issue #578, opened 2023-09-29
Print list of supported protocols


gavv:
roc-recv and roc-send tools have option to print list of supported audio backends and formats:

```
$ roc-recv --list-supported
supported schemes for audio devices and files:
  pulse:// alsa:// oss:// file://

supported formats for audio files:
  .aifc .aiffc .aiff .aif .al .au .snd .avr .cdda .cdr .cvsd .cvs .cvu .dat
  .dvms .vms .f32 .f64 .gsrt .hcom .htk .ima .la .lu .maud .prc .s8 .s16
  .s24 .s32 .sf .ircam .sln .smp .sndr .sndt .sox .sph .nist .8svx .txw
  .u8 .sou .fssd .u16 .u24 .u32 .ul .voc .vox .wav .wavpcm .amb .wve .xa
  .amr-nb .anb .amr-wb .awb .caf .fap .flac .gsm .lpc10 .lpc .mat4 .mat
  .mat5 .mp3 .mp2 .paf .pvf .sd2 .sds .vorbis .ogg .w64 .wv .xi
```

We need to teach them to include information about supported protocols as well, something like this:

```
supported network protocols:
  audiosrc: rtp:// rs8m+rtp:// ldpc+rtp://
  audiorpr: rs8m:// ldpc://
  audioctl: rtcp://
```

Here, each line has [interface name](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_address/interface.h) (`audiosrc` and others), and then lists [protocols](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_address/protocol.h) supported by this interface.

Steps:

* add two new methods to [ProtocolMap](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_address/protocol_map.h): `get_supported_interfaces(core::Array<Interface>&)` (returns list of all interfaces for which there is at least one protocol) and `get_supported_protocols(Interface, core::StringList&)` (returns list of all protocols for given interface), similar to what we have in [BackendDispatcher](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_sndio/backend_dispatcher.h) for supported schemes and formats

[...]

PR #642, opened 2023-11-12
print_supported implemented for recv and send tools.

Hrick87:
Issue #578 

I am uncertain if the consolidated interface enumeration and its corresponding protocol should be a part of the printed list, but otherwise I believe I've done this all correctly.

### DIFF
--- a/src/internal_modules/roc_address/print_supported.cpp
+++ b/src/internal_modules/roc_address/print_supported.cpp
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2023 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_address/print_supported.h"
+#include "roc_address/protocol_map.h"
+#include "roc_core/log.h"
+#include "roc_core/printer.h"
+
+namespace roc {
+namespace address {
+
+namespace {
+
+enum { LineSize = 70 };
+
+void print_interface_protos(core::Printer& prn,
+                            Interface interface,
+                            const core::StringList& list) {
+    const char* str = list.front();
+
+    while (str != NULL) {
+        prn.writef(" ");
+
+        size_t size = 0;
+
+        prn.writef(" %s:", interface_to_str(interface));
+
+        while (size < LineSize) {
+            size += prn.writef(" %s%s%s", "", str, "://");
+
+            str = list.nextof(str);
+            if (!str) {
+                break;
+            }
+        }
+
+        prn.writef("\n");
+    }
+}
+
+} // namespace
+
+bool print_supported(ProtocolMap& protocol_map, core::IArena& arena) {
+    core::Printer prn;
+    core::Array<Interface> interface_array(arena);
+    core::StringList list(arena);
+
+    if (!protocol_map.get_supported_interfaces(interface_array)) {
+        roc_log(LogError, "can't retrieve interface array");
+        return false;
+    }
+
+    for (size_t n_interface = 0; n_interface < interface_array.size(); n_interface++) {
+        if (!protocol_map.get_supported_protocols(interface_array[n_interface], list)) {
+            roc_log(LogError, "can't retrieve protocols list");
+            return false;
+        }
+
+        if (n_interface == 0) {
+            prn.writef("\nsupported network protocols:\n");
+        }
+
+        print_interface_protos(prn, interface_array[n_interface], list);
+    }
+    return true;
+}
+
+} // namespace address
+} // namespace roc
--- a/src/internal_modules/roc_address/print_supported.h
+++ b/src/internal_modules/roc_address/print_supported.h
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2023 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_address/print_supported.h
+//! @brief Print supported interfaces and protocols.
+
+#ifndef ROC_ADDRESS_PRINT_SUPPORTED_H_
+#define ROC_ADDRESS_PRINT_SUPPORTED_H_
+
+#include "roc_address/protocol_map.h"
+#include "roc_core/attributes.h"
+#include "roc_core/iarena.h"
+
+namespace roc {
+namespace address {
+
+//! Print supported interfaces and protocols.
+ROC_ATTR_NODISCARD bool print_supported(ProtocolMap&, core::IArena&);
+
+} // namespace address
+} // namespace roc
+
+#endif // ROC_ADDRESS_PRINT_SUPPORTED_H_
--- a/src/internal_modules/roc_address/protocol_map.cpp
+++ b/src/internal_modules/roc_address/protocol_map.cpp
@@ -121,6 +121,56 @@ const ProtocolAttrs* ProtocolMap::find_by_scheme(const char* scheme) const {
     return NULL;
 }
 
+bool ProtocolMap::get_supported_interfaces(core::Array<Interface>& interface_array) {
+    interface_array.clear();
+    bool interfaces_exist = false;
+
+    for (unsigned n_iface = (unsigned)Iface_Consolidated; n_iface != (unsigned)Iface_Max;
+         n_iface++) {
+        for (size_t n_proto = 0; n_proto < MaxProtos; n_proto++) {
+            if (protos_[n_proto].protocol == Proto_None) {
+                continue;
+            }
+
+            if (n_iface == (unsigned)protos_[n_proto].iface) {
+                if (!interface_array.grow(interface_array.size() + 1)) {
+                    return false;
+                }
+                interface_array.push_back(protos_[n_proto].iface);
+                interfaces_exist = true;
+                break;
+            }
+        }
+    }
+
+    return interfaces_exist;
+}
+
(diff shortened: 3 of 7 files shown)


src/internal_modules/roc_address/print_supported.cpp
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) 2019 Roc Streaming authors
```
gavv, line 2:
```suggestion
 * Copyright (c) 2023 Roc Streaming authors
```

src/internal_modules/roc_address/print_supported.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace {
+
+enum { ArraySize = 100, LineSize = 70 };
+
+void print_string_list(core::Printer& prn,
+                       Interface interface,
+                       const core::StringList& list,
+                       const char* prefix,
+                       const char* suffix) {
```
gavv, line 25: Since this function knows about Interface, I suggest to rename it to something a bit more specific, e.g. "print_interface" or "print_interface_proto(col)s".

Also I suggest to remove "prefix" and "suffix" arguments and instead hard-code their values inside the function.

src/internal_modules/roc_address/print_supported.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2019 Roc Streaming authors
```
gavv, line 2:
```suggestion
 * Copyright (c) 2023 Roc Streaming authors
```

src/internal_modules/roc_address/print_supported.h
```diff
@@ -0,0 +1,28 @@
...
+#define ROC_ADDRESS_PRINT_SUPPORTED_H_
+
+#include "roc_address/protocol_map.h"
+#include "roc_core/attributes.h"
+#include "roc_core/iarena.h"
+
+namespace roc {
+namespace address {
+
+//! Print supported schemes and formats.
```
gavv, line 22:
```suggestion
//! Print supported interfaces and protocols.
```

src/internal_modules/roc_address/protocol_map.h
```diff
@@ -66,6 +68,12 @@ class ProtocolMap : public core::NonCopyable<> {
     //! Get protocol attributes by scheme name.
     const ProtocolAttrs* find_by_scheme(const char* scheme) const;
 
+    //! Get list of interfaces with at least one protocol
+    ROC_ATTR_NODISCARD bool get_supported_interfaces(core::Array<Interface>&);
+
+    //! Get all supported protocols
+    ROC_ATTR_NODISCARD bool get_supported_protocols(Interface, core::StringList&);
```
gavv, line 75: nitpick: Please add dot at the end of each doxygen comment, like:

```
//! Get list of interfaces with at least one protocol.
```

IIRC doxygen auto-brief mode that we use requires it.

src/internal_modules/roc_address/print_supported.cpp
```diff
@@ -0,0 +1,80 @@
...
+} // namespace
+
+bool print_supported(ProtocolMap& protocol_map, core::IArena& arena) {
+    core::Printer prn;
+    core::Array<Interface> interface_array(arena);
+    core::StringList list(arena);
+
+    if (!interface_array.resize(Iface_Max)) {
+        return false;
+    }
```
gavv, line 57: This call is not quite correct: we resize() array here, but then we clear() it in get_supported_interfaces(), so it's size becomes zero again. It happens that it's capacity remains Iface_Max, but it's not obvious.

I suggest to remove resize() call here, and instead do the following in get_supported_interfaces() when we're adding interface to array:

```cpp
if (!interface_array.grow(interface_array.size() + 1)) {
    return false;
}
interface_array.push_back(protos_[y].iface);
```

src/internal_modules/roc_address/protocol_map.cpp
```diff
@@ -129,5 +129,44 @@ void ProtocolMap::add_proto_(const ProtocolAttrs& proto) {
     protos_[proto.protocol] = proto;
 }
 
+bool ProtocolMap::get_supported_interfaces(core::Array<Interface>& interface_array) {
+    interface_array.clear();
+    bool interfaces_exist = false;
+
+    for (unsigned x = (unsigned)Iface_Consolidated + 1; x != (unsigned)Iface_Max; x++) {
+        for (size_t y = 0; y < MaxProtos; y++) {
```
gavv, line 137: nitpick: here and in other places; let's rename counters like "x" and "y" to something a bit more meaningful so that they are not confused with each other.

E.g. "x" can be renamed to "n_iface" or "ni", and "y" to "n_proto" or "np", or something like this.

src/internal_modules/roc_address/protocol_map.cpp
```diff
@@ -129,5 +129,44 @@ void ProtocolMap::add_proto_(const ProtocolAttrs& proto) {
...
+
+    return interfaces_exist;
+}
+
+bool ProtocolMap::get_supported_protocols(Interface interface, core::StringList& list) {
+    list.clear();
+
+    bool protocolsExist = false;
+
+    for (size_t x = 0; x < MaxProtos; x++) {
```
gavv, line 154: In get_supported_interfaces and get_supported_protocols, we should skip empty protocol attrs:

```cpp
    for (size_t x = 0; x < MaxProtos; x++) {
        if (protos_[x].protocol == Proto_None) {
            continue;
        }
        ...
```

src/internal_modules/roc_address/protocol_map.cpp
```diff
@@ -129,5 +129,44 @@ void ProtocolMap::add_proto_(const ProtocolAttrs& proto) {
...
+        }
+    }
+
+    return interfaces_exist;
+}
+
+bool ProtocolMap::get_supported_protocols(Interface interface, core::StringList& list) {
+    list.clear();
+
+    bool protocolsExist = false;
```
gavv, line 152: We use snake_case for variables.

src/internal_modules/roc_address/print_supported.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2019 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_address/print_supported.h
+//! @brief Print supported schemes and formats.
```
gavv, line 10:
```suggestion
//! @brief Print supported interfaces and protocols.
```

src/internal_modules/roc_address/protocol_map.cpp
```diff
@@ -129,5 +129,44 @@ void ProtocolMap::add_proto_(const ProtocolAttrs& proto) {
...
+
+    for (size_t x = 0; x < MaxProtos; x++) {
+        if (interface == ProtocolMap::instance().protos_[x].iface) {
+            const char* proto_name =
+                proto_to_str(ProtocolMap::instance().protos_[x].protocol);
```
gavv, line 157:
```suggestion
            const char* proto_name = protos_[x].scheme_name;
```

(proto_to_str just returns scheme name)

src/internal_modules/roc_address/protocol_map.cpp
```diff
@@ -129,5 +129,44 @@ void ProtocolMap::add_proto_(const ProtocolAttrs& proto) {
     protos_[proto.protocol] = proto;
 }
 
+bool ProtocolMap::get_supported_interfaces(core::Array<Interface>& interface_array) {
+    interface_array.clear();
+    bool interfaces_exist = false;
+
+    for (unsigned x = (unsigned)Iface_Consolidated + 1; x != (unsigned)Iface_Max; x++) {
```
gavv, line 136: Why do you skip Iface_Consolidated? It should be included as well, so we can start from zero.

src/internal_modules/roc_address/protocol_map.cpp
```diff
@@ -129,5 +129,44 @@ void ProtocolMap::add_proto_(const ProtocolAttrs& proto) {
     protos_[proto.protocol] = proto;
 }
 
+bool ProtocolMap::get_supported_interfaces(core::Array<Interface>& interface_array) {
```
gavv, line 132: Nitpick: please move definitions of get_supported_interfaces and get_supported_protocols before definition of add_proto_, because usually we place definitions in .cpp in same order as declarations in header.

src/internal_modules/roc_address/print_supported.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "roc_address/protocol_map.h"
+#include "roc_core/log.h"
+#include "roc_core/printer.h"
+
+namespace roc {
+namespace address {
+
+namespace {
+
+enum { ArraySize = 100, LineSize = 70 };
```
gavv, line 19: Please remove ArraySize, it's not used.

BTW could you also remove ArraySize from src/internal_modules/roc_sndio/print_supported.cpp, from where you copied it? It's not used there too.

gavv (commented):
Nice, thanks for PR! Here are a few comments.

gavv:
> I am uncertain if the consolidated interface enumeration and its corresponding protocol should be a part of the printed list, but otherwise I believe I've done this all correctly.

Yep, consolidated should be included.

gavv:
Thanks for update, LGTM!

BTW, please don't resolve discussions by yourself, keep them for reviewer, otherwise it's hard to track what is needed to be re-reviewed. Instead you can add comment or :+1: on discussion to indicate that it was addressed.

Hrick87:
> Thanks for update, LGTM!
> 
> BTW, please don't resolve discussions by yourself, keep them for reviewer, otherwise it's hard to track what is needed to be re-reviewed. Instead you can add comment or 👍 on discussion to indicate that it was addressed.

Ah my mistake, understood. I'll stick to using comments/thumbs up instead. I thought the resolve something each of us did individually on our own ends.

PR #646, opened 2023-11-13
Rename Pool

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/610, specifically https://github.com/roc-streaming/roc-toolkit/issues/610#issuecomment-1806516586

# What

* Renamed Pool to SlabPool, including files.
* Renamed PoolImpl to SlabPoolImpl, including files.

* Renamed file using snake_case.

* There might be some general references to pool in the documentation - not sure if we want to change that.

# Testing

There are tests.


gavv:
Thanks!

I think we should rename PoolFlags/PoolFlag_EnableGuards/DefaultPoolFlags too.

gavv:
> 🤖 Upon creation, pull request description does not have a link to an issue. If there is a related issue, please add it to the description using any of the [supported formats](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls).

I've fixed that check to support links to issue comment.

nolan-veed:
> I think we should rename PoolFlags/PoolFlag_EnableGuards/DefaultPoolFlags too.

Yes. I wasn't sure about that.

PR #652, opened 2023-12-03
Added MemoryLimiter and other classes etc.

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/610

# What

* Added MemoryLimiter, LimitedArena, LimitedPool classes.
* Updated and added new tests.

# Testing

There are tests.

### DIFF
--- a/src/internal_modules/roc_core/heap_arena.cpp
+++ b/src/internal_modules/roc_core/heap_arena.cpp
@@ -118,5 +118,31 @@ void HeapArena::deallocate(void* ptr) {
     free(chunk);
 }
 
+size_t HeapArena::compute_allocated_size(size_t size) const {
+    return sizeof(ChunkHeader) + sizeof(ChunkCanary) + size + sizeof(ChunkCanary);
+}
+
+size_t HeapArena::allocated_size(void* ptr) const {
+    if (!ptr) {
+        roc_panic("heap arena: null pointer");
+    }
+
+    ChunkHeader* chunk =
+        ROC_CONTAINER_OF((char*)ptr - sizeof(ChunkCanary), ChunkHeader, data);
+
+    const bool is_owner = chunk->owner == this;
+
+    if (!is_owner) {
+        if (AtomicOps::load_seq_cst(flags_) & HeapArenaFlag_EnableGuards) {
+            roc_panic("heap arena: attempt to get allocated size of chunk not belonging "
+                      "to this arena: this_arena=%p chunk_arena=%p",
+                      (const void*)this, (const void*)chunk->owner);
+        }
+        return 0;
+    }
+
+    return compute_allocated_size(chunk->size);
+}
+
 } // namespace core
 } // namespace roc
--- a/src/internal_modules/roc_core/ipool.h
+++ b/src/internal_modules/roc_core/ipool.h
@@ -24,8 +24,8 @@ class IPool {
 public:
     virtual ~IPool();
 
-    //! Get size of object allocated by pool.
-    virtual size_t object_size() const = 0;
+    //! Get size of the allocation per object.
+    virtual size_t allocation_size() const = 0;
 
     //! Reserve memory for given number of objects.
     //! @returns
@@ -54,8 +54,7 @@ public:
 //! Placement new for core::IPool.
 //! @note
 //!  nothrow forces compiler to check for NULL return value before calling ctor.
-inline void* operator new(size_t size, roc::core::IPool& pool) throw() {
-    roc_panic_if_not(size <= pool.object_size());
+inline void* operator new(size_t, roc::core::IPool& pool) throw() {
     return pool.allocate();
 }
 
--- a/src/internal_modules/roc_core/limited_arena.cpp
+++ b/src/internal_modules/roc_core/limited_arena.cpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2023 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_core/limited_arena.h"
+
+namespace roc {
+namespace core {
+
+LimitedArena::LimitedArena(IArena& arena, MemoryLimiter& memory_limiter)
+    : arena_(arena)
+    , memory_limiter_(memory_limiter) {
+}
+
+void* LimitedArena::allocate(size_t size) {
+    size_t allocated_size = arena_.compute_allocated_size(size);
+    if (memory_limiter_.acquire(allocated_size)) {
+        void* ptr = arena_.allocate(size);
+        if (!ptr)
+            memory_limiter_.release(allocated_size);
+        return ptr;
+    }
+    return NULL;
+}
+
+void LimitedArena::deallocate(void* ptr) {
+    size_t allocated_size = arena_.allocated_size(ptr);
+    arena_.deallocate(ptr);
+    memory_limiter_.release(allocated_size);
+}
+
+size_t LimitedArena::compute_allocated_size(size_t size) const {
+    return arena_.compute_allocated_size(size);
+}
+
+size_t LimitedArena::allocated_size(void* ptr) const {
+    return arena_.allocated_size(ptr);
+}
+
+} // namespace core
+} // namespace roc
--- a/src/internal_modules/roc_core/limited_arena.h
+++ b/src/internal_modules/roc_core/limited_arena.h
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2023 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_core/limited_arena.h
+//! @brief Limited Arena.
+
+#ifndef ROC_CORE_LIMITED_ARENA_H_
+#define ROC_CORE_LIMITED_ARENA_H_
+
+#include "roc_core/iarena.h"
+#include "roc_core/memory_limiter.h"
+#include "roc_core/noncopyable.h"
+
+namespace roc {
+namespace core {
+
+//! Decorator around IArena to make it memory limited.
+class LimitedArena : public NonCopyable<>, public IArena {
+public:
+    //! Initialize.
+    LimitedArena(IArena& arena, MemoryLimiter& memory_limiter);
+
+    //! Allocate memory after checking with the memory limiter.
+    //! @returns
+    //!  pointer to a maximum aligned uninitialized memory at least of @p size
+    //!  bytes or NULL if memory can't be allocated.
+    virtual void* allocate(size_t size);
+
+    //! Deallocate previously allocated memory.
+    virtual void deallocate(void* ptr);
+
+    //! Computes how many bytes will be actually allocated if allocate() is called with
+    //! given size. Covers all internal overhead, if any.
+    virtual size_t compute_allocated_size(size_t size) const;
+
+    //! Returns how many bytes was allocated for given pointer returned by allocate().
+    //! Covers all internal overhead, if any.
+    //! Returns same value as computed by compute_allocated_size(size).
+    virtual size_t allocated_size(void* ptr) const;
+
+private:
+    IArena& arena_;
+    MemoryLimiter& memory_limiter_;
+};
+
+} // namespace core
+} // namespace roc
+
+#endif // ROC_CORE_LIMITED_ARENA_H_
--- a/src/internal_modules/roc_core/limited_pool.cpp
+++ b/src/internal_modules/roc_core/limited_pool.cpp
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2023 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_core/limited_pool.h"
+
+namespace roc {
+namespace core {
+
+LimitedPool::LimitedPool(IPool& pool, MemoryLimiter& memory_limiter)
+    : pool_(pool)
+    , memory_limiter_(memory_limiter) {
+}
+
+size_t LimitedPool::allocation_size() const {
+    return pool_.allocation_size();
+}
+
+ROC_ATTR_NODISCARD bool LimitedPool::reserve(size_t n_objects) {
+    return pool_.reserve(n_objects);
+}
+
+void* LimitedPool::allocate() {
+    size_t allocation_size = pool_.allocation_size();
+    if (memory_limiter_.acquire(allocation_size)) {
+        void* ptr = pool_.allocate();
+        if (!ptr)
+            memory_limiter_.release(allocation_size);
+        return ptr;
+    }
+    return NULL;
+}
+
+void LimitedPool::deallocate(void* memory) {
+    pool_.deallocate(memory);
+    memory_limiter_.release(pool_.allocation_size());
+}
+
+} // namespace core
+} // namespace roc
--- a/src/internal_modules/roc_core/memory_limiter.cpp
+++ b/src/internal_modules/roc_core/memory_limiter.cpp
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2023 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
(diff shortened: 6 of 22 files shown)


gavv:
Yep, it should be atomic, and ideally lock free. I guess it would be enough to use core::Atomic or core::AtomicOps.

We likely need to get rid of would_alliw because it's usage will be always a race.

Otherwise, the class looks fine!

gavv:
Oh, also %zu is not portable, please use %lu with a cast to ulong.

nolan-veed:
> Oh, also %zu is not portable, please use %lu with a cast to ulong.

OK. I've used it elsewhere in the code as well.

nolan-veed:
> Yep, it should be atomic, and ideally lock free. I guess it would be enough to use core::Atomic or core::AtomicOps.

See my approach now.

nolan-veed:
OK. I should have kept this in draft.

I've just added the LimitedArena, LimitedPool, and could use your thoughts on it and a few things. Not yet added tests.

src/internal_modules/roc_core/limited_arena.cpp
```diff
@@ -0,0 +1,24 @@
...
+void* LimitedArena::allocate(size_t size) {
+    if (memoryLimiter_.acquire(size)) {
+        return arena_.allocate(size);
+    }
+    return NULL;
+}
+
+void LimitedArena::deallocate(void* memory) {
+    // TODO: We don't know the size of allocated memory in order to check with limited.
+    arena_.deallocate(memory);
```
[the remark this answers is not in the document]
gavv: Indeed. I think we could modify IArena interface and add something like `IArena::size_of(void*)`, or make `deallocate()` returning size of deallocated memory.

Currently IArena is implemented only by HeapArena + some mocks in tests.

gavv:
Some notes:

1. Ideally we should also acquire the size of the overhead for every allocated chunk/slot. Arena and pool add headers and guards, and for many allocations this unaccounted overhead will accumulate and may become significant. I guess we need to add methods IArena and IPool to report the overhead, and to take it into account in Limited versions.

2. I think MemoryLimiter::release() should never return false (as well as deallocate). When invariant is failed, it indicates a critical bug, we can panic, but there is no need for error checking because there is anyway to meaningful handling of this error except panic.

3. In CAS loops, please add core::cpu_relax() call.

4. I think we should first deallocate and only then release memory from limiter, to avoid over-allocation.

5. I think we can allow `max_bytes = 0` which would mean "no limit". This way we can create all memory limiters unconditionally (even when there is no actual limit), which I think will simplify code.

6. Please log errors when MemoryLimiter::acquire() fails, and include the limit and requested size into the log message. Also, similar to SlabPool, it would be nice to add a name to MemoryLimiter, and include it into the log message too.

7. Please inherit classes from NonCopyable when appreciable.

8. Probably acquiring/releasing zero bytes should be a panic?

9. I think MemoryLimiter destructor should panic if there is unreleased memory. Though, if it would complicate some usages, it's not necessary.

nolan-veed:
Quick q about... how to approach LimitedPool.

* SlabPool has an EmbeddedCapacity - this means that memory will already be "booked" when it's created.
* IPool has reserve() and allocate(). So, how should the limited version work? I guess we check with MemoryLimiter only on reserve()? So, allocate() and deallocate() just don't bother with the MemoryLimiter?

gavv:
1. The goal of per-pool limits is to limit number of packets given to one socket/session, so that we can be sure that if one connection flooded, we will not overallocate and other connections won't be affected.

Hence, I think for per-session LimitedPool it doesn't matter how many packets are booked in embedded array or in slab, since they remain available for other sessions. It only matters how many packets were allocated for this specific session.

So I think LimitedPool should only acquire and release in allocate/deallocate, and doesn't need to bother about reserve.

2. To control the total amount of memory, we will have root LimitedArena with global limit (max_memory). In the end, all allocations, including objects allocated by pools and pools themselves and their embedded capacity, all will go through this limited arena.

gavv:
I see some commits, please ping me when it's ready for review :)

nolan-veed:
> I see some commits, please ping me when it's ready for review :)

Yea. Sorry. Small tidy ups are remaining and a few more tests.

gavv:
I've looked at the implementation and have small suggestion about IPool and IArena interface.

size_of + overhead / object_size + overhead looks a bit too low level, because the caller needs to know that there is "payload" size and "overhead" size. The caller also should assume that the overhead is constant (same for all allocations), which by the way may not be true when arena applies alignment?

Instead of that, I think it would be a bit simpler if the caller will operate with just one "allocation size", which includes payload + overhead.

For IPool, it can be one method, e.g.:

```cpp
    virtual size_t allocation_size() = 0;
```

We can remove object_size() since it's not really used and keep only allocation_size() which takes overhead into account.

For IArena, we'd still need 2 methods, but they will both include payload + overhead:

```cpp
    //! Computes how much bytes will be actually allocated if allocate() is called with given size.
    //! Covers all internal overhead, if any.
    virtual size_t compute_allocated_size(size_t size) = 0;

    //! Returns how much bytes was allocated for given pointer returned by allocate().
    //! Covers all internal overhead, if any.
    //! Returns same value as computed by compute_allocated_size(size).
    virtual size_t allocated_size(void*) = 0;
```

Feel free to change naming and comments of course.

What do you think?

src/internal_modules/roc_core/limited_arena.h
```diff
@@ -0,0 +1,51 @@
...
+
+    //! Get size of previously allocated memory, excluding overhead.
+    virtual size_t size_of(void* ptr);
+
+    //! Get number of overhead bytes per allocation.
+    virtual size_t overhead();
+
+private:
+    IArena& arena_;
+    MemoryLimiter& memoryLimiter_;
```
gavv, line 45: nit: please use snake_case for members.

src/internal_modules/roc_core/memory_limiter.cpp
```diff
@@ -0,0 +1,80 @@
...
+        }
+        if (bytes_acquired_.compare_exchange(current, next)) {
+            return true;
+        }
+        cpu_relax();
+    } while (true);
+    roc_log(LogError,
+            "memory limiter (%s): could not acquire %lu bytes due to limit, currently "
+            "acquired %lu",
+            name_, (unsigned long)num_bytes, (unsigned long)current);
```
gavv, line 50: Could you please also include limit into the message?

src/internal_modules/roc_core/memory_limiter.cpp
```diff
@@ -0,0 +1,80 @@
...
+        current = bytes_acquired_;
+        if (current < num_bytes) {
+            break;
+        }
+        size_t next = current - num_bytes;
+        if (bytes_acquired_.compare_exchange(current, next)) {
+            return;
+        }
+        cpu_relax();
+    } while (true);
```
gavv, line 69: Seems that we don't need CAS loop in release, and can just do atomic subtraction (`-=`) and check its return value?

src/internal_modules/roc_core/memory_limiter.h
```diff
@@ -0,0 +1,49 @@
...
+#include "roc_core/stddefs.h"
+
+namespace roc {
+namespace core {
+
+//! Memory limiter.
+class MemoryLimiter : public NonCopyable<> {
+public:
+    //! Initialize memory limiter.
+    explicit MemoryLimiter(const char* name, size_t max_bytes);
```
gavv, line 26: Could you please add a small comment, describing:

* the purpose of the class (why we need it / how we use it)
* meaning of acquire/release and max_bytes (including zero value)
* contracts (can't release more than acquired, should release all before dtor)

src/internal_modules/roc_core/memory_limiter.h
```diff
@@ -0,0 +1,49 @@
...
+
+    //! Destroy memory limiter.
+    ~MemoryLimiter();
+
+    //! Acquire memory.
+    bool acquire(size_t num_bytes);
```
gavv, line 32: Could you please document return value? Also I think we should add ROC_ATTR_NODISCARD.

src/internal_modules/roc_core/memory_limiter.h
```diff
@@ -0,0 +1,49 @@
...
+
+    //! Release memory.
+    void release(size_t num_bytes);
+
+    //! Get number of bytes currently acquired.
+    size_t acquired();
```
gavv, line 38: I guess we can make it const.

gavv: Also usually we name such method "num_acquired" or "acquired_count" or something like this, to make it clear that it's not boolean.

src/internal_modules/roc_core/limited_arena.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "roc_core/iarena.h"
+#include "roc_core/memory_limiter.h"
+#include "roc_core/noncopyable.h"
+
+namespace roc {
+namespace core {
+
+//! Decorator around IArena to make it memory limited.
+class LimitedArena : public NonCopyable<LimitedArena>, public IArena {
```
gavv, line 23: nit: `<LimitedArena>` is not needed here, `NonCopyable<>` would be enough.

That template argument is only used in base classes, to prevent NonCopyable to be inherited multiple times via different bases, to suppress compiler warnings. E.g. an object can inherit both RefCounted and ListNode, so RefCounted uses `NonCopyable<RefCounted>` and ListNode uses `NonCopyable<ListNode>`.

Since LimitedArena is not supposed to be inherited, this doesn't matter.

gavv:
Thanks for update, I've posted a few more comments.

nolan-veed:
> I've looked at the implementation and have small suggestion about IPool and IArena interface.
> ...
> What do you think?

It does make sense. I did think about something similar to this initially - should have suggested.
I'll go with those names, they are clear.

src/internal_modules/roc_core/memory_limiter.cpp
nolan-veed: OK. I had to check for underflow, please double-check.

src/internal_modules/roc_core/ipool.h
```diff
@@ -54,8 +54,7 @@ class IPool {
 //! Placement new for core::IPool.
 //! @note
 //!  nothrow forces compiler to check for NULL return value before calling ctor.
-inline void* operator new(size_t size, roc::core::IPool& pool) throw() {
-    roc_panic_if_not(size <= pool.object_size());
```
nolan-veed, line 58: @gavv 
Note: I had to remove this only good use of object_size() as well.

src/internal_modules/roc_core/heap_arena.cpp
```diff
@@ -118,5 +118,33 @@ void HeapArena::deallocate(void* ptr) {
...
+        roc_panic("heap arena: null pointer");
+    }
+
+    ChunkHeader* chunk =
+        ROC_CONTAINER_OF((char*)ptr - sizeof(ChunkCanary), ChunkHeader, data);
+
+    const bool is_owner = chunk->owner == this;
+
+    if (!is_owner) {
+        if (AtomicOps::load_seq_cst(flags_) & HeapArenaFlag_EnableGuards) {
```
nolan-veed, line 136: I had to remove `num_guard_failures_++;` from here, because this method is const. Would you rather have the member variable as mutable?

gavv: It makes sense to make it mutable. I'd say it's not part of the logical object state.

src/internal_modules/roc_core/heap_arena.cpp
```diff
@@ -118,5 +118,33 @@ void HeapArena::deallocate(void* ptr) {
...
+            roc_panic(
+                "heap arena:"
+                " attempt to get deallocated size of chunk not belonging to this arena:"
+                " this_arena=%p chunk_arena=%p",
+                (const void*)this, (const void*)chunk->owner);
```
gavv, line 141: "deallocated size" => "allocated size"

src/internal_modules/roc_core/limited_arena.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+LimitedArena::LimitedArena(IArena& arena, MemoryLimiter& memory_limiter)
+    : arena_(arena)
+    , memory_limiter_(memory_limiter) {
+}
+
+void* LimitedArena::allocate(size_t size) {
+    size_t allocated_size = arena_.compute_allocated_size(size);
+    if (memory_limiter_.acquire(allocated_size)) {
+        return arena_.allocate(size);
```
gavv, line 22: If nested arena returns NULL, we should release acquired memory.

src/internal_modules/roc_core/limited_pool.cpp
```diff
@@ -0,0 +1,40 @@
...
+    return pool_.allocation_size();
+}
+
+ROC_ATTR_NODISCARD bool LimitedPool::reserve(size_t n_objects) {
+    return pool_.reserve(n_objects);
+}
+
+void* LimitedPool::allocate() {
+    if (memory_limiter_.acquire(pool_.allocation_size())) {
+        return pool_.allocate();
```
gavv, line 29: If nested pool returns NULL, we should release acquired memory.

src/internal_modules/roc_core/memory_limiter.cpp
```diff
@@ -0,0 +1,73 @@
...
+    return false;
+}
+
+void MemoryLimiter::release(size_t num_bytes) {
+    if (num_bytes == 0) {
+        roc_panic("memory limiter (%s): tried to release zero bytes", name_);
+    }
+    size_t current = bytes_acquired_;
+    size_t next = bytes_acquired_ -= num_bytes;
+    if (next > current) {
```
gavv, line 61: We can implement it using only one atomic read:

```cpp
    const size_t next = bytes_acquired_ -= num_bytes;
    const size_t prev = next + num_bytes;
    if (next > prev) {
```

src/internal_modules/roc_core/memory_limiter.cpp
```diff
@@ -0,0 +1,73 @@
...
+            break;
+        }
+        if (bytes_acquired_.compare_exchange(current, next)) {
+            return true;
+        }
+        cpu_relax();
+    } while (true);
+    roc_log(LogError,
+            "memory limiter (%s): could not acquire %lu bytes due to limit, currently "
+            "acquired %lu, limit is %lu",
```
gavv, line 49: nitpick: usually in this project logs has the following style:

```
component: text message: var=value var=value...
```

for example:

```
memory limiter (%s): could not acquire bytes due to limit: requested=%lu acquired=%lu limit=%lu
```

It's not strictly necessary, but recommended.

Such logs are easier to grep:

* you can grep entire "text message" in source code
* you can grep "var=" in console output

src/internal_modules/roc_core/slab_pool.h
```diff
@@ -89,9 +89,9 @@ class SlabPool : public IPool, public NonCopyable<> {
                 flags) {
     }
 
-    //! Get size of objects in pool.
-    size_t object_size() const {
-        return impl_.object_size();
+    //! Get size of the allocation per object.
+    size_t allocation_size() const {
+        return SlotSize;
     }
```
gavv, line 95: Shouldn't it be `slot_size_` ?

src/internal_modules/roc_core/slab_pool.h
```diff
@@ -109,6 +109,12 @@ class SlabPool : public IPool, public NonCopyable<> {
         impl_.deallocate(memory);
     }
 
+    //! Get number of overhead bytes per allocation.
+    size_t overhead() {
+        return sizeof(SlabPoolImpl::SlotHeader) + sizeof(SlabPoolImpl::SlotCanary)
+            + sizeof(SlabPoolImpl::SlotCanary);
+    }
```
gavv, line 116: Shouldn't it be removed?

gavv (commented):
Thanks for update, a few more issues remaining.

src/internal_modules/roc_core/ipool.h
gavv: I think we can add `size` parameter to allocate(). Implementation can then check that requested size fits within slot size. This way we can be sure that any allocation on pool is safe (no matter via new or directly via allocate).

src/internal_modules/roc_core/limited_arena.cpp
nolan-veed: Good point. For sure.

nolan-veed: It would be hard to test this, unless I build a mock IArena that returns nullptr. Let me know if you desire this.

src/internal_modules/roc_core/memory_limiter.cpp
nolan-veed: Totally get the purpose. I wasn't sure we had a convention.

src/internal_modules/roc_core/slab_pool.h
nolan-veed: Having a think... I think there's some duplication.

src/internal_modules/roc_core/limited_arena.cpp
gavv: Up to you, 100% test coverage is not a requirement. It would be fine without a test too.

src/internal_modules/roc_core/slab_pool.h
nolan-veed: OK. I've gone with your suggestion.

PR #653, opened 2023-12-05
Clean up FEC tests

dshil:
- Use UNSIGNED_LONGS_EQUAL instead of CHECK for more clear test output
- Unify UNSIGNED_LONGS_EQUAL and LONGS_EQUAL usage


gavv:
LGTM, thanks!

Issue #246, opened 2019-09-02
libsndfile source and sink


gavv:
_Last revised: Oct 2023_

_**Update:** Issue is still relevant. Updated to reflect nowadays codebase._

## Intro

[libsndfile](http://www.mega-nerd.com/libsndfile/) is a minimalistic cross-platform C library allowing to read and write audio files in several audio formats.

Currently [roc-send](https://roc-streaming.org/toolkit/docs/manuals/roc_send.html) can read files using SoxSource and [roc-recv](https://roc-streaming.org/toolkit/docs/manuals/roc_recv.html) can write files using SoxSink. These source and sink implementation use [SoX](https://sox.sourceforge.net/libsox.html), which can work both with audio files and devices.

However, SoX has some disadvantages:

* it doesn't support Android
* it's audio device support has significant restrictions described in #231

In other issues like #231 and #232 we will implement sources and sinks for ALSA, PulseAudio, CoreAudio, and Android.

**It would be also nice to be able to read and write files using libsndfile instead of SoX:**

* It will allow us to have file support in roc-recv and roc-send on Android.

* It will also reduce number of dependencies because we anyway depend on libsndfile indirectly. SoX will become an optional dependency needed only for some exotic audio backends like sndio.

## Implementation

So in this issue, we should:

* implement `sndio::SndfileSource`, similar to [SoxSource](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_sndio/target_sox/roc_sndio/sox_source.h), but working only with files using libsndfile

* implement `sndio::SndfileSink`, similar to [SoxSink](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_sndio/target_sox/roc_sndio/sox_sink.h), but working only with files using libsndfile

* implement `sndio::SndfileBackend`, similar to [SoxBackend](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_sndio/target_sox/roc_sndio/sox_backend.h) and `sndio::PulseaudioBackend`

* register `sndio::SndfileBackend` in [BackendDispatcher](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_sndio/backend_dispatcher.h)

## Implementation notes

* SndfileSource, SndfileSink, and SndfileBackend should be placed under target_sndfile directory (similar to target_sox and target_pulseaudio)

[...]

Issue #608, opened 2023-10-10
Configurable encoding (format/rate/channels)


gavv:
Allow specifying format/rate/channels of:

- [x] input frames (sender API)
- [x] input device (sender CLI)
- [x] output frames (receiver API)
- [x] output device (receiver CLI)
- [x] network packets (sender API, receiver API)
- [x] network packets (sender CLI, receiver CLI)

Internals:

- [x] EncodingMap::add_encoding
- [x] Frame format (#547) 
- [x] Configurable channel set (#86)
- [x] Backends
   - [x] sox (rate/chans)
   - [x] pulseaudio (rate/chans/format)
   - [x] sndfile (rate/chans/format)
   - [x] wav (rate/chans/format)

API:

- [x] roc_packet_encoding
- [x] roc_format: native (raw)
- [x] roc_format: other formats
- [x] roc_channel_layout: mono, stereo, multitrack
- [x] roc_media_encoding
- [x] roc_sender_config: frame_encoding, packet_encoding
- [x] roc_receiver_config: frame_encoding
- [x] roc_context_register_encoding

CLI:

- [x] sample spec parser
- [x] packet encoding parser
- [x] remove --rate
- [x] add --io-encoding SPEC
- [x] add  --packet-encoding ID:SPEC
- [x] --print-supported

gavv:
* 7e8b0bd62671cf36e9c081e6f516055ee2e6ce40
* e1c741af5c48f1a5ba954b0bba7edc4037bed7bb
* 14d46f28d1f1400fac08033dc9ca8edc7b6d86c9
* 0548221ed2fdc5507ca11e225f48332285736397
* bcf594d44a5f498a3c65de5aa7a9bb575fa774dd
* 703583c956ceb0c88e9e221c63ed6eb4d1c0fbaa
* b5db264bed8e7c43388896d6cdc5c4968e387326
* a63b8ce4952daced9107bbbdce62b0d846319b33
* adc90bfe43595ceeb67698f0d9f5dcb6061e17e7
* 919873e94d43eaee06fe9885defbc1fa7f05075f

PR #660, opened 2023-12-16
Libsndfile Sink and Source 

Hrick87:
#246


Hrick87:
@gavv 

I was hoping you might have an idea as to why the macos13-x86_64/build-3rdparty.sh, macos13-x86_64/standard-build.sh and macos/universal-binaries.sh scripts are failing? The log and error code returned has not been helpful. Commit [779319d](https://github.com/roc-streaming/roc-toolkit/pull/660/commits/779319dcc9749160191bdcc4a20d373fdfd8d6d7) passed macos13-x86_64/standard-build.sh though I do not understand why. I was making small changes to try and understand what was happening. 

I apologize for all the commits of me testing minor changes to the scripts, but I was unable to figure out how to run the macOS scripts locally like I did with the other environments. It seems they do not rely on a docker image. Also, GitHub Actions runs the build scripts **MUCH** faster than my Virtual Box instance can, so its more efficient for me to do this. I can squash them once this is resolved if that is preferred. 

Once this is resolved,  my code will be ready for review.

Thank you!

gavv:
Sorry for delay. I triggered build on develop & master, and the same issue is reproduced there, so it's no specific to your PR.

https://github.com/roc-streaming/roc-toolkit/actions/runs/7242028057/job/19726898808

I'll take a look.

gavv:
> I apologize for all the commits of me testing minor changes to the scripts, but I was unable to figure out how to run the macOS scripts locally like I did with the other environments. Also, GitHub Actions runs the build scripts MUCH faster than my Virtual Box instance can. I can squash them once this is resolved if that is preferred.

No worries, github allows to squash commits into one when merging PR, I can just use that button.

gavv:
This issue with macos checks is gone.

I think the reason was this: https://github.com/actions/setup-python/issues/577

It's not reproducing now, but I've pushed a fix that I hope will prevent it from appearing again: c55443c1e6eace80d7e37bb586678bc165b0c115

Hrick87:
> This issue with macos checks is gone.
> 
> I think the reason was this: [actions/setup-python#577](https://github.com/actions/setup-python/issues/577)
> 
> It's not reproducing now, but I've pushed a fix that I hope will prevent it from appearing again: [c55443c](https://github.com/roc-streaming/roc-toolkit/commit/c55443c1e6eace80d7e37bb586678bc165b0c115)

The python issue was the only error message I was seeing on my end too, so the override argument should resolve it, but this is a bit beyond me. I have a lot to learn about containerization and GitHub Actions still. Anyways, as of now all the builds are passing, so I'm going to open up my PR for review. Thank you for your help.

3rdparty/SConscript
gavv, line 1: Currently, when we use --build-3rparty=pulseaudio, we also build sndfile:

```python
if 'pulseaudio' in autobuild_dependencies:
    ...
    env.BuildThirdParty(thirdparty_versions, 'sndfile')
```

Since sndfile was used only by pulseaudio, this solution was OK. However now there are more possible combinations:

* user may ask only sndfile - then we should build only sndfile
* user may ask only pulseaudio - then we should build sndfile + pulseaudio
* user may ask both sndfile and pulseaudio - then we should build sndfile + pulseaudio

Also we should make sure that we don't build sndfile twice.

To handle this, we can add something like this

```python
if 'pulseaudio' in external_dependencies:
    external_dependencies.add('sndfile')
```

to the beginning of SConscript, near this piece:

```python
# on Linux, PulseAudio needs ALSA
if meta.platform in ['linux'] and 'pulseaudio' in external_dependencies:
    external_dependencies.add('alsa')
```

Also we should ensure that building sndfile goes before building pulseaudio.

And we should remove `env.BuildThirdParty(thirdparty_versions, 'sndfile')` from section that builds pulseaudio.

3rdparty/SConscript
```diff
@@ -319,6 +319,50 @@ elif 'sox' in system_dependencies:
...
+if 'sndfile' in autobuild_dependencies:
+    sndfile_deps = []
+
+    if 'alsa' in autobuild_dependencies:
+        sndfile_deps += ['alsa']
+
+    if 'pulseaudio' in autobuild_dependencies:
+        sndfile_deps += ['pulseaudio']
+
+    env.BuildThirdParty(thirdparty_versions, 'sndfile', deps=sndfile_deps)
```
gavv, line 332: I think libsndfile does not depend on ALSA and pulseaudio?

Actually it's reverse: pulseaudio depends on libsndfile.

3rdparty/SConscript
```diff
@@ -319,6 +319,50 @@ elif 'sox' in system_dependencies:
...
+    if not 'alsa' in autobuild_dependencies:
+        for lib in [
+                'asound',
+                ]:
+            conf.CheckLib(lib)
+
+    if meta.platform in ['darwin']:
+        env.Append(LINKFLAGS=[
+            '-Wl,-framework,CoreAudio'
+        ])
```
gavv, line 345: Does libsndfile need CoreAudio?

3rdparty/SConscript
```diff
@@ -319,6 +319,50 @@ elif 'sox' in system_dependencies:
...
+elif 'sndfile' in system_dependencies:
+    conf = Configure(env, custom_tests=env.CustomTests)
+
+    if not conf.AddPkgConfigDependency('sndfile', '--cflags --libs', exclude_from_pc=True):
+        conf.env.AddManualDependency(libs=['sndfile'], exclude_from_pc=True)
+
+    if not is_crosscompiling:
+        if not conf.CheckLibWithHeaderExt(
+                'sndfile', 'sndfile.h', 'C',
+                expr='SNDFILE_LIB_VERSION_CODE >= SNDFILE_LIB_VERSION(1, 0, 28'):
```
gavv, line 358: Could you clarify why 1.0.28 is minimal version that we support?

Ideally we should be able to use the oldest version among our CI images. Quick googling suggests that ubuntu 14.04 used libsndfile 1.0.25, and this is probably the oldest one, but ideally we should check other distros too.

See also https://roc-streaming.org/toolkit/docs/portability/supported_platforms.html

SConstruct
```diff
@@ -209,6 +209,11 @@ AddOption('--disable-sox',
           action='store_true',
           help='disable SoX support in tools')
 
+AddOption('--disable-sndfile',
+          dest='disable_sndfile',
+          action='store_true',
+          help='disable SndFile support in tools')
```
gavv, line 215: I believe it's "libsndfile" or "sndfile", not "SndFile".

scripts/android_emu/run.sh
```diff
@@ -73,7 +73,7 @@ then
           --enable-tests \
           --disable-soversion \
           --disable-tools \
-          --build-3rdparty=libuv,openfec,openssl,speexdsp,cpputest
+          --build-3rdparty=libuv,openfec,openssl,speexdsp,cpputest,sndfile
```
gavv, line 76: Since --disable-tools is present, I think sndfile is not needed.

scripts/ci_checks/android/linux.sh
```diff
@@ -7,6 +7,6 @@ scons -Q \
       --disable-soversion \
       --disable-tools \
       --enable-tests \
-      --build-3rdparty=libuv,openfec,speexdsp,openssl,cpputest \
+      --build-3rdparty=libuv,openfec,speexdsp,openssl,cpputest,sndfile \
```
gavv, line 10: Same.

scripts/ci_checks/linux-checks/conditional-build.sh
```diff
@@ -11,7 +11,8 @@ scons -Q --enable-werror --build-3rdparty=all \
       --disable-openfec \
       --disable-speex \
       --disable-sox \
-      --disable-pulseaudio
+      --disable-pulseaudio \
+      --disable-sndfile \
```
gavv, line 16: nit: Trailing `\` should be removed.

scripts/ci_checks/linux-x86_64/alpine.sh
```diff
@@ -8,5 +8,5 @@ scons -Q \
       --enable-benchmarks \
       --enable-examples \
       --enable-doxygen \
-      --build-3rdparty=openfec,cpputest,google-benchmark \
+      --build-3rdparty=openfec,cpputest,google-benchmark,sndfile \
       test
```
gavv, line 12: Scripts in `ci_checks/linux-x86_64` and ``ci_checks/linux-checks` use docker images as described here: https://roc-streaming.org/toolkit/docs/development/continuous_integration.html

The idea is to test build on different distros. So instead of using --build-3rdparty for all dependencies, we use pre-installed dependencies from distro as much as possible, and use --build-3rdparty only for dependencies missing in distro.

That's why here you see, for example openfec (it's not present in alpine), but don't see, for example, speexdsp (it's present).

I think libsndfile is available in all or most distros. So, instead of adding it to --build-3rdparty, we should add it to docker images here: https://github.com/roc-streaming/dockerfiles

Here is example PR that added a dependency to all images: https://github.com/roc-streaming/dockerfiles/pull/12/files

scripts/ci_checks/macos/standard-build.sh
```diff
@@ -15,7 +15,7 @@ scons -Q \
       --enable-examples \
       --enable-debug \
       --sanitizers=all \
-      --build-3rdparty=openfec \
+      --build-3rdparty=openfec,sndfile \
```
gavv, line 18: On macos, libsndfile should be available via brew. So instead of adding sndfile to --build-3rdparty, we should add it to brew install command (above in this file).

gavv (commented):
Thanks for PR, that's big!

Here is the first part of PR, related to build system.

I'll review the remaining part in upcoming days.

src/tests/roc_sndio/target_sndfile/test_sndfile_sink.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+namespace {
+
+enum { FrameSize = 500, SampleRate = 44100, ChMask = 0x3 };
+
+core::HeapArena arena;
+
+} // namespace
+
+TEST_GROUP(sndfile_sink) {
```
gavv, line 26: I guess these tests are mostly (or completely) identical for all file backends.

It would be a big burden to maintain a copy for each backend, especially when we'll add more.

Can we instead replace tests for sox + test for sndfile with common backend-independent tests? Each test could query registered backends from BackendMap, select those which support files, and test each of them.

We use similar approach in other tests:

* https://github.com/roc-streaming/roc-toolkit/blob/master/src/tests/roc_fec/test_encoder_decoder.cpp#L124
* https://github.com/roc-streaming/roc-toolkit/blob/master/src/tests/roc_fec/test_writer_reader.cpp#L227
* https://github.com/roc-streaming/roc-toolkit/blob/master/src/tests/roc_audio/test_resampler.cpp#L303

gavv: And I think test_pump should use the same approach.

gavv (commented):
Thanks again for the patch! Here is the remaining review.

gavv:
image

Could you please rebase on fresh develop?

gavv:
Hi, I've pushed commit 03d29eb97211ca87593566998c5087590c1bae38 to develop branch that adds new fields sample_format and pcm_format to SampleSpec.

It mostly does not affect this PR, but SampleSpec ctor now has new parameter pcm_format, which you can just set to audio::Sample_RawFormat (which means that source/sink produces/consumes samples in "raw" format - 32-bit PCM floats).

So please rebase on fresh develop.

src/internal_modules/roc_sndio/target_sndfile/roc_sndio/sndfile_source.cpp
```diff
@@ -0,0 +1,456 @@
...
+                "sndfile source: can't open: unsupported # of channels: "
+                "expected=%lu actual=%lu",
+                (unsigned long)sample_spec_.num_channels(),
+                (unsigned long)sf_info_in_.channels);
+        return false;
+    }
+
+    if (sample_rate_ != 0) {
+        sf_info_in_.samplerate = sample_rate_;
+    }
```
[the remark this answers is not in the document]
Hrick87: This allows for a check to see if the user has set a sample-rate of their own using --rate. From my understanding sample_rate_, which is initialized by config.sample_spec.sample_rate(), will be 0 if this argument isn't passed. Otherwise, it will force the new sample rate to be what the user entered. Perhaps I am misunderstanding what the --rate argument and the tests for sample_rate_mismatch and sample_rate_auto are for, but without this line, those tests fail.

src/internal_modules/roc_sndio/target_sndfile/roc_sndio/sndfile_sink.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+    sndfile_output_ = sf_open(path, SFM_WRITE, &sf_info_out_);
+    if (!sndfile_output_) {
+        roc_log(LogDebug, "sndfile sink: can't open: driver=%s path=%s", driver, path);
+        return false;
+    }
+
+    sf_command(sndfile_output_, SFC_SET_UPDATE_HEADER_AUTO, NULL, SF_TRUE);
+
+    if (in_rate != 0 && in_rate != out_rate) {
```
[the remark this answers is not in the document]
Hrick87: I've erased the check for if (in_rate != 0 && in_rate != out_rate). This is never a problem. sndfile in general is capable of changing the sample rate of a file, but it would not be happening here, nor would it cause a problem. Thanks.

src/internal_modules/roc_sndio/target_sndfile/roc_sndio/sndfile_sink.cpp
```diff
@@ -0,0 +1,360 @@
...
+            frame_data++;
+            frame_size--;
+        }
+
+        if (buffer_pos > 0) {
+            if (sf_write_float(sndfile_output_, buffer_data, (sf_count_t)buffer_size_)
+                    == 0
+                && frame_size > 0) {
+                roc_panic(
+                    "sndfile sink: Unable to write entire frame to file. Reached eof");
```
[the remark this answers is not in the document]
Hrick87: From my understanding sf_write_float has no error return as it only returns the number of items it was able to read, whether that be 0 or any other number. I can use sf_strerror(SNDFILE * file_) here if you like, but I think it will always return a SF_ERR_NO_ERROR = 0 message.

3rdparty/SConscript
Hrick87: @gavv the correct way to get libsndfile's version number is sf_version_string(). That was a copy and paste error on my part, sorry! 

However, I am struggling with how I can compare two string versions against each other in a similar fashion to SOX or libuv in a neat single line, since there is no macro that returns the version number as something immediately comparable. I know several long multi-line methods of parsing and converting a version strings to individual numbers for comparison in C, but I'm unsure if that would be okay or where/how to place that in the script.

Also you are correct, after researching all the available platforms version 1.0.25 is the oldest we must support.

Hrick87:
@gavv I was hoping to get some advice on one of the tests. I am trying to remove the use of core::Array<audio::sample_t> buffer_ and size_t buffer_size_, and am working on refactoring the read() function for sndfile_source. 

I am failing this test at the line ```CHECK(!sndfile_source.read(frame));```
        
```C++
TEST(sndfile_source, eof_restart) {

    {
        test::MockSource mock_source;
        mock_source.add(FrameSize * NumChans * 2);

        SndfileSink sndfile_sink(arena, sink_config);
        CHECK(sndfile_sink.open(NULL, file.path()));

        Pump pump(buffer_factory, mock_source, NULL, sndfile_sink, FrameDuration,
                  SampleSpecs, Pump::ModeOneshot);
        CHECK(pump.is_valid());
        CHECK(pump.run());
    }

    SndfileSource sndfile_source(arena, source_config);

    CHECK(sndfile_source.open(NULL, file.path()));

    audio::sample_t frame_data[FrameSize * NumChans] = {};
    audio::Frame frame(frame_data, FrameSize * NumChans);

    for (int i = 0; i < 3; i++) {
        CHECK(sndfile_source.read(frame));
        CHECK(sndfile_source.read(frame));
        CHECK(!sndfile_source.read(frame)); //This line fails

        CHECK(sndfile_source.restart());
    }
}
```

My understanding of the test is that the second ```read()``` call is expected to reach eof and set ```eof_``` to true, so that when the third ```read()``` call happens in the for loop, the call will return false. I think the issue is that the second ```read()``` call is not reaching the ```eof_ = true``` line and so the third call fails. I cannot find the issue in my code that is causing this. I've been stuck on this problem for quite some time.

I currently have:
```C++
bool SndfileSource::read(audio::Frame& frame) {
    roc_panic_if(!valid_);

    if (eof_) {
        return false;
    }

    if (!file_) {
        roc_panic("sndfile source: read: non-open input file");
    }
    audio::sample_t * frame_data = frame.samples();
    sf_count_t frame_left = (sf_count_t)frame.num_samples();
    sf_count_t buffer_size = frame_left;

    while (frame_left > 0) {
        sf_count_t n_samples = sf_read_float(file_, frame_data, buffer_size);

        printf("read %ld bytes\n", n_samples);

        if (n_samples == 0) {
            printf("eof, n_samples = 0\n");
            roc_log(LogDebug, "sndfile source: got eof from sndfile");
            eof_ = true;
            break;
        }
        
        frame_left -= n_samples;
    }

    if (frame_left == (sf_count_t)frame.num_samples()) {
        return false;
    }

    if (frame_left > 0) {
        memset(frame_data, 0, (size_t)frame_left * sizeof(audio::sample_t));
    }

    return true;
}
```

3rdparty/SConscript
gavv: I think let's just skip version check for simplicity, i.e. omit `expr=`.

src/internal_modules/roc_sndio/target_sndfile/roc_sndio/sndfile_sink.cpp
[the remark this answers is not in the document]
gavv: Got it.

src/internal_modules/roc_sndio/target_sndfile/roc_sndio/sndfile_sink.cpp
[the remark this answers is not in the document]
gavv: Actually it seems it can raise quite a lot of errors: https://github.com/libsndfile/libsndfile/blob/e5ee50fbda1b9049a45fc65d06c34825feb4f237/src/sndfile.c#L2463

Most likely errors for us are failure to write to a file (e.g. disk is full, unmounted, etc).

You're right, seems we need sf_strerror().

I'd do something like this:

```
if (sf_write_float(...) != frame_size || sf_strerror(...) != 0) {
    // error
}
```

This way we'll catch all unusual situations. If everything is good, we should always write as many samples as requested, and error flag should not be set.

src/internal_modules/roc_sndio/target_sndfile/roc_sndio/sndfile_source.cpp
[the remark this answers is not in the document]
gavv: I see... Yes, --rate argument forces specific rate. However, it's intended only for devices, e.g. when you open ALSA or PulseAudio device, it often supports multiple rates natively, and you may want to select one of them.

For files, it doesn't make much sense to force any rate. File has fixed rate. Pipeline will do resampling if pipeline rate differs from file rate.

So I think what we need to do is:

* get sample rate of the file from libsndfile
* report it via sample_spec()
* when opening file, also do a check:
  * if rate in config is zero, any rate is ok, so no problems
  * if rate in config is non-zero, fail if it differs from file rate, because user asks for specific rate, but we can't do it

gavv:
```cpp
sf_count_t n_samples = sf_read_float(file_, frame_data, buffer_size);
```

First, seems that you request the whole buffer_size from sndfile, but I think you need to request only frame_left?

Second, I think you don't need a loop at all?

Maybe you can do something like:

```cpp
samples_per_ch = frame.num_samples() / num_channels;

n_read = sf_read_float(file_, frame.samples(), samples_per_ch);

if (n_read == 0) {
    eof = true;
}

if (n_read < samples_per_ch) {
    memset(frame.samples() + n_read * num_channels, 0,
        (samples_per_ch - n_read) * num_channels * sizeof(sample_t));
}

return !eof;
```

BTW note that sf_read_float() works with number of samples per channel (a.k.a. number of frames, not to be confused with audio::Frame in roc) and frame.num_samples() is number of samples for all channels.

Also I think we need to check for error here too using sf_error_number(). If there is error, we can also add a panic with TODO comment.

gavv:
If you'll have troubles you could push your code somewhere (to this PR or other branch) and I'll take a look.

Hrick87:
> ```c++
> sf_count_t n_samples = sf_read_float(file_, frame_data, buffer_size);
> ```
> 
> First, seems that you request the whole buffer_size from sndfile, but I think you need to request only frame_left?
> 
> Second, I think you don't need a loop at all?
> 
> Maybe you can do something like:
> 
> ```c++
> samples_per_ch = frame.num_samples() / num_channels;
> 
> n_read = sf_read_float(file_, frame.samples(), samples_per_ch);
> 
> if (n_read == 0) {
>     eof = true;
> }
> 
> if (n_read < samples_per_ch) {
>     memset(frame.samples() + n_read * num_channels, 0,
>         (samples_per_ch - n_read) * num_channels * sizeof(sample_t));
> }
> 
> return !eof;
> ```
> 
> BTW note that sf_read_float() works with number of samples per channel (a.k.a. number of frames, not to be confused with audio::Frame in roc) and frame.num_samples() is number of samples for all channels.
> 
> Also I think we need to check for error here too using sf_error_number(). If there is error, we can also add a panic with TODO comment.

Your code greatly simplified things for me, enough to see the problem! The check should not be ```if(n_samples == 0)``` but instead ```if(n_samples < frame_left)``` because eof can be signified by 0 or less than read's count argument. I was not accounting for that possibility. Its been fixed. Thank you!

gavv:
> The check should not be if(n_samples == 0) but instead if(n_samples < frame_left) because eof can be signified by 0 or less than read's count argument

Oh, really. Great!

Hrick87:
@gavv I've been having some trouble refactoring the test cases to be done in a for loop. I'm only attempting to change one test as of now in test_pump.cpp. The print statements and new name method I implemented in IBackend are just for debugging purposes. This refactoring:

```C++
TEST(pump, write_read) {
{ 

    enum { NumSamples = BufSize * 10 };
    
    for(size_t n_backend = 0; n_backend < BackendMap::instance().num_backends(); n_backend++){
        test::MockSource mock_source;
        mock_source.add(NumSamples);
        

        IBackend &backend = BackendMap::instance().nth_backend(n_backend);
        printf("Currently on: %s\n", backend.name());
        fflush(stdout);
        
        {
            IDevice *backend_device = backend.open_device(DeviceType_Sink, DriverType_File, "wav", file.path(), config, arena);
            if(backend_device == NULL){
                printf("Failing sink test: %s\n", backend.name());
                fflush(stdout);
            }
            else{
                printf("Passing sink test: %s\n", backend.name());
                fflush(stdout);
                ISink *backend_sink = backend_device->to_sink();
                Pump pump(buffer_factory, mock_source, NULL, *backend_sink, BufDuration, SampleSpecs,
                        Pump::ModeOneshot);
                CHECK(pump.is_valid());
                CHECK(pump.run());

                CHECK(mock_source.num_returned() >= NumSamples - BufSize);
            }
        }
        printf("File path: %s\n", file.path());
        fflush(stdout);
        
        IDevice *backend_device = backend.open_device(DeviceType_Source, DriverType_File, "wav", file.path(), config, arena);
        if(backend_device == NULL){
            printf("Failing source test: %s\n", backend.name());
            fflush(stdout);
                continue;
        }

        printf("Passing source test: %s\n\n", backend.name());
        
        ISource * backend_source = backend_device->to_source();

        test::MockSink mock_writer;

        Pump pump(buffer_factory,
                *backend_source,
                NULL,
                mock_writer,
                BufDuration,
                SampleSpecs,
                Pump::ModePermanent);
        CHECK(pump.is_valid());
        CHECK(pump.run());

        mock_writer.check(0, mock_source.num_returned());
    }
}
} // namespace roc
```

results in this output:
```
Currently on: wav
Passing sink test: wav
File path: /tmp/rocr4GtK2/test.wav
Failing source test: wav
Currently on: PulseAudio
Failing sink test: PulseAudio
File path: /tmp/rocIyNcKx/test.wav
Failing source test: PulseAudio
Currently on: sndfile
Passing sink test: sndfile
File path: /tmp/rocXWcJjr/test.wav
Passing source test: sndfile

Currently on: SoX
Passing sink test: SoX
File path: /tmp/rocNy0rfi/test.wav
Passing source test: SoX


src/tests/roc_sndio/test_pump.cpp:58: error: Failure in TEST(pump, write_read)
src/tests/roc_sndio/test_helpers/mock_sink.h:84: error:
        expected <0>
        but was  <0.921875> threshold used was <0.0001>
```

If I disable SoX in my build, the test passes. I also tried reverting the test cases back to their individual formats as I have them in my initial pull request and both sndfile and SoX pass. This leads me to believe something about how my for loop works is causing SoX to fail. Any ideas would be much appreciated!

gavv:
The test will pass if you do this change:

```cpp
// from:
ISink *backend_sink = backend_device->to_sink();
// to:
core::ScopedPtr<ISink> backend_sink(backend_device->to_sink(), arena);
```

and:

```cpp
// from:
ISource * backend_source = backend_device->to_source();
// to:
core::ScopedPtr<ISource> backend_source(backend_device->to_source(), arena);
```

Sox sink uses buffering, and last part of buffer may not be flushed until destructor is called. By using ScopedPtr, we ensure that destructor is called in the end of the block.

Actually, I think this behavior is very unexpected. I think we should add ISink::flush() and modify Pump to call flush() when it exits from loop. For most sinks, flush() will be just no-op, but sinks that use buffering will implement it.

I've created an issue: https://github.com/roc-streaming/roc-toolkit/issues/703

Note that, apparently, you can't use:

```cpp
core::ScopedPtr<IDevice> backend_device(backend.open_device(...))
```

and instead have to use `core::ScopedPtr<ISink>` after you call `to_sink()`. That's because ISink/ISource uses virtual inheritance to derive IDevice, and pointer to IDevice has different address than pointer to ISink/ISource.

Usually we create sinks/sources via BackendDispatcher, which returns ISink/ISource, not IDevice, so this problem usually not present. However this test is a bit more low-level as it uses BackendMap directly,

Another note: in your test, WAV backend is failing because WavSource forbids providing non-empty sample spec. I guess we should use empty sample spec when constructing WavSource/SoxSource/SndfileSource, and SndfileSource should forbid non-empty spec as well.

It corresponds to this previous discussion from chat, because --rate is stored in sample spec:

> given it second thought... I think we could simplify convention and just forbid using --rate argument with files at all, and just fail if rate is non-zero. (we can do it in sox too, but only in case if we're opening a file)

And final note - instead of skipping backend if it fails to open file, it'd be better to check if wav is reported by discover_drivers(), and if yes, require backend to be able to open file and fail the test if open fails. Otherwise the test can succeed if backend supports wav but there is a bug in open().

BTW, nice idea to add name() to backends!

Hrick87:
> 
> And final note - instead of skipping backend if it fails to open file, it'd be better to check if wav is reported by discover_drivers(), and if yes, require backend to be able to open file and fail the test if open fails. Otherwise the test can succeed if backend supports wav but there is a bug in open().
> 


Thank you, I was trying to use scoped_ptr with IBackend and could not figure it out. This explains a lot!

As for your suggestion with discover_drivers(), I am confused. discover_drivers() is a void function and will panic on failure to push a driver into the driver_list. I don't see how it can be used to do that. What do you mean by "if wav is reported by discover_drivers()"?

gavv:
> Should the auto sample rate of sndfile_sink be 48000?

I think right now 44100 is a better default for us, for two reasons:

* for now, CLI tools use static (predefined) RTP payload types and can't use dynamic types; there are static types for 44100, but not for 48k, so if sink/source is 48k, pipeline will do resampling (C API allows to use custom rates, but CLI tools can't do it yet)

* consumer sound cards like Intel HDA usually support 44100 natively

So if 44100 is default, then on the typical case when the user runs roc-recv/roc-send on laptop or raspberry, the whole chain between file on on side and speakers/mic on the other won't need resampling.

> Should I have a guard in place for number of channels and format as well?

Yes, we need something like this:

https://github.com/roc-streaming/roc-toolkit/blob/e86ccda389feaa0674af925d0590ebd5492142ea/src/internal_modules/roc_sndio/wav_sink.cpp#L44-L57

(Since we're saving sample_spec, I suggest to apply defaults to sample_spec before copying values to file_info_).

> Is setting the format to a signed 32 bit integer PCM format for all writes ok?

Yes, ideally I think we need to do this:

* if format is not specified in sample spec, use whatever format we think is good
* if format is specified, query sndfile what subformats are supported for given file type, and map sample spec format to it (e.g. wav file type supports floats, integers, etc); if format can't be mapped, return error

For now CLI tools never set format in sample spec, so we don't need to handle it in this PR.

We have two issues for handling format in backends: #696, #608.

Would be nice if you add a comment like `TOTO(gh-696): map format from sample_space`.

>  Sndfile_sink currently uses the SF_FORMAT_PCM_32 as its PCM format for all writes. I chose this because it has the highest compatibility with most file extensions versus the highest level of quality achievable. In map_to_sndfile() I have a section of code that essentially accounts for the only two file extensions that need a different sub format, and they first attempt the higher quality, then the lower quality, before giving up and returning false. (FORMAT_COUNT is 2).

> This feels hacky and probably not intended.

Actually I think this logic makes sense, though maybe we can modify it a bit: instead of hard-coding specific types like SF_FORMAT_XI, we instead can create a static array of subformats sorted from high priority (like float32 as you said) to low priority (like smaller integers).

Then, no matter of the file type, we can first try use format without subformat, and then try each subformat from the list, until first successful attempt.

For now this can be the only behavior we support. Later this behavior will be used only if format is not specified in sample spec.

> Is this acceptable until we have is_state() implemented?

I think that's OK.

> Is noop even necessary in source and sink tests anymore/is this proper implementation if it is?

The test probably still makes sense, it checks that after we created a file using sink, we can *open* it using source.

So yeah, maybe just rename it to something like `write_open` or something.

> I think this code is unnecessary, can I remove it?

I think we need the following behavior: if sndfile returns less samples than requested, we should return a frame with what we got from sndfile, padded with zeros until frame end. On the *next* call we should return false.

Why:

* if we return false immediately, we'll loose last chunk instead of sending & playing it; that's not cool
* currently we can't report partial read, so we have to pad chunk with zeros up to the full frame size

I think same behavior is implemented in Sox.

gavv:
> How can I resolve the build error for the two failing ARM architecture checks?

If you look here: https://github.com/roc-streaming/roc-toolkit/actions/runs/8101072211/job/22140438520?pr=660

You can see:

```
     BUILD   build/3rdparty/arm-linux-gnueabihf/gcc-4.9.4-release/sndfile-1.0.25
[download] http://www.mega-nerd.com/libsndfile/files/libsndfile-1.0.25.tar.gz
[unpack] libsndfile-1.0.25.tar.gz
```

Now we can go to sndfile sources.

For 1.0.25: https://github.com/libsndfile/libsndfile/blob/1.0.25/src/sndfile.h.in#L318

For master: https://github.com/libsndfile/libsndfile/blob/master/include/sndfile.h#L368

So actually it's not int64_t in 1.0.25, and I guess this is a bug that was fixed in some later version.

So I think we can just bump libsndile version to the lowest one that doesn't have the bug.

Again, from this link: https://github.com/roc-streaming/roc-toolkit/actions/runs/8101072211/job/22140438520?pr=660

You can copy the very first command:

```
scripts/ci_checks/docker.sh rocstreaming/toolchain-arm-linux-gnueabihf:gcc-4.9 scripts/ci_checks/linux-arm/arm-linux-gnueabihf-gcc-4.9.sh
```

This way you can run exact same CI check locally (in docker). So you can try different sndfile versions and see which one would work.

> My research has been unhelpful, as many suggestions were to typedef the signed long variable type to int64_t to stay consistent between 32 bit and 64 bit platforms, but obviously that's already being done. Let me know if you have any ideas.

Yes, that platform is 32-bit, it uses 32-bit pointers and size_t, but it still supports 64-bit integers and int64_t, they're just not as efficient as 32-bit.

So it looks like sndfile is inconsistent when it first resolves sf_count_t to a 32-bit type but then requires it to be 64-bit. The latest version that just uses int64_t should work fine on 32-bit platform.

gavv:
> How can I resolve the build error for the two failing macOS checks?

>  It seems to me that this is a know issue with libsndfile version 1.0.25 on macOS that was patched in later versions of sndfile. It also looks like some people have created their own patches for version 1.0.25 (see second link). I am struggling with how I can go about fixing this issue within SConscript using this information. Let me know if you have any ideas.

If it's fixed in later version, let's just bump to minimum version that works properly.

If you know what's that version, let's try it. If you need to do some testing and don't have macOS, let me know, I could do the testing. (I hate that you can't cross-compile to mac from other systems...)

gavv:
> Any other concerns?

Three small comments.

---

```cpp
        if ((strcmp(format_info.extension, "wav") == 0)
            || (strcmp(format_info.extension, "mat") == 0)) {
```

Seems that we can remove this check and just always use table? (file_type_map) If there is an entry for the format, use table, otherwise use format_info.extension.

This way the knowledge about extensions & driver will be completely isolated in the table.

---

And the second note, `map_to_sndfile` looks a bit too long, maybe we can split it into two functions? E.g. one to select format, another to select subformat.

---

Finally, regarding FileMap:

* it would be great to add a comment briefly explaining *why* we need it and who this mappings are used
* we don't use `#pragma once`
* I think it makes sense to move table definition to a .cpp file so that it's not duplicated in every .cpp that uses this header

Hrick87:
> > How can I resolve the build error for the two failing macOS checks?
> 
> > It seems to me that this is a know issue with libsndfile version 1.0.25 on macOS that was patched in later versions of sndfile. It also looks like some people have created their own patches for version 1.0.25 (see second link). I am struggling with how I can go about fixing this issue within SConscript using this information. Let me know if you have any ideas.
> 
> If it's fixed in later version, let's just bump to minimum version that works properly.
> 
> If you know what's that version, let's try it. If you need to do some testing and don't have macOS, let me know, I could do the testing. (I hate that you can't cross-compile to mac from other systems...)

version 1.0.26 resolved both build issues! Everything you mentioned has been fixed. Ready for a (hopefully) final review and merge :)

gavv:
Quick question.

In SndfileSink::open, you have:

```cpp
sample_spec_.set_sample_rate((unsigned long)file_info_.samplerate);
```

In which case opening file for *writing* may change requested rate?

(Note that the rate is never zero, we explicitly set it to 44100 in ctor if it's zero).

gavv:
And another one.

Is there a reason to have this code:

```cpp
    if (!eof_) {
        if (!seek_(0)) {
            roc_log(LogError, "sndfile source: seek failed when restarting");
            return false;
        }
    } else {
        if (file_) {
            close_();
        }

        if (!open_(path_)) {
            roc_log(LogError, "sndfile source: open failed when restarting");
            return false;
        }
    }
```

which may reopen file, instead of just always using seek?

gavv:
> which may reopen file, instead of just always using seek?

Oh, I see, some file types are unseekable.

gavv:
Thanks a lot for PR! That was big!

I have pushed a few follow-up patches:

* ebd0c7a2a468069f0df73364ff6dff3eb1c5c056
* 19c77d3369ae9cdd2e31ba77604fdf334c22aa29 - cosmetic changes
* 35c87b90589902d89984c347fb2e665348d6271a - fix a few bugs in sndfile backend
* 608963db5478d74d8871154a0c60314ee5460e21 - cleanup sox backend a bit
* 71b47492b94bfb54d0ad26a79e8b930f21047a01 - better channel count handling

PR #661, opened 2023-12-18
Fix dead example links in docs

angelthorns:
The code links on this page https://roc-streaming.org/toolkit/docs/api/examples.html were dead


gavv:
Thank you!

PR #662, opened 2023-12-21
Parsing float for duration and size

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/654

# What

* Updated the functions to use strod which parses a double.
  * After multiplying with multiplier, used round() function. 
  * Ensured that the nanoseconds_t and size_t are not overflowed, by checking difference.
* Updated tests

# Testing

There are tests.

### DIFF
--- a/docs/sphinx/manuals/roc_copy.rst
+++ b/docs/sphinx/manuals/roc_copy.rst
@@ -59,6 +59,12 @@ The path component of the provided URI is `percent-decoded <https://en.wikipedia
 
 For example, the file named ``/foo/bar%/[baz]`` may be specified using either of the following URIs: ``file:///foo%2Fbar%25%2F%5Bbaz%5D`` and ``file:///foo/bar%25/[baz]``.
 
+Time units
+----------
+
+*TIME* should have one of the following forms:
+  123ns; 1.23us; 1.23ms; 1.23s; 1.23m; 1.23h;
+
 EXAMPLES
 ========
 
--- a/src/internal_modules/roc_core/parse_units.cpp
+++ b/src/internal_modules/roc_core/parse_units.cpp
@@ -57,30 +57,38 @@ bool parse_duration(const char* str, nanoseconds_t& result) {
     }
 
     if (str == suffix) {
-        roc_log(
-            LogError,
-            "parse duration: invalid format, missing number, expected <number><suffix>");
+        roc_log(LogError,
+                "parse duration: invalid format, missing number, expected "
+                "<float><suffix>");
         return false;
     }
 
     if (!isdigit(*str) && *str != '-') {
-        roc_log(
-            LogError,
-            "parse duration: invalid format, not a number, expected <number><suffix>");
+        roc_log(LogError,
+                "parse duration: invalid format, not a number, expected "
+                "<float><suffix>");
         return false;
     }
 
     char* number_end = NULL;
-    long number = strtol(str, &number_end, 10);
+    const double number = strtod(str, &number_end);
 
-    if (number == LONG_MAX || number == LONG_MIN || !number_end || number_end != suffix) {
+    if ((number == 0. && str == number_end) || !number_end || number_end != suffix) {
         roc_log(LogError,
                 "parse duration: invalid format, can't parse number, expected "
-                "<number><suffix>");
+                "<float><suffix>");
         return false;
     }
 
-    result = nanoseconds_t(number) * multiplier;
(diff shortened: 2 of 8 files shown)


src/internal_modules/roc_core/parse_units.cpp
```diff
@@ -57,30 +57,38 @@ bool parse_duration(const char* str, nanoseconds_t& result) {
     }
 
     if (str == suffix) {
-        roc_log(
-            LogError,
-            "parse duration: invalid format, missing number, expected <number><suffix>");
+        roc_log(LogError,
+                "parse duration: invalid format, missing number, expected "
+                "<number>[.<number>]<suffix>");
```
gavv, line 62: nitpick: maybe `<float>` would be more intuitive than `<number>[.<number>]`?

src/internal_modules/roc_core/parse_units.cpp
```diff
@@ -57,30 +57,38 @@ bool parse_duration(const char* str, nanoseconds_t& result) {
...
+    if ((number == 0. && str == number_end) || !number_end || number_end != suffix) {
         roc_log(LogError,
                 "parse duration: invalid format, can't parse number, expected "
-                "<number><suffix>");
+                "<number>[.<number>]<suffix>");
         return false;
     }
 
-    result = nanoseconds_t(number) * multiplier;
+    double number_multiplied = round(number * (double)multiplier);
```
gavv, line 83: nit: we can use const here.

src/tools/roc_copy/cmdline.ggo
gavv, line 1: Could you please also update man pages?

https://github.com/roc-streaming/roc-toolkit/tree/develop/docs/sphinx/manuals

https://roc-streaming.org/toolkit/docs/manuals.html

gavv (commented):
Thanks!

gavv:
> 🤖 Upon creation, pull request description does not have a link to an issue. If there is a related issue, please add it to the description using any of the [supported formats](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls).

I'll try to fix the linter...

docs/sphinx/manuals/roc_copy.rst
```diff
@@ -59,6 +59,13 @@ The path component of the provided URI is `percent-decoded <https://en.wikipedia
 
 For example, the file named ``/foo/bar%/[baz]`` may be specified using either of the following URIs: ``file:///foo%2Fbar%25%2F%5Bbaz%5D`` and ``file:///foo/bar%25/[baz]``.
 
+Time units
+----------
+
+*TIME* should have one of the following forms:
+  123ns, 123us, 123ms, 123s, 123m, 123h
+  1.23us; 1.23ms; 1.23s; 1.23m; 1.23h;
```
gavv, line 67: I think we can keep only floats examples, I think it's pretty obvious that if floats are supported, then integers are supported too.

Applies to all man pages and `.ggo` files.

src/tests/roc_core/test_parse_units.cpp
```diff
@@ -11,6 +11,8 @@
 #include "roc_core/parse_units.h"
 #include "roc_core/time.h"
 
+#include <inttypes.h>
```
gavv, line 14: IIRC it's POSIX header, and these tests are supposed to work on other OSes too (in future). It seems currently it's not used anyway?

src/tests/roc_core/test_parse_units.cpp
```diff
@@ -87,18 +140,52 @@ TEST(parse_units, parse_size) {
...
+    CHECK(result == 1153434);
+
+    CHECK(parse_size("1.1G", result));
+    CHECK(result == 1181116006);
 }
 
 TEST(parse_units, parse_size_overflows_due_to_multiplier) {
     char s[32];
-    snprintf(s, 32, "%zuK", SIZE_MAX - 1);
+    snprintf(s, 32, "%zuK", SIZE_MAX);
```
gavv, line 188: `%z` is not portable, you can use `%llu` with `(unsigned long long)SIZE_MAX`

src/tests/roc_core/test_parse_units.cpp
nolan-veed: It's a stray header, I introduced to get access to PRId64 macros, etc. But, then gave up.

gavv:
LGTM, thanks!

PR #664, opened 2023-12-27
Seqlock implementation extract

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/599

# What

* Created a SeqlockImpl class and used it in SeqLock.

# Testing

There are tests.

### DIFF
--- a/src/internal_modules/roc_core/seqlock_impl.cpp
+++ b/src/internal_modules/roc_core/seqlock_impl.cpp
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2020 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_core/seqlock_impl.h"
+#include "roc_core/atomic_ops.h"
+#include "roc_core/cpu_instructions.h"
+
+namespace {
+// We use hand-rolled loop instead of memcpy() or default (trivial) copy constructor
+// to be sure that the copying will be covered by our memory fences. On some
+// platforms, memcpy() and copy constructor may be implemented using streaming
+// instructions which may ignore memory fences.
+void volatile_copy(volatile void* dst, const volatile void* src, size_t val_size) {
+    volatile char* dst_ptr = reinterpret_cast<volatile char*>(dst);
+    const volatile char* src_ptr = reinterpret_cast<const volatile char*>(src);
+
+    for (size_t n = 0; n < val_size; n++) {
+        dst_ptr[n] = src_ptr[n];
+    }
+}
+
+} // namespace
+
+namespace roc {
+namespace core {
+
+SeqlockImpl::SeqlockImpl()
+    : ver_(0) {
+}
+
+seqlock_version_t SeqlockImpl::version() const {
+    return AtomicOps::load_seq_cst(ver_);
+}
+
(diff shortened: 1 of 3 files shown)


nolan-veed:
For some reason, my tests are hanging here...
image

nolan-veed:
> For some reason, my tests are hanging here...

Oh wait... the Timer is using SeqLock, so it's clearly uncovered a bug.

src/internal_modules/roc_core/seqlock_impl.h
```diff
@@ -0,0 +1,57 @@
...
+    void wait_load(void* value, seqlock_version_t& ver) const;
+
+private:
+    static void
+    volatile_copy_(volatile void* dst, const volatile void* src, size_t val_size);
+
+    bool try_load_(void* value, seqlock_version_t& ver) const;
+
+    void* val_;
+    size_t val_size_;
```
gavv, line 50: If we will store pointer and size in members, then every call (try_store, try_load, etc) will involve two additional reads from memory, and the whole operation will depend on these reads.

If, instead, we will pass `&val` and `sizeof(val)` to these methods as arguments, then chances are that no memory reads are needed: `&val` can be calculated as `this + offset`, and this is likely already in register, and `sizeof(val)` is a constant. And then on many architectures like x86_64 ARM function arguments will be passed via registers.

In our code base Seqlock is often used for small frequently used variables, like counters. E.g. seqlock is a replacement for 64-bit atomics (because 64-bit atomics are not available everywhere). So it makes sense to keep seqlock overhead low and comparable with a regular atomic.

src/internal_modules/roc_core/seqlock_impl.h
```diff
@@ -0,0 +1,57 @@
...
+
+    //! Try to load value and version.
+    bool try_load_repeat(void* value, seqlock_version_t& ver) const;
+
+    //! Load value and version.
+    void wait_load(void* value, seqlock_version_t& ver) const;
+
+private:
+    static void
+    volatile_copy_(volatile void* dst, const volatile void* src, size_t val_size);
```
gavv, line 45: I guess we can move this method into .cpp file in anon namespace.

src/internal_modules/roc_core/seqlock_impl.cpp
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_core/seqlock_impl.h"
+
+#include "roc_core/atomic_ops.h"
+#include "roc_core/cpu_instructions.h"
```
gavv, line 12: nit: extra empty line.

gavv (commented):
Thanks!

src/internal_modules/roc_core/seqlock_impl.cpp
nolan-veed: Oh... we don't want to include the header related to the cpp file first?

gavv: Oh, it was intentional. Yes, current "defacto" code style uses:

```
<one block with stdlib headers>

<one block with external libs headers>

<one block with roc headers>
```

e.g.:

```
#include <stdio.h>
#include <stdlib.h>

#include <uv.h>

#include "roc_core/array.h"
#include "roc_core/time.h"
```

gavv:
Thanks!

Small refactoring: 3ebb8c86f5797b6d23dced13f28b18966412600d (see commit message for details)

PR #665, opened 2023-12-27
Ubuntu 24.04 builds

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/634

This adds the ci builds.

# What

* Wrote script for Ubuntu 24.04.
  * Supported the 2 latest g++ and clang versions for building.
* Updated build.yml.

# Testing

Should run on CI.

### DIFF
--- a/scripts/ci_checks/linux-x86_64/ubuntu-24.04.sh
+++ b/scripts/ci_checks/linux-x86_64/ubuntu-24.04.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+for comp in gcc-13 clang-15 clang-17
+do
+    scons -Q \
+          --enable-werror \
+          --enable-tests \
+          --enable-benchmarks \
+          --enable-examples \
+          --build-3rdparty=openfec \
+          --compiler=${comp} \
+          test
+done
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - script: linux-x86_64/ubuntu-24.04
+            image: rocstreaming/env-ubuntu:24.04
+
           - script: linux-x86_64/ubuntu-22.04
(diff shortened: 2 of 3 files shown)


scripts/ci_checks/linux-x86_64/ubuntu-24.04.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+for comp in gcc-12 gcc-13 clang-16 clang-17
```
gavv, line 5: Let's remove gcc-12 (already covered by other image) and replace clang 16 with 15, because clang 15 is available in more ubuntu versions than 16 and so is more important to cover I think.

gavv:
As mentioned in https://github.com/roc-streaming/dockerfiles/pull/16, please also update https://roc-streaming.org/toolkit/docs/development/continuous_integration.html

gavv:
We forgot to add 24.04 to CI, done: https://github.com/roc-streaming/dockerfiles/commit/aecc5adfcfa5bb4f8608f479441b5a2aa64f0efd

(most images build all tags automatically, however for ubuntu each tag should be added manually, because we need to split ubuntu images into multiple CI jobs to avoid exceeding github limits for maximum job times)

nolan-veed:
> We forgot to add 24.04 to CI...

Don't think I have perms to run the jobs again.

gavv:
There are two other failures:

1. Formatting. The problem appeared after we switched env-ubuntu:latest to 24.04. It uses more recent clang-format that disagrees with previous versions about formatting of one line. Unfortunately clang-format is not yet fully stabilized and it happens from time to time. For now, I just excluded the problematic file from auto-formatting: 9b159e8b76a18690e98af25f706a62034798a899

2. Pulseaudio build. Will take a look.

gavv:
LGTM!

gavv:
I've fixed pulseaudio build: e252b2bba4078492b6c15e22921635bd8d6931c9

It was caused by update to clang 17.

Issue #670, opened 2024-01-04
Implement Array front() and back()


gavv:
[core::List](https://github.com/roc-streaming/roc-toolkit/blob/develop/src%2Finternal_modules%2Froc_core%2Flist.h), [core::Hashmap](https://github.com/roc-streaming/roc-toolkit/blob/develop/src%2Finternal_modules%2Froc_core%2Fhashmap.h) and other classes in roc_core have front() and back() methods.

It would be nice to add front() and back() to [core::Array](https://github.com/roc-streaming/roc-toolkit/blob/develop/src%2Finternal_modules%2Froc_core%2Farray.h) too.

Unlike List, Array holds values, not pointers, so these methods should return references, not pointers. They should panic if array is empty.

New methods should be covered with tests.

jlofti:
Hi I'd like to take this. Have working experience with C++, but want to contribute to open source. Thanks!

khansamad47:
@gavv would like to work on this one if not already assigned.

gavv:
Sure, thanks!

gavv:
Landed

PR #687, opened 2024-02-02
Adding accessors front and back to core::Array

khansamad47:
Addressing #670 
* Adding accessors front and back to core::Array
* Adding relevant tests


gavv:
Thanks!

PR #690, opened 2024-02-06
Issue 682: add ```List::insert_after```

### LINKED ISSUE #682: Implement List::insert_after()
[core::List](https://github.com/roc-streaming/roc-toolkit/blob/develop/src%2Finternal_modules%2Froc_core%2Flist.h) has insert_before() method.

It would be convenient to implement complementary insert_after().

New method should be covered with tests.

veronikaro:
gh-682

Closes #682.
Adds ```List::insert_after``` to the doubly-linked list with accompanying tests. 
Two existing tests for ```List::insert_before``` were renamed for differentiating them better.

### DIFF
--- a/src/internal_modules/roc_core/list.h
+++ b/src/internal_modules/roc_core/list.h
@@ -149,7 +149,7 @@ public:
         if (size_ == 0) {
             insert_(element, NULL);
         } else {
-            insert_(element, container_of_(head_.next));
+            insert_(element, head_.next);
         }
     }
 
@@ -205,7 +205,26 @@ public:
     //!  @p element should not be member of any list.
     //!  @p before should be member of this list or NULL.
     void insert_before(T& element, T& before) {
-        insert_(element, &before);
+        insert_(element, before.list_node_data());
+    }
+
+    //! Insert element into list.
+    //!
+    //! @remarks
+    //!  - inserts @p element after @p after
+    //!  - acquires ownership of @p element
+    //!
+    //! @pre
+    //!  @p element should not be member of any list.
+    //!  @p after should be member of this list.
+    void insert_after(T& element, T& after) {
+        ListNode::ListNodeData* data_after = after.list_node_data();
+
+        if (data_after->next == &head_) {
+            insert_(element, NULL);
+        } else {
+            insert_(element, data_after->next);
+        }
     }
 
     //! Remove element from list.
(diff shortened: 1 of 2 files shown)


veronikaro:
@gavv I would appreciate a review, thanks!

src/internal_modules/roc_core/list.h
```diff
@@ -208,6 +208,19 @@ class List : public NonCopyable<> {
...
+    //! @remarks
+    //!  - inserts @p element after @p after
+    //!  - acquires ownership of @p element
+    //!
+    //! @pre
+    //!  @p element should not be member of any list.
+    //!  @p after should be member of this list.
+    void insert_after(T& element, T& after) {
+        insert_(element, nextof(after));
+    }
```
gavv, line 222: We can't use nextof() here. nextof() returns Pointer; if OwnershipPolicy is RefCountedOwnership, then Pointer will resolve to `SharedPtr<T>` instead of `T*`, and this code won't compile.

We could use `&*nextof(after)`, but this would give unnecessary overhead for creating/destroying SharedPtr. I guess instead it would be better to obtain that pointer manually.

Actually, I think it makes sense to change `insert_()` second argument from `T* before` to `ListNode::ListNodeData* before`. For `insert_before()` and `push_back()` there is no difference, but for `insert_after()` we'll save unnecessary conversion: from `ListNodeData*` to `T*` and then back to `ListNodeData*`. Instead we can pass `ListNodeData*` immediately.

src/tests/roc_core/test_list_operations.cpp
```diff
@@ -193,6 +193,32 @@ TEST(list_operations, insert_middle) {
     POINTERS_EQUAL(&objects[2], list.nextof(*list.front()));
 }
 
+TEST(list_operations, insert_after_back) {
+    list.push_back(objects[0]);
+    list.insert_after(objects[1], *list.front());
```
gavv, line 198: nitpick:

```suggestion
    list.insert_after(objects[1], *list.back());
```

gavv (commented):
Thanks for the patch! Here is a couple of comments.

src/tests/roc_core/test_list_operations.cpp
veronikaro: Haha, nice catch, thank you!

src/internal_modules/roc_core/list.h
veronikaro: Thanks for the clarification! This indeed makes sense. 

I adjusted the code according to your last suggestion.
Contrary to what you write (at least this is how I understood you), changes must be made to the all corresponding insert functions.
The changes are rather minor, though.

Please let me know what you think!

gavv: Yes, sorry, I mean that their *behavior* is not affected. However there are syntactic changes, yes.

gavv:
Thanks for PR!

I've pushed small follow-up refactoring: eaf35524e40fa6b07fac03bc7414352c833e3911

(see commit message for details)

veronikaro:
> Thanks for PR!
> 
> I've pushed small follow-up refactoring: [eaf3552](https://github.com/roc-streaming/roc-toolkit/commit/eaf35524e40fa6b07fac03bc7414352c833e3911)
> 
> (see commit message for details)

Thank you for a thorough and timely review!

Issue #691, opened 2024-02-08
No-op implementation of IArena


gavv:
[core::IArena](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/iarena.h) is an interface that abstracts heap-like allocator. It is typically implemented by [core::HeapArena](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/heap_arena.h).

We have two classes that can be constructed either _with_ or _without_ IArena: `core::Array` and `core::Hashmap`. When constructed with arena, they use it to grow dynamically, otherwise they can only store data that fits within embedded capacity. The latter is used when we need container but want to be sure that it never allocates.

It would be nice to make this second feature a bit more explicit, and instead of a constructor overload without IArena, use a special arena that does not allocate.

Steps:

* Add new header `roc_core/noop_arena.h` with something like this:

    ```cpp
    class NoopArenaImpl : public IArena {
    ...
    };

    static NoopArenaImpl NoopArena;
    ```

* In Array and Hashmap, remove constructor overload without arena. Also remove internal logic that supports arena to be NULL.

* In places where that constructor was used, instead use NoopArena like this:

    ```cpp
    : array_(core::NoopArena)
    ```

* There are several tests that have NoMemArena. Search for those tests, remove NoMemArena, and instead reuse core::NoopArena.

nolan-veed:
I took the liberty of doing this.

gavv:
Thanks!

PR #698, opened 2024-02-11
IArena's NoopArena implementation

nolan-veed:
# Why

For #691

# What

* Created NoopArenaImpl class and instantiated NoopArena.
* Used it in Array, Hashmap, dropping their constructors that supported no arena.
* Updated other places.

# Testing

Existing tests should suffice.


gavv:
Thanks for PR!

1. I've merged #652, so now NoopArena needs to implement new interface.
2. In Array in Hashmap, we can now use `IArena&` instead of `IArena*`.

gavv:
CI failure is unrelated, you can ignore it.

gavv:
LGTM!

Small follow-up: ec82adab1ca2ea32ef904a82ff994c292219cf0c

PR #700, opened 2024-02-13
Refactor arena & pool guards and logs

gavv:
- Instead of more generic "flags" enum, use more specific "guards", so that it's easier to understand the purpose

- Use separate flag for each type of guard

- Extract method that increments failure counter and checks if panics are enabled for a guard

- Use same logging format as in MemoryLimiter

### DIFF
--- a/src/internal_modules/roc_core/slab_pool.h
+++ b/src/internal_modules/roc_core/slab_pool.h
@@ -23,14 +23,21 @@
 namespace roc {
 namespace core {
 
-//! Memory pool flags.
-enum SlabPoolFlags {
-    //! Enable guards for buffer overflow, invalid ownership, etc.
-    SlabPoolFlag_EnableGuards = (1 << 0),
+//! Memory pool guards.
+enum SlabPoolGuard {
+    //! Panic if leaks detected in pool destructor.
+    SlabPool_LeakGuard = (1 << 0),
+    //! Panic if detected buffer overflow when deallocating object.
+    SlabPool_OverflowGuard = (1 << 1),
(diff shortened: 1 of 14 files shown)


src/internal_modules/roc_core/slab_pool.h
```diff
@@ -23,14 +23,21 @@
...
+    //! Panic if detected ownership mismatch when deallocating object.
+    SlabPool_OwnershipGuard = (1 << 2),
 };
 
-//! Default memory pool flags.
-enum { DefaultSlabPoolFlags = (SlabPoolFlag_EnableGuards) };
+//! Default memory pool guards.
+enum {
+    SlabPool_DefaultGuards =
+        (SlabPool_LeakGuard | SlabPool_OverflowGuard | SlabPool_OwnershipGuard)
```
nolan-veed, line 39: With HeapArena, we don't want leak guard being the default, but, over here, we're ok with that?

nolan-veed (approved):
Nice one!

src/internal_modules/roc_core/slab_pool.h
gavv: Yes,

We would enable it for heap arena too, but when roc is used from a C library, some allocation & deallocation requests come from the user. We don't want to crash user's app if user decides not to free some resources upon exit, for example. So leak detection is disabled by default and is enabled explicitly in tools and tests.

All requests to pool, on the other hand, are controlled only by us, so we can place stricter defaults.

gavv:
Thanks for review!

Issue #598, opened 2023-10-06
Extract ListImpl class from List class


gavv:
[`core::List<T>`](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/list.h) template implements intrusive doubly-linked list.

To reduce code size and compilation times, it would be nice to extract its internals to a non-template implementation class. All existing code will continue using `List<T>`, however its methods will mostly just invoke similar methods of `ListImpl` and make some type casts. This way, the template part will be small, and non-template part will be compiled once and reused.

We did a very similar job for core::MpscQueue here:

* #580
* #593

rjwignar:
I might need a couple days to study the changes in the linked issue and PR but I think I can do this.
I would like to try it out.

gavv:
You're welcome, thanks!

rjwignar:
Apologies for the delay. I've been preoccupied with school and other commitments.
Wanted to give an update on my progress.

So far I've only been able to extract the `head_` and `size_` data members (along with the `List()` no-argument constructor) to the `list_impl` module.

I'll need some more time (and perhaps some guidance) to extract the remaining methods : 
- `container_of_`,
- `check_is_member_`, and
- `insert_`, 

but I'd like to keep working on this.

gavv:
@rjwignar Hi, do you still have plans on this?

rjwignar:
@gavv Apologies for the radio silence.
I couldn't figure out how to extract the remaining methods, even after referring to #593 . Please unassign me.

I am however, interested in seeing how this would be done.

veronikaro:
@gavv is this issue still relevant? If so, I would like to work on it if you do not mind since I am already got familiar with the class itself. Thanks!

gavv:
@veronikaro Absolutely, thanks!

rjwignar:
@veronikaro Thank you for taking this on!
I'm excited to learn how you refactor this.

gavv:
Merged via bef7210761ad2ff063da7ce55e7a974a3dc9c6fd

PR #702, opened 2024-02-19
Issue 598: Refactoring List class

veronikaro:
Solves #598.

### DIFF
--- a/src/internal_modules/roc_core/list.h
+++ b/src/internal_modules/roc_core/list.h
@@ -209,8 +209,7 @@ public:
     }
 
 private:
-    
-    //! Insert element preceding list node data into list.
+    //! Insert element preceding element with given list node data into list.
     //!
     //! @remarks
     //!  - inserts @p element before @p data_before
--- a/src/internal_modules/roc_core/list.h
+++ b/src/internal_modules/roc_core/list.h
@@ -44,7 +44,7 @@ public:
     ~List() {
         ListNode::ListNodeData* next_data;
 
-        for (ListNode::ListNodeData* data = impl_.head_.next; data != &impl_.head_;
+        for (ListNode::ListNodeData* data = impl_.head.next; data != &impl_.head;
              data = next_data) {
             roc_panic_if(data == NULL);
             ListImpl::check_is_member(data, &impl_);
@@ -55,7 +55,7 @@ public:
             OwnershipPolicy<T>::release(*static_cast<T*>(impl_.container_of(data)));
         }
 
-        impl_.head_.list = NULL;
+        impl_.head.list = NULL;
     }
 
     //! Get number of elements in list.
@@ -123,7 +123,7 @@ public:
     //! @pre
     //!  @p element should not be member of any list.
     void push_front(T& element) {
-        insert_(element, impl_.head_.next);
+        insert_(element, impl_.head.next);
     }
 
     //! Append element to list.
@@ -135,7 +135,7 @@ public:
     //! @pre
     //!  @p element should not be member of any list.
     void push_back(T& element) {
-        insert_(element, &impl_.head_);
+        insert_(element, &impl_.head);
     }
 
     //! Pop first element from list.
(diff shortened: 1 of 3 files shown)


veronikaro:
@gavv I would appreciate your feedback to this PR. Thanks!

gavv:
Hi, thanks for PR!

A few comments:

1. Please fix CI. There is a crash in tests reported by several jobs. (You can reproduce it locally by using --sanitizers option).

1. Some ListImpl methods use ListNodeData, and some use ListNode. For consistency, I suggest to use ListNodeData everywhere.

1. I think we could keep container_of_() in template class, after all ListImpl methods will switch to ListNodeData.

1. ListImpl::is_empty() can be removed. We can implement List::is_empty() using ListImpl.size().

1. We can simplify `~List()` to something like (see below why):

    ```cpp
    while (!is_empty()) {
        pop_front();
    }
    ```

1. Also we can move `impl_.head.list = NULL;` from `~List()` to `~ListImpl()`. Since ListImpl is responsible to initialize head, it's logical that it's also responsible to deinitialize it.

1. After these changes we can make check_is_member() a private method of ListImpl.

1. Since ListImpl is a class with methods, not a C-like struct, for consistency I suggest to make head field private and add a public getter `ListNodeData& head()`.

veronikaro:
Hi @gavv, 
sorry for the delay. 
I addressed all your comments, thank you very much for the feedback.

>     8. Since ListImpl is a class with methods, not a C-like struct, for consistency I suggest to make head field private and add a public getter `ListNodeData& head()`.

I did this at the cost of implementing `push_back` as a function of `ListImpl` level. The reason for this is that `push_back` requires a pointer to the head. 

I would appreciate your review!

src/internal_modules/roc_core/list_impl.h
```diff
@@ -0,0 +1,102 @@
...
+    void push_back(ListNode::ListNodeData* data_new);
+
+    //! Remove list node data from list.
+    //!
+    //! @pre
+    //!  @p data must be registered in this list.
+    void remove(ListNode::ListNodeData* data);
+
+    //! Get head of list
+    ListNode::ListNodeData head() const;
```
gavv, line 89: I guess we should return a reference or a pointer here. It's expected that you can take address of any node in list.

src/internal_modules/roc_core/list_impl.h
```diff
@@ -0,0 +1,102 @@
...
+
+private:
+    size_t size_;
+    //! Head of list
+    ListNode::ListNodeData _head;
```
gavv, line 94: nit: `_head` => `head_`

src/internal_modules/roc_core/list_impl.h
```diff
@@ -0,0 +1,102 @@
...
+    //! Check if list node data is registered in this list.
+    static void check_is_member(const ListNode::ListNodeData* data, const ListImpl* list);
```
gavv, line 96: nit: `check_is_member` => `check_is_member_` (because it's private)

src/internal_modules/roc_core/list.h
```diff
@@ -158,7 +125,9 @@ class List : public NonCopyable<> {
     //! @pre
     //!  @p element should not be member of any list.
     void push_back(T& element) {
-        insert_(element, &head_);
+        impl_.push_back(element.list_node_data());
+
+        OwnershipPolicy<T>::acquire(element);
```
gavv, line 130:
```
impl_.insert(element.list_node_data(), &impl_.head());
```

this should work (if you'll return a reference from head()) - and impl.push_back won't be needed.

gavv (commented):
Thanks for update! Some final comments.

gavv:
@veronikaro Hi, my recent commit (2510edd81dbd8003d1ea0a74cecefcc0622c9534) introduced lots of conflict with this patch, so I ported your code to the new version and finished remaining issues: bef7210761ad2ff063da7ce55e7a974a3dc9c6fd

Thanks!

PR #708, opened 2024-03-24
GNU/Hurd: Fix build

sthibaul:


sthibaul:
Oh, development does not happen on the master branch, how odd... Reworking this.

PR #709, opened 2024-03-24
Fix GNU/Hurd build

sthibaul:


gavv:
Thanks! It's always cool to see support for more platforms.

I've pushed two follow-up commits:

* bab2d0d1e5c59ed095d3efec972227f1caae199f - cleanup handling of GNU(-compatible) toolchains
* 1e3ecdfa8cb601c1b63f2c164452e1af94dd610e - remove PATH_MAX (I don't want to pollute header and it's not really needed)

Please let me know if my update still works on Hurd, since I don't have one to test.

PR #710, opened 2024-03-25
Splitting main() functionality

nolan-veed:
# Why

For https://github.com/roc-streaming/roc-toolkit/issues/656

# What

* Created Configs to hold all the configs.
* Wrote functions to setup various configs.
* Wrote functions to setup input and sender.

# Testing

WIP

### DIFF
--- a/src/tools/roc_send/main.cpp
+++ b/src/tools/roc_send/main.cpp
@@ -29,24 +29,17 @@
 
 using namespace roc;
 
-int main(int argc, char** argv) {
-    core::HeapArena::set_flags(core::DefaultHeapArenaFlags
-                               | core::HeapArenaFlag_EnableLeakDetection);
-
-    core::HeapArena heap_arena;
-
-    core::CrashHandler crash_handler;
-
-    gengetopt_args_info args;
-
-    const int code = cmdline_parser(argc, argv, &args);
-    if (code != 0) {
-        return code;
+struct Configs {
+    Configs(core::IArena& arena)
+        : arena(arena) {
     }
+    core::IArena& arena;
+    pipeline::SenderSinkConfig sender_config;
+    sndio::Config io_config;
+    node::ContextConfig context_config;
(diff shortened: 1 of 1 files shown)


nolan-veed:
@gavv Can I get some thoughts on approach?

gavv:
Thanks, will take a look in upcoming days!

I think the conflict is from [this commit](https://github.com/roc-streaming/roc-toolkit/commit/ebd0c7a2a468069f0df73364ff6dff3eb1c5c056) from [here](https://github.com/roc-streaming/roc-toolkit/pull/660).

gavv:
Also: d873794b234b49a32a4e19ad2be37354957d90bd

gavv:
Hi, sorry for delay. I like the approach!

A few minor comments / thoughts:

* Instead of a big `Configs` struct with all the things it maybe would be better to pass every argument explicitly.

    This way it would be clear what variables are used by what functions, and so easier to understand the flow.

* All functions are named `setup_xxx()`, but actually some of them do different things.

    Some build configs, some open devices, some configure endpoints. I suggest to reflect that in their names, again, it should make it easier to understand the flow in main().

* It would be nice if we could create all top-level objects in main().

    It seems currently the only exception is `input_source`, which we create in separate function. We could keep endpoint parsing in a function, but move construction of input source to main().

* The piece where we propagate parameters from `io_config` to `sender_config` can be extracted into a small function too.

    This way this step will be reflected in main() more explicitly and main() will be less cluttered.

gavv:
If you'll decide to proceed with this PR, I think it's good idea to review & merge just one tool (e.g. roc_send) and then work on the rest, because we can find more pitfalls during review.

gavv:
Hi, I've addressed this issue in 96ac193820c707a81451ac06e9e8afcbc55bbcec (a big commit that also includes this refactoring).

Thanks for inspiration!

PR #716, opened 2024-04-26
Assign names to threads

dmklepp:
Added a thread name char * argument to Thread constructor and applied changes to inherited classes. Added assign_name_() method for different architectures. Also added print_name() method for debugging to keep name_ member private. 

Issue link: https://github.com/roc-streaming/roc-toolkit/issues/618

### DIFF
--- a/src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
+++ b/src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
@@ -61,9 +61,39 @@ bool Thread::enable_realtime() {
     return true;
 }
 
-Thread::Thread()
+// Thread::Thread()
+//     : started_(0)
+//     , joinable_(0)
+//     , name_(nullptr)
+//     {
+// }
+
+Thread::Thread(const char* name_)
     : started_(0)
-    , joinable_(0) {
+    , joinable_(0)
+    , name_(name_) {
+}
+
+bool Thread::assign_thread_name_() {
+    if (name_) {
+#if defined(__FreeBSD__)
+        pthread_set_name_np(pthread_self(), name_);
+        return true;
+#elif defined(__NetBSD__)
+        pthread_setname_np(pthread_self_lwp(), "%s", name_);
+        return true;
+#elif defined(__APPLE__)
+        pthread_setname_np(name_);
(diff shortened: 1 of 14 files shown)


gavv:
Hi, thanks for the patch! Could you please rebase on fresh develop? There are conflicts.

src/tests/public_api/test_helpers/receiver.h
```diff
@@ -36,7 +36,8 @@ class Receiver : public core::Thread {
              size_t num_chans,
              size_t frame_size,
              unsigned flags)
-        : recv_(NULL)
+        : Thread("roc_receiver")
```
gavv, line 39: Here in below: to avoid inventing names for each test, since these names in tests are not important anyway, maybe let's just use same name like "roc_test", "test_thread", or something like this?

I don't want to make name optional, so we have compiler error if non-test code forgets to assign a name. But on the other hand maintaining those names in tests feels unnecessary.

src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
```diff
@@ -61,9 +61,39 @@ bool Thread::enable_realtime() {
...
 
-Thread::Thread()
+// Thread::Thread()
+//     : started_(0)
+//     , joinable_(0)
+//     , name_(nullptr)
+//     {
+// }
+
+Thread::Thread(const char* name_)
```
gavv, line 71: `name_` => `name`

(trailing underscore is for private members)

src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
```diff
@@ -61,9 +61,39 @@ bool Thread::enable_realtime() {
...
     : started_(0)
-    , joinable_(0) {
+    , joinable_(0)
+    , name_(name_) {
+}
```
gavv, line 75: Let's add a panic triggered if name is NULL.

gavv (commented):
Here are a few comments on the code itself.

gavv:
See https://github.com/roc-streaming/roc-toolkit/issues/618#issuecomment-2225851162

Issue #688, opened 2024-02-03
Dynamic latency adjustment


gavv:
### Implementation

- [x] in LinkMeter, compute jitter, total_packets, lost_packets
- [x] teach fec::Reader and fec::Writer to compute FEC block duration
- [x] teach LatencyMonitor to compute target latency continuously based on min latency, max latency, FEC block duration, and jitter, and pass updated target to FreqEstimator
- [x] update pipeline to support dynamic latency mode, enabled when target latency is zero (currently some components may be not ready for this)
- [x] make this mode default in CLI and API (in CLI, enable it via --target-latency=auto, in API via target_latency=0)
- [x] add new metrics to API: jitter, total_packets, lost_packets (see #681)

### Fixes

- [x] RTCP crash
- [x] seqnum crash
- [x] incorrect calculated FEC block duration
- [x] rework defaults deduction (min/max/target) 
- [x] min/max/tolerance:
  - without auto-tune:
     - min/max not used
     - session is terminated if `latency < target - tolerance` or `latency > target + tolerance`
  - with auto-tune:
     - target floats between min/max
     - session is terminated if `latency < min - tolerance` or `latency > max + tolerance`
- [x] use CsvDumper
- [x] use LogNote
- [x] test/fix sender-side adaptive latency

### Tests

- [x] LinkMeter tests
- [x] pipeline tests: `test_receiver_source`

    https://github.com/roc-streaming/roc-toolkit/blob/6eb0a5933c2419f5325862f845509b7150cf2fa0/src/tests/roc_pipeline/test_receiver_source.cpp#L2296-L2305

    - [x] `metrics_packet_counters` (simulate losses and reorders and check total_packets, lost_packets, ext_first_seqnum, ext_last_seqnum)
    - `metrics_jitter` (simulate timestamp jitter and check jitter metric) - moved to #765

- [x] pipeline tests: `test_loopback_sink_2_source`: cover new metrics (jitter, total_packets, lost_packets - smoke test that all 3 metrics are available on both sender and receiver)

   https://github.com/roc-streaming/roc-toolkit/blob/6eb0a5933c2419f5325862f845509b7150cf2fa0/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp#L350-L353

   https://github.com/roc-streaming/roc-toolkit/blob/6eb0a5933c2419f5325862f845509b7150cf2fa0/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp#L381-L384

[...]

PR #717, opened 2024-04-28
Issue 688: FEC estimates block duration

baranovmv:
Reader and writer esimate duration of
FEC block in ns as this value could vary
even though number of packets is known.
For now it is decided to keep the greates
value seen in the current session, not
average.

#688

### DIFF
--- a/src/internal_modules/roc_audio/latency_monitor.cpp
+++ b/src/internal_modules/roc_audio/latency_monitor.cpp
@@ -21,6 +21,7 @@ LatencyMonitor::LatencyMonitor(IFrameReader& frame_reader,
                                const packet::SortedQueue& incoming_queue,
                                const Depacketizer& depacketizer,
                                const packet::ILinkMeter& link_meter,
+                               const fec::Reader* fec_reader,
                                ResamplerReader* resampler,
                                const LatencyConfig& config,
                                const SampleSpec& packet_sample_spec,
@@ -30,6 +31,7 @@ LatencyMonitor::LatencyMonitor(IFrameReader& frame_reader,
     , incoming_queue_(incoming_queue)
     , depacketizer_(depacketizer)
     , link_meter_(link_meter)
+    , fec_reader_(fec_reader)
     , resampler_(resampler)
     , enable_scaling_(config.tuner_profile != audio::LatencyTunerProfile_Intact)
     , capture_ts_(0)
@@ -71,7 +73,7 @@ bool LatencyMonitor::read(Frame& frame) {
 
     if (alive_) {
         compute_niq_latency_();
-        query_link_meter_();
+        query_metrics_();
 
         if (!pre_process_(frame)) {
             alive_ = false;
@@ -103,7 +105,6 @@ bool LatencyMonitor::reclock(const core::nanoseconds_t playback_timestamp) {
 
 bool LatencyMonitor::pre_process_(const Frame& frame) {
     tuner_.write_metrics(latency_metrics_, link_metrics_);
-
     if (!tuner_.update_stream()) {
         // TODO(gh-183): forward status code
         return false;
@@ -177,12 +178,19 @@ void LatencyMonitor::compute_e2e_latency_(const core::nanoseconds_t playback_tim
     latency_metrics_.e2e_latency = playback_timestamp - capture_ts_;
 }
 
-void LatencyMonitor::query_link_meter_() {
+void LatencyMonitor::query_metrics_() {
     if (!link_meter_.has_metrics()) {
         return;
     }
 
     link_metrics_ = link_meter_.metrics();
+
+    if (fec_reader_) {
+        latency_metrics_.fec_block_duration =
+            packet_sample_spec_.stream_timestamp_2_ns(fec_reader_->max_block_duration());
+    } else {
+        latency_metrics_.fec_block_duration = 0;
+    }
 }
 
 bool LatencyMonitor::init_scaling_() {
--- a/src/internal_modules/roc_fec/reader.cpp
+++ b/src/internal_modules/roc_fec/reader.cpp
@@ -17,13 +17,13 @@ namespace fec {
 
 Reader::Reader(const ReaderConfig& config,
                packet::FecScheme fec_scheme,
-               IBlockDecoder& decoder,
+               IBlockDecoder& block_decoder,
                packet::IReader& source_reader,
                packet::IReader& repair_reader,
                packet::IParser& parser,
                packet::PacketFactory& packet_factory,
                core::IArena& arena)
-    : decoder_(decoder)
+    : block_decoder_(block_decoder)
     , source_reader_(source_reader)
     , repair_reader_(repair_reader)
     , parser_(parser)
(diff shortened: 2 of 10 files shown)


src/internal_modules/roc_audio/latency_monitor.cpp
```diff
@@ -102,6 +104,12 @@ bool LatencyMonitor::reclock(const core::nanoseconds_t playback_timestamp) {
 }
 
 bool LatencyMonitor::pre_process_(const Frame& frame) {
+    if (fec_reader_) {
+        latency_metrics_.fec_block_duration =
+            packet_sample_spec_.stream_timestamp_2_ns(fec_reader_->max_block_duration());
+    } else {
+        latency_metrics_.fec_block_duration = 0;
+    }
```
gavv, line 112: In LatencyMonitor::read, we have:

```
        compute_niq_latency_();
        query_link_meter_();

        if (!pre_process_(frame)) {
            alive_ = false;
        }
```

All metrics except fec_block_duration are filled before calling pre-process. So for consistency, I suggest to add a method like query_fec_reader() and also call it there.

src/internal_modules/roc_fec/reader.cpp
```diff
@@ -139,6 +141,13 @@ status::StatusCode Reader::get_next_packet_(packet::PacketPtr& ptr) {
     fill_block_();
 
     packet::PacketPtr pp = source_block_[next_packet_];
+    if (next_packet_ == 0) {
+        if (pp) {
+            update_block_duration_(pp);
+        } else {
+            prev_block_timestamp_valid_ = false;
+        }
+    }
```
gavv, line 150: So this is triggered when the very first packet of a block is received.

What about moving this logic to `next_block_()`?

It is triggered when we're going to switch to the next block.

Inside `next_block_()`, even if the first packet was lost, there is the maximum chance that it was repaired.

src/internal_modules/roc_fec/reader.cpp
```diff
@@ -789,5 +804,24 @@ void Reader::drop_repair_packets_from_prev_blocks_() {
     }
 }
 
+void Reader::update_block_duration_(const packet::PacketPtr& ptr) {
+    if (!ptr->rtp()) {
+        return;
+    }
```
gavv, line 810: Let's use `Packet::stream_timestamp()` instead of `Packet::rtp()`, this way fec::Reader will continue to be RTP-independent.

src/internal_modules/roc_fec/reader.cpp
```diff
@@ -789,5 +804,24 @@ void Reader::drop_repair_packets_from_prev_blocks_() {
...
+    packet::stream_timestamp_diff_t block_dur = 0;
+    if (prev_block_timestamp_valid_) {
+        block_dur = packet::stream_timestamp_diff(ptr->rtp()->stream_timestamp,
+                                                  prev_block_timestamp_);
+    }
+    roc_panic_if_msg(block_dur < 0, "fec reader: negative block duration");
```
gavv, line 816: I think panic is not appropriate here, because corrupted/malformed packets would lead to a crash?

Instead we can skip the check. Probably also log to LogTrace.

src/internal_modules/roc_fec/reader.cpp
```diff
@@ -789,5 +804,24 @@ void Reader::drop_repair_packets_from_prev_blocks_() {
...
+    block_max_duration_ = std::max(block_max_duration_, block_dur);
```
gavv, line 817: Shouldn't we compute maximum in LatencyTuner instead of fec::Reader? And report only immediate value from fec::Reader?

Why:

* Most likely we want running maximum instead of total maxium. Maybe at least in later versions. But this kind of details (running window, etc) should belong to LatencyTuner?

* This logic (running max) would probably be the same for sender & receiver.

src/internal_modules/roc_fec/writer.cpp
```diff
@@ -109,6 +114,7 @@ status::StatusCode Writer::write(const packet::PacketPtr& pp) {
     }
 
     if (cur_packet_ == 0) {
+        update_block_duration_(pp);
```
gavv, line 117: I suggest to call this from begin_block_ or end_block_ - so that everything called on block boundary would be in same place.

src/internal_modules/roc_fec/writer.cpp
```diff
@@ -342,5 +348,27 @@ bool Writer::validate_source_packet_(const packet::PacketPtr& pp) {
     return true;
 }
 
+void Writer::update_block_duration_(const packet::PacketPtr& ptr) {
+    if (!ptr->rtp()) {
+        return;
+    }
```
gavv, line 354: Same comment about `rtp()`.

src/tests/roc_fec/test_writer_reader.cpp
gavv, line 1: Can we create a separate set of tests for block_duration feature?

This feature feels pretty isolated from the main "purpose" of fec::Reader (repairing packets) and I think the story of its test is a bit different from story of the rest tests.

The story of original tests goes from simpler to more complicated, starting from one block, then introducing a loss, then switching to multiple blocks, introducing more losses, reordering, etc.

The story of block_duration tests I think should include:

* good case
* lost first packet of first block
* lost first packet of non-first block
* lost packet inside block (maybe even everything except first packet)
* lost whole block (in other words, switched to other block)
* resizing block (look at other `resize_xxx` tests for an example)

Sorry for complicating your life, but fec reader & writer are so complicated that I think having so many tests for every feature is worth it.

src/internal_modules/roc_fec/reader.cpp
baranovmv: With the manner the metrics are passed into tuner, it would be difficult to be sure that it has updates of fec block duration on every new block arrived. And this could potentially lead to  wrong estimation of maximum.

gavv: Fair enough.

gavv: Hm, given it one more thought, this is probably not the case.

While latency tuner updates FE with some big period, like 200ms, latency monitor and feedback monitor are invoked *every frame*.

And, while user is allowed to read and write big frames, top-level pipeline code will anyway split them in smaller frames (to ensure that we don't block in processing for long time and that frames can fit max size of pool buffer).

So in practice all pipeline components are invoked with small frames (e.g. 1ms), much lower than FEC block size, and I think in our design (in general) we can assume that it's always true, because it simplifies many things.

For example, we already do the same with incoming queue - we're sampling it's size every frame. I think we in general can sample anything per frame.

So I guess you can just sample FEC block duration in LatencyMonitor::read?

baranovmv: It is not an easy task to reproduce this reasoning based on source code only, so my concern could potentially be raised during some debug session and it would be difficult to confirm or deny, so I'd rather keep it as it is

gavv: Makes sense.

gavv:
Thanks!!

Follow-up: 5fedaf4b14ecfba1011ea44ca9c14b53c8993467

PR #719, opened 2024-04-28
Issue 688: UDP packet has timestamp field

baranovmv:
UDP::enqueue_ts is filled in
ReceiverSession::route_packet() right before
the packet gets into jitter buffer queue.

#688

### DIFF
--- a/src/internal_modules/roc_packet/target_libuv/roc_packet/udp.h
+++ b/src/internal_modules/roc_packet/target_libuv/roc_packet/udp.h
@@ -33,6 +33,13 @@ struct UDP {
     //! Packet receive timestamp (RTS), nanoseconds since Unix epoch.
     core::nanoseconds_t receive_timestamp;
 
+    //! Timestamp in ns since unix-epoch. It points to a moment when
+    //! the packet was transferred to a sink-thread, that "consumes"
+    //! this packet. The reason to have it separate is that this
+    //! allows us to account additional jitter introduced by
+    //! thread-switch time.
+    core::nanoseconds_t queue_ts;
+
     //! Sender request state.
     uv_udp_send_t request;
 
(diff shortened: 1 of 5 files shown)


src/internal_modules/roc_packet/target_libuv/roc_packet/udp.h
```diff
@@ -37,6 +37,11 @@ struct UDP {
     uv_udp_send_t request;
 
     UDP();
+
+    //! It points to a moment when the packet was transferred to a sink-thread, that
+    //! "consumes" this packet. The reason to have it separate is that this allows
+    //! us to account additional jitter introduced by thread-switch time.
+    core::nanoseconds_t enqueue_ts;
```
gavv, line 44: 1. We need to initialize it to zero in udp.cpp.

1. Please mention in comment that it's nanoseconds since unix epoch.

1. Could you place it near receive_timestamp above please?

1. What about "queue_ts" instead of "enqueue_ts"? Then we can use nice acronym QTS for it (easily distinguishable compared to e.g. "ETS").

1. Could you also document it here? https://github.com/roc-streaming/roc-toolkit/blob/develop/docs/sphinx/internals/timestamps.rst

src/internal_modules/roc_pipeline/receiver_session.cpp
```diff
@@ -246,6 +246,7 @@ audio::IFrameReader& ReceiverSession::frame_reader() {
 status::StatusCode ReceiverSession::route_packet(const packet::PacketPtr& packet) {
     roc_panic_if(!is_valid());
 
+    packet->udp()->enqueue_ts = core::timestamp(core::ClockUnix);
     return packet_router_->write(packet);
```
gavv, line 250: 1. Maybe move this into packet::Router::write? Packet router is supposed to be entry point to the packets pipeline, so it looks like a good place to be responsible for filling timestamps. On the other hand, it would be nice to keep pipeline more high-level.

2. We need a guard for the case when packet->udp() returns NULL. (I think we can just skip timestamp filling in this case).

gavv (commented):
Thanks, a few comments.

PR #721, opened 2024-05-09
Fix debian rules for host specification

sthibaul:
Debian's multiarch string can be different from the GNU type string. Only the latter should be used for finding a toolchain. For instance, on i386, the gnu type is i686-linux-gnu, while multiarch is i386-linux-gnu.


gavv:
Thanks.

PR #722, opened 2024-05-09
Fix unix-like platforms build

sthibaul:
semaphore.h is used in various places, and the implementation based on unix semaphores is largely supported by unix platforms.

Fixes:
src/internal_modules/roc_ctl/control_task.h:20:10: fatal error: roc_core/semaphore.h: No such file or directory
   20 | #include "roc_core/semaphore.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -772,7 +772,7 @@ else:
             'target_posix_pc',
         ])
 
-    if meta.platform in ['linux', 'android']:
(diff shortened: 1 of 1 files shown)


gavv:
Thanks. I forgot that we have semaphore only for posix_ext and darwin right now.

PR #723, opened 2024-05-09
Fix debian rules for libdir installation

sthibaul:
efbce5e94f99 ("Fix debian rules for host specification") fixed finding the proper toolchain, but roc-toolkit now installs libraries under the gnu type string, while Debian expects the multiarch path. So we need to explicitly set libdir to the multiarch path.


gavv:
Thanks.

Issue #683, opened 2024-01-30
Implement secure_random() function using OpenSSL


gavv:
There are a few RTP/RTCP fields that are required to be populated using cryptographically secure PRNG. Currently we fill them using regular PRNG ([core::fast_random](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/fast_random.h)), which will become a security issue when we add encryption.

We need to implement a **new function** `core::secure_random`:

```cpp
ROC_ATTR_NODISCARD bool secure_random(void* buf, size_t bufsz);
```

We already have OpenSSL support, so we can implement `secure_random()` using RAND module from OpenSSL.

Since OpenSSL can be disabled at build time via `--disable-openssl` scons option, we should actually add two implementations:

* in `roc_core/target_openssl` - main implementation, using OpenSSL
* in `roc_core/target_nocsprng` - alternative fallback implementation that uses fast_random()

Scons will automatically use target_openssl when OpenSSL is enabled, and target_nocsprng when it's disabled (this feature is already implemented in SConstruct). See [here](https://roc-streaming.org/toolkit/docs/internals/code_structure.html#targets) about target directories.

Then we should **switch the following classes/functions** from fast_random() to secure_random():

* [rtp::Identity](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_rtp/identity.h)
* [rtp::Sequencer](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_rtp/sequencer.h)
* [fec::BlockWriter](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_fec/block_writer.h)
* [core::uuid_generare()](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/uuid.h)

Each one has a few calls to fast_random(), which should be replaced with secure_random() and error checking (use StatusErrRand when status code is expected).

We also should **add simple tests** for secure_random() similar to fast_random().

gavv:
@lovvik @Yannise-A you're both welcome, thank you :)

gavv:
Quote from chat with clarifications:

> you need to create two target directories in roc_core (not in tests, but in module itself):
> * target_openssl
> * target_nocsprng
>
> in target_openssl, place secure_random.cpp with implementation that uses RAND_bytes
[...]

PR #727, opened 2024-06-05
Implement secure_random()

lovvik:
as asked by issue #683 
- Implemented `secure_random()`
- Added 2 build directories
- Changed calls of ` fast_random()` in src, tests and benchmarks
- Added `-lcrypto` linking

### DIFF
--- a/.github/workflows/pr_opened.yml
+++ b/.github/workflows/pr_opened.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Check issue
         id: issue
         run: |
-          echo url=$(scripts/pr.py show ${{ github.event.pull_request.number }} --json \
+          echo url=$(scripts/rgh.py show ${{ github.event.pull_request.number }} --json \
             | jq -r .issue.url) >> "$GITHUB_OUTPUT"
         env:
          GH_TOKEN: ${{ secrets.github_token }}
--- a/scripts/rgh.py
+++ b/scripts/rgh.py
@@ -64,7 +64,7 @@ def run_cmd(cmd, input=None, env=None, retry_fn=None):
 
 def random_worktree():
     while True:
-        path = '/tmp/prpy_' + ''.join(random.choice(string.ascii_lowercase + string.digits)
+        path = '/tmp/rgh_' + ''.join(random.choice(string.ascii_lowercase + string.digits)
             for _ in range(8))
         if not os.path.exists(path):
             return path
@@ -643,7 +643,7 @@ def stealth_rebase(base_branch):
         'GIT_SEQUENCE_EDITOR': ':',
     })
 
-parser = argparse.ArgumentParser(prog='pr.py')
+parser = argparse.ArgumentParser(prog='rgh.py')
 
 common_parser = argparse.ArgumentParser(add_help=False)
 common_parser.add_argument('--org', default='roc-streaming',


src/internal_modules/roc_core/target_nocsprng/roc_core/secure_random.cpp
```diff
@@ -81,5 +80,20 @@ double fast_random_gaussian() {
...
+};
+
+uint32_t secure_random_range(uint32_t from, uint32_t to) {
+    return fast_random_range(from, to);
+};
+
+double secure_random_gaussian() {
+    return fast_random_gaussian();
+};
+
```
adrianopol, line 97: this should be implemented in target_nocsprng (as written in #683)

src/internal_modules/roc_core/target_nocsprng/roc_core/secure_random.h
```diff
@@ -6,11 +6,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-//! @file roc_core/fast_random.h
+//! @file roc_core/target_nocsprng/roc_core/secure_random.h
```
adrianopol, line 9: see above

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
adrianopol, line 1: it was intended to implement
```
ROC_ATTR_NODISCARD bool secure_random(void* buf, size_t bufsz);
```
function signature while you have implemented another ones:
```
uint32_t secure_random();
uint32_t secure_random_range(uint32_t from, uint32_t to);
double secure_random_gaussian();
```

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "secure_random.h"
+#include "roc_core/log.h"
+#include <openssl/err.h>
+#include <openssl/rand.h>
+
+namespace roc {
+namespace core {
+
+namespace {} // namespace
```
adrianopol, line 17: useless

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "secure_random.h"
+#include "roc_core/log.h"
+#include <openssl/err.h>
+#include <openssl/rand.h>
```
adrianopol, line 12: please separate local and system headers with an empty line like this:

```suggestion
#include "secure_random.h"

#include "roc_core/log.h"

#include <openssl/err.h>
#include <openssl/rand.h>
```

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.h
```diff
@@ -0,0 +1,35 @@
...
+//! @returns 0 or random value between 0 and UINT32_MAX.
+uint32_t secure_random();
+
+//! Get a random integer from a cryptographically secure PRNG.
+//! @returns random value in inclusive range [from; to].
+uint32_t secure_random_range(uint32_t from, uint32_t to);
+
+//! Get a random double from a cryptographically secure PRNG.
+//! @returns normally distibure random value with 1 variance.
+double secure_random_gaussian();
```
adrianopol, line 30: see above

src/internal_modules/roc_packet/interleaver.cpp
```diff
@@ -96,7 +96,7 @@ void Interleaver::reinit_seq_() {
         send_seq_[i] = i;
     }
     for (size_t i = block_size_; i > 0; --i) {
-        const size_t j = core::fast_random_range(0, (unsigned int)i - 1);
+        const size_t j = core::secure_random_range(0, (unsigned int)i - 1);
```
adrianopol, line 99: this should not be changed (see task description)

src/tests/roc_core/bench_random.cpp
```diff
@@ -17,7 +17,7 @@ namespace {
 void BM_Random_Fast(benchmark::State& state) {
     uint32_t r = 0;
     while (state.KeepRunning()) {
-        r = fast_random_range(r, (uint32_t)-1);
+        r = secure_random_range(r, (uint32_t)-1);
```
adrianopol, line 20: this should not be changed

src/tests/roc_core/test_secure_random.cpp
adrianopol, line 1: the original file should be kept

src/tests/roc_ctl/bench_task_queue_contention.cpp
adrianopol, line 1: this should not be changed

src/tests/roc_fec/test_encoder_decoder.cpp
adrianopol, line 1: this should not be changed

src/tests/roc_netio/test_helpers/conn_reader.h
adrianopol, line 1: this should not be changed

src/tests/roc_netio/test_helpers/conn_writer.h
adrianopol, line 1: this should not be changed; checkout the task description and revert other places please

adrianopol (changes requested):

src/internal_modules/roc_core/target_nocsprng/roc_core/secure_random.cpp
lovvik: Hello, i don't understand. That is target_nocsprng.

src/tests/roc_core/test_secure_random.cpp
lovvik: this was asked by gevv : https://github.com/roc-streaming/roc-toolkit/issues/683#issuecomment-2148359306

src/internal_modules/roc_packet/interleaver.cpp
lovvik: So, the changes made to files not mentioned in the issue but calling fast_random derivatives were necessary.
`target_nocsprng` content is ignored in a default build (using openssl).

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
lovvik: my bad, i was focused on uniformity.
Are we sure secure_random() won't be used like fast_random() ?

3rdparty/SConscript
```diff
@@ -211,9 +211,9 @@ elif 'openssl' in system_dependencies:
     conf = Configure(env, custom_tests=env.CustomTests)
 
     if not conf.AddPkgConfigDependency('libssl', '--cflags --libs', add_prefix=ossl_prefix):
-        conf.env.AddManualDependency(libs=['ssl'], prefix=ossl_prefix)
+        conf.env.AddManualDependency(libs=['ssl','crypto'], prefix=ossl_prefix)
 
-    if not conf.CheckLibWithHeaderExt('ssl', 'openssl/rand.h', 'C', run=not is_crosscompiling):
+    if not conf.CheckLibWithHeaderExt('ssl', 'openssl/rand.h', 'C', run=not is_crosscompiling) or not conf.CheckLibWithHeaderExt('crypto', 'openssl/crypto.h', 'C', run=not is_crosscompiling):
```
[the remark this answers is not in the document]
lovvik: i don't really get why, but this  how i could get `-lcrypto` to be linked.
This results in linkage errors : 
```if not conf.AddPkgConfigDependency('libssl', '--cflags --libs', add_prefix=ossl_prefix):
        conf.env.AddManualDependency(libs=['ssl','crypto'], prefix=ossl_prefix)

    if not conf.CheckLibWithHeaderExt('ssl', 'openssl/rand.h', 'C', run=not is_crosscompiling): #or not conf.CheckLibWithHeaderExt('crypto', 'openssl/crypto.h', 'C', run=not is_crosscompiling):
        env.Die("OpenSSL not found (see 'config.log' for details)") ```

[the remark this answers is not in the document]
Comment (gavv) [reply 3rdparty/SConscript:216]: Seems we have `AddPkgConfigDependency('libssl')`, but don't have `AddPkgConfigDependency('libcrypto')`, it should be added, because they're different packages in pkg-config:

```
pkg-config --cflags --libs libssl
-lssl 

pkg-config --cflags --libs libcrypto
-lcrypto 
```

And in this case it indeed makes sense to add a second check. `openssl/rand.h` checks that libssl worked, and `openssl/crypto.h` checks that lubcrypto worked.

> i don't really get why, but this how i could get -lcrypto to be linked.

Likely CheckLibWithHeaderExt adds it, but we should no rely on it. Actually it's probably a bug that should be fixed (not in this PR of course).

Also please avoid long lines.

File: src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
Comment (gavv) [reply src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp:1]: > Are we sure secure_random() won't be used like fast_random()?

I think we're sure. I can't think of any such use case where we need ranges or gaussian distribution with CSPRNG (instead of normal PRNG), at least for now.

Also fast random signatures don't return errors (can never fail), but CSPRNG can always fail and errors should be always handled.

So I agree with Andrew, let's keep only one function here.

File: src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp (hunk 4 above)
Comment (gavv) [reply src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp:12]: Actually the first block should be:

```
#include "roc_core/secure_random.h"
#include "roc_core/log.h"
```

(we always use full path to file)

**Comment by gavv** [comment]
Thanks for PR!

Regarding my comment from matrix that you linked above: https://github.com/roc-streaming/roc-toolkit/issues/683#issuecomment-2148359306

I guess I was not very clear.

The idea is:

* fast_random is always available and has a single cross-platform implementation; it's not affected by this task at all
* secure_random is always available
* but secure_random can have two different implementations: using OpenSSL, and fallback using fast_random (which as you remember is always available)
* some parts of the code use fast_random, some use secure_random, depending on requirements; CSPRNG is needed when RTP RFC mandates it; but it's slower and can fail, so it's not used everywhere
* regarding tests:
  * we need tests for fast_random and for secure_random; they're different tests at least because these functions have different signatures
  * tests for fast_random should be be affected by this task at all
  * tests for secure_random can have only one implementation, no matter if secure_random is implemented via OpenSSL or fast_random

That's what I tried to say in that comment, we don't need two separate implementations of tests for secure_random, however we still need separate tests for fast_random and secure_random.

File: 3rdparty/SConscript (hunk 8 above)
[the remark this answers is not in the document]
Comment (lovvik) [reply 3rdparty/SConscript:216]: Hello, i now have this. The `AddPkgConfigDependency('libcrypto')` does it (even without CheckLib...) 
```
elif 'openssl' in system_dependencies:
    ossl_prefix = None
    # macOS: fallback to use a brew-installed openssl
    if meta.platform == 'darwin' and not is_crosscompiling:
        ossl_prefix = env.FindBrewPackage('openssl')

    if not conf.AddPkgConfigDependency('libssl', '--cflags --libs', add_prefix=ossl_prefix):
        conf.env.AddManualDependency(libs=['ssl'], prefix=ossl_prefix)

    if not conf.AddPkgConfigDependency('libcrypto', '--cflags --libs', add_prefix=ossl_prefix):
        conf.env.AddManualDependency(libs=['crypto'], prefix=ossl_prefix)

    if not conf.CheckLibWithHeaderExt('ssl', 'openssl/rand.h', 'C', run=not is_crosscompiling) or \
        not conf.CheckLibWithHeaderExt('crypto', 'openssl/crypto.h', 'C', run=not is_crosscompiling):
        env.Die("OpenSSL not found (see 'config.log' for details)")

    env = conf.Finish()
    ``` 
 is it ok?

PR #732, opened 2024-06-18
Support using ccache when building gengetopt and ragel

### LINKED ISSUE #639: Support using ccache when building gengetopt and ragel
User can enable ccache using `--compiler-launcher=ccache` scons option.

When user specifies `--build-3rdparty=<dependencies>`, and ccache is enabled, we automatically use ccache for almost all dependencies. However, we don't enable ccache for `gengetopt` and `ragel`. See [build-3rdparty.py](https://github.com/roc-streaming/roc-toolkit/blob/develop/scripts/scons_helpers/build-3rdparty.py) script.

The reason is that gengetopt and ragel are tools needed during build. Unlike other dependencies, they're not cross-compiled. Instead, they are built natively and then scons is using them. Because of that, `build-3rdparty.py` doesn't provide `CC` and `CXX` vars when it builds gengetopt and ragel, and so there is no place where it could inject `ccache`.

So the fix is straightforward:

* teach `build-3rdparty.py` to detect CC and CXX for gengetopt and ragel (it should check if gcc/clang/cc is available in PATH and use it)
* if ccache is enabled (i.e. COMPILER_LAUNCHER is set), add it to CC and CXX (e.g. `CC=gcc` becomes `CC="ccache gcc"`) 

See also #638.

jeshwanthreddy13:
Fixes #639

- Implemented ccache support when building gengetopt and ragel in file 'build-3rdparty.py'.
- Added 'CC' and 'CXX' paramters when executing './configure' so as to inject 'ccache' when '--compiler-launcher=ccache' is used.

Additional Notes:
- Added a new function "detect_compiler_presence" for detecting if a particular compiler is present or not. 
- Respective 'C' and 'C++' compilers are added to CC and CXX respectively by checking if the specific compilers can be executed. 
- Compiler launcher is added in 'CC' and 'CXX' only if the parameter is passed during build.

### DIFF
--- a/scripts/scons_helpers/build-3rdparty.py
+++ b/scripts/scons_helpers/build-3rdparty.py
@@ -381,9 +381,11 @@ def apply_patch(ctx, dir_path, patch_url, patch_name):
         dir_path,
         '../' + patch_name), ignore_error=True)
 
-def format_vars(ctx, disable_launcher=False):
+def format_vars(ctx, disable_launcher=False, env=None):
     ret = []
-    for k, v in ctx.env.items():
+    if not env:
+        env = ctx.env
+    for k, v in env.items():
         if k == 'COMPILER_LAUNCHER':
             continue
         elif k in ['CC', 'CXX'] and not disable_launcher:
@@ -720,6 +722,24 @@ def detect_compiler_family(env, toolchain, family):
                 return False
 
     return True
+
+def detect_compiler_presence(compiler):
+    try:
+        subprocess.run([compiler, '-v'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        return True
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return False
+
+def detect_native_cc_cxx_compilers():
+    for compiler in ["gcc", "clang", "cc"]:
+        if detect_compiler_presence(compiler):
+            CC_native_compiler = compiler
+            break
+    for compiler in ["g++", "clang++"]:
+        if detect_compiler_presence(compiler):
+            CXX_native_compiler = compiler
+            break
+    return {'CC' : CC_native_compiler, 'CXX' : CXX_native_compiler}
 
 # Guess platform argument for OpenSSL's Configure script basing on toolchain string.
 # (see `./Configure LIST` for a list of all platforms and their format)
@@ -1462,6 +1482,7 @@ elif ctx.pkg_name == 'json-c':
     install_tree(ctx, '.', ctx.pkg_inc_dir, include=['*.h'])
     install_files(ctx, '.libs/libjson-c.a', ctx.pkg_lib_dir)
 elif ctx.pkg_name == 'gengetopt':
+    native_compiler = detect_native_cc_cxx_compilers(); 
     download(
         ctx,
         'https://ftp.gnu.org/gnu/gengetopt/gengetopt-{ctx.pkg_ver}.tar.gz',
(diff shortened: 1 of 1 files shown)


jeshwanthreddy13:
> Hi!
> 
> Please take a look at the part of task about cross-compilation. We can't just use CC and CXX because they're set to cross-compiler, not native compiler.

Okay, so we need to make sure that CC and CXX are always passed with native compilers instead of the cross compilers for ragel and gengetopt?

gavv:
Yes, because they are build tools running during build.

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -1471,10 +1490,26 @@ def die(text, *args):
         'gengetopt-{ctx.pkg_ver}.tar.gz',
         'gengetopt-{ctx.pkg_ver}')
     changedir(ctx, 'src/gengetopt-{ctx.pkg_ver}')
-    execute(ctx, './configure', clear_env=True)
+    compiler_launcher = ctx.env.get('COMPILER_LAUNCHER', None)
+    execute(ctx, './configure CC="{compiler_launcher}{CC_compiler}" CXX="{compiler_launcher}{CXX_compiler}"'.format(
+    compiler_launcher='%s' % (compiler_launcher + ' ') if compiler_launcher else '',
+    CC_compiler=CC_compiler,
+    CXX_compiler=CXX_compiler),
```
gavv, line 1497: Quoting is not quite correct because is provided by user and may have any symbols (e.g. `"`).

We already have `format_vars()` that handles both quoting and COMPILER_LAUNCHER, I suggest to reuse it.

Currently it formats vars from ctx.env, but we could add an optional argument `env`, unset by default. When it's set, it would use it instead of ctx.env.

Then we could use it like this:

```
format_vars(..., env={'CC': CC_compiler, 'CXX': ..., 
   'COMPILER_LAUNCHER': ctx.env['COMPILER_LAUNCHER']})
```

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -721,6 +721,14 @@ def _check_tool(toolchain, tool):
 
     return True
 
+def detect_compiler_presence(compiler):
+
+    try:
```
gavv, line 726: nit: Please remove unnecessary empty lines.

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -1471,10 +1490,26 @@ def die(text, *args):
         'gengetopt-{ctx.pkg_ver}.tar.gz',
         'gengetopt-{ctx.pkg_ver}')
     changedir(ctx, 'src/gengetopt-{ctx.pkg_ver}')
-    execute(ctx, './configure', clear_env=True)
+    compiler_launcher = ctx.env.get('COMPILER_LAUNCHER', None)
+    execute(ctx, './configure CC="{compiler_launcher}{CC_compiler}" CXX="{compiler_launcher}{CXX_compiler}"'.format(
+    compiler_launcher='%s' % (compiler_launcher + ' ') if compiler_launcher else '',
```
gavv, line 1495: nit: Please use indentation.

gavv (commented):
Thanks for update! A few comments.

scripts/scons_helpers/build-3rdparty.py
```diff
@@ -1462,6 +1470,17 @@ def die(text, *args):
...
+
+    for compiler in ["gcc", "clang", "cc"]:
+        if detect_compiler_presence(compiler):
+            CC_compiler = compiler
+            break
+
+    for compiler in ["g++", "clang++"]:
+        if detect_compiler_presence(compiler):
+            CXX_compiler = compiler
+            break
```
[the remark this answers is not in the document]
jeshwanthreddy13: Done, Created a new function "detect_native_cc_cxx_compiler" which returns a pair and also renamed the variables accordingly.

scripts/scons_helpers/build-3rdparty.py
jeshwanthreddy13: Done, reused `format_vars` by adding a new parameter `env` which is set to `None` by default and is used instead of `ctx.env` when used.

scripts/scons_helpers/build-3rdparty.py
jeshwanthreddy13: Done

scripts/scons_helpers/build-3rdparty.py
jeshwanthreddy13: Done

gavv:
Thanks!!

Small follow-up commit: bf5e3656925678e0d0872c50224ac7bb4cf40d1e

(see commit message for details)

PR #736, opened 2024-06-22
implement safe backtrace print and exit upon SIGQUIT

MicJagger:
https://github.com/roc-streaming/roc-toolkit/issues/566


gavv:
Hi,

This patch is odd. signal_string() returns a string name of a signal, why on earth are you calling exit from it? Also you're not installing signal handler. _exit() doesn't take a signal as an argument (see man page). And all other signals are handled via die_gracefully(), not by calling exit.

Issue #726, opened 2024-05-30
Add unit tests for core::RingQueue


gavv:
Cover [RingQueue](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/ring_queue.h) with tests.

[test_array.cpp](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/tests/roc_core/test_array.cpp) can be used as an example.

mihir-mihir:
I can work on this!

gavv:
Great, thanks!

gavv:
Landed.

PR #743, opened 2024-06-26
added test_ring_queue.cpp, modified RingQueue capacity() definition

mihir-mihir:
#726 

**Changes**
1. [880ee1c](https://github.com/roc-streaming/roc-toolkit/pull/743/commits/880ee1c9dc9625df5752d9fca624fe276f899904)
- Wrote unit tests for RingQueue class
- Changed the definition of capacity() in ring_queue.h
    - Before making this change, if an element was added when `size == buff_len_ - 1`, this caused `begin_` and `end_` to be identical, which resulted in `size()` incorrectly returning 0
    - It looks like in the current RingQueue implementation, where the end pointer points to one slot ahead of the end data element, one empty slot needs to be reserved to distinguish between:
        -  the empty state (when `begin_` and `end_` are identical)
        -  the full state (when the end pointer has "caught up" to one slot behind the begin pointer in the ring)
    - This means that the actual capacity of the RingQueue would be `buff_len_ - 1` instead of `buff_len_`
    - The tests passed after changing `capacity()` to return `buff_len_ - 1` instead of `buff_len_`
2. [d3ef99f](https://github.com/roc-streaming/roc-toolkit/pull/743/commits/d3ef99f2ac8e4162de6e6a582cecca88841faecd)
- Redefined `is_full()` for RingQueue to calculate the full condition by working with the pointers directly
- Modified the RingQueue constructor definition to clarify that the actual capacity of the queue is `max_len - 1`

### DIFF
--- a/src/internal_modules/roc_core/mov_stats.h
+++ b/src/internal_modules/roc_core/mov_stats.h
@@ -40,9 +40,9 @@ public:
         , mov_max_cntr_(0)
         , full_(false)
         , first_(true)
-        , queue_max_(arena, win_len + 1)
+        , queue_max_(arena, win_len)
         , curr_max_(T(0))
-        , queue_min_(arena, win_len + 1)
+        , queue_min_(arena, win_len)
         , curr_min_(T(0))
         , valid_(false) {
         if (win_len == 0) {
--- a/src/internal_modules/roc_core/ring_queue.h
+++ b/src/internal_modules/roc_core/ring_queue.h
@@ -38,18 +38,21 @@ template <class T, size_t EmbeddedCapacity = 0> class RingQueue : public NonCopy
 public:
     //! Initialize.
     //! @remarks
-    //!  Preallocate buffer in @p arena with @p max_len number of elements.
+    //!  Preallocate buffer in @p arena for a queue with a capacity of `max_len` elements.
+    //!  In this implementation, an empty slot needs to be reserved in the buffer to be
+    //!  able to distinguish between the queue's empty and full states, so `max_len + 1`
+    //!  elements are allocated for the buffer.
     RingQueue(core::IArena& arena, size_t max_len)
         : buff_(NULL)
-        , buff_len_(max_len)
+        , buff_len_(max_len + 1)
         , begin_(0)
         , end_(0)
         , arena_(arena) {
-        if (max_len == 0) {
+        if (max_len <= 0) {
             roc_panic("ring queue: the length must be greater than 0");
         }
 
-        buff_ = allocate_(max_len);
+        buff_ = allocate_(buff_len_);
     }
 
     ~RingQueue() {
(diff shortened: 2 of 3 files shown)


gavv:
Thanks for PR and for catching the bug!

I suggest a bit different fix for the issue. We can make that extra chunk an implementation detail - if user requested N chunks, visible capacity would be exactly N chunks, although internally we can allocate N+1 chunks.

So instead of fixing capacity(), we can fix constructor I guess, and allocate +1 chunk. Also, if EmbeddedCapacity is non-zero, we likely need to add +1 to it as well.

We'll also need to fix core::MovStats and use `win_len` instead of `win_len + 1` there when initializing RingQueue.

Thanks!

@baranovmv FYI

Also would be nice to add one more test with for the boundary case when queue size is 1 element.

mihir-mihir:
Sounds good, made these changes in [da43de9](https://github.com/roc-streaming/roc-toolkit/pull/743/commits/da43de9c7cae6fe0f06a6e364920271e478ffbe8). Also, do we want EmbeddedCapacity to represent (a), the max queue capacity that we allow for using data embedding (in this case the embedded buffer would contain one more element than the value of EmbeddedCapacity), or (b), the max number of buffer elements allowed when using data embedding? I read your comment as meaning (a), but let me know if you meant (b) and I can make the change.

gavv:
> Sounds good, made these changes in [da43de9](https://github.com/roc-streaming/roc-toolkit/pull/743/commits/da43de9c7cae6fe0f06a6e364920271e478ffbe8). Also, do we want EmbeddedCapacity to represent (a), the max queue capacity that we allow for using data embedding (in this case the embedded buffer would contain one more element than the value of EmbeddedCapacity), or (b), the max number of buffer elements allowed when using data embedding? I read your comment as meaning (a), but let me know if you meant (b) and I can make the change.

Yep, I meant (a). I think EmbeddedCapacity, max_len, and capacity() all should be consistent and have same semantics, since they're visible to user of RingQueue.

gavv:
I guess we should do something like:

```
AlignedStorage<(EmbeddedCapacity != 0 ? EmbeddedCapacity+1 : 0) * sizeof(T)> embedded_data_;
```

?

mihir-mihir:
> I guess we should do something like:
> 
> ```
> AlignedStorage<(EmbeddedCapacity != 0 ? EmbeddedCapacity+1 : 0) * sizeof(T)> embedded_data_;
> ```
> 
> ?

Got it, made that change thanks. I think everything should be covered in [880cdd7](https://github.com/roc-streaming/roc-toolkit/pull/743/commits/880cdd7ecb4aaaf1e8672448ec2e92d5e41bb59b) now.

gavv:
Thank you!

Small improvement: 81b2235b8b1971885947310ce7abf052ae78bc35

PR #745, opened 2024-07-02
Implement StringList pop_back()

### LINKED ISSUE #669: Implement StringList pop_back()
[core::StringList](https://github.com/roc-streaming/roc-toolkit/blob/develop/src%2Finternal_modules%2Froc_core%2Fstring_list.h) has push_back() method.

It would be convenient to add complementary method:

```c
void pop_back();
```

New method should be covered with tests.

jeshwanthreddy13:
Fixes #669 

- Implemented a new "void pop_back()" method.
- Removes the last string from StringList.
- Returns false if deallocation fails
- Panics if StringList is empty.
- Clears when last string is poped.

Test cases covered:

- Size should be zero after removing last element from the list. 
- front_ and back_ have to point to same string when initial list with two strings is poped. 
- back_ should point to 'n-1' string after poping initial 'n' number of strings list.

### DIFF
--- a/src/internal_modules/roc_core/string_list.cpp
+++ b/src/internal_modules/roc_core/string_list.cpp
@@ -130,6 +130,32 @@ bool StringList::push_back(const char* str_begin, const char* str_end) {
     return true;
 }
 
+bool StringList::pop_back() {
+    if (size_ == 0) {
+        roc_panic("stringlist: list is empty");
+    }
+
+    const size_t blk_sz = back_->len;
+    const Footer* prev_footer = NULL;
+    if (size_ > 1) {
+        prev_footer = (const Footer*)((const char*)back_ - sizeof(Footer));
+    }
+
+    if (!data_.resize(data_.size() - blk_sz)) {
+        return false;
+    }
+
+    size_--;
+    if (size_) {
+        back_ = (Header*)(data_.data() + data_.size() - prev_footer->len);
+    } else {
+        front_ = NULL;
+        back_ = NULL;
+    }
+
+    return true;
+}
+
 const char* StringList::find(const char* str) {
     if (str == NULL) {
         roc_panic("stringlist: string is null");
--- a/src/internal_modules/roc_core/string_list.h
+++ b/src/internal_modules/roc_core/string_list.h
@@ -87,6 +87,11 @@ public:
     //!  false if allocation failed.
     ROC_ATTR_NODISCARD bool push_back(const char* str_begin, const char* str_end);
 
+    //! Remove string from end of the list.
+    //! @returns
+    //!  false if deallocation failed.
+    ROC_ATTR_NODISCARD bool pop_back();
+
     //! Find string in the list.
     //! @returns
     //!  the string in the list or NULL if it is not found.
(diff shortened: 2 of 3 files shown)


src/internal_modules/roc_core/string_list.cpp
```diff
@@ -130,6 +130,28 @@ bool StringList::push_back(const char* str_begin, const char* str_end) {
...
+    }
+
+    size_--;
+    if (size_ > 0) {
+        back_ = (Header*)(data_.data() + data_.size() - blk_sz);
+    }
+
+    else {
+        clear();
+    }
```
gavv, line 150: Cosmetic suggestion:

```suggestion
    if (size_ > 0) {
        back_ = (Header*)(data_.data() + data_.size() - blk_sz);
    } else {
        front_ = NULL;
        back_ = NULL;
    }
```

(Since in other places we manipulate front & back directly, let's do it here too for consistency, instead of calling clear()).

src/internal_modules/roc_core/string_list.h
```diff
@@ -80,6 +80,11 @@ class StringList : public NonCopyable<> {
     //!  false if allocation failed.
     ROC_ATTR_NODISCARD bool push_back(const char* str);
 
+    //! Remove string from end of the list.
+    //! @returns
+    //!  false if deallocation failed.
+    ROC_ATTR_NODISCARD bool pop_back();
```
gavv, line 86: Nitpick: please move method after push_back(), to keep two push_backs grouped together.

src/internal_modules/roc_core/string_list.cpp
```diff
@@ -130,6 +130,28 @@ bool StringList::push_back(const char* str_begin, const char* str_end) {
...
+
+    const size_t blk_sz = back_->len;
+    if (!data_.resize(data_.size() - blk_sz)) {
+        return false;
+    }
+
+    size_--;
+    if (size_ > 0) {
+        back_ = (Header*)(data_.data() + data_.size() - blk_sz);
```
gavv, line 145: Bug: here we subtract size of removed block, but instead we should get size of block before removed one and use it here.

Test passes because in test all strings have same lengths (3 chars). Please update test to use strings of different sizes.

gavv (commented):
Thanks for PR! A few comments.

src/internal_modules/roc_core/string_list.cpp
jeshwanthreddy13: Done

src/internal_modules/roc_core/string_list.h
jeshwanthreddy13: Done

src/internal_modules/roc_core/string_list.cpp
jeshwanthreddy13: Thanks for pointing it out. I had completely missed it.
Fixed in recent commit with test cases for different sized strings.

gavv:
Thank you!

gavv:
Follow-up: use different string sizes in other tests too: 91581a05e5cfe6d89466afd4336a570dffe37cfb

PR #747, opened 2024-07-10
Return status code instead of bool in IBlockEncoder & IBlockDecoder

### LINKED ISSUE #738: Return status code instead of bool in IBlockEncoder & IBlockDecoder
## Summary

[fec::IBlockEncoder](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_fec/iblock_encoder.h) and [fec::IBlockDecoder](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_fec/iblock_decoder.h) are interfaces for codec-specific encoding and decoding of FEC packets used for packet loss recovery. See [documentation](https://roc-streaming.org/toolkit/docs/internals/fec.html).

Currently they both have `begin_block()` method that returns bool, which is true on success or false on error. We need to replace bool with [status::StatusCode](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_status/status_code.h) and return code that described why the operation failed.

## Implementation

* Update IBlockEncoder and IBlockDecoder interfaces
* Update implementations:
   * fec::OpenfecEncoder
   * fec::OpenfecDecoder
* Report appropriate statuses:
   * when allocation failed, return StatusNoMem
   * on success, return StatusOK
* Update fec::BlockWriter and fec::BlockReader, which use IBlockEncoder and IBlockDecoder interfaces. When encoder or decoder return errors, writer and reader should forward it to upper level.

## Testing

* Add unit test to `test_block_encoder_decoder.cpp` to check that encoder/decoder returns StatusNoMem when allocation fails.
* Add unit test to `test_block_writer_reader_errors.cpp` to check that writer/reader forward error from encoder/decoder to upper level.

runei:
Fixes #738

### DIFF
--- a/src/internal_modules/roc_fec/block_writer.cpp
+++ b/src/internal_modules/roc_fec/block_writer.cpp
@@ -120,8 +120,7 @@ status::StatusCode BlockWriter::write(const packet::PacketPtr& pp) {
     roc_panic_if(!pp);
 
     if (!alive_) {
-        // TODO(gh-183): return StatusDead
-        return status::StatusOK;
+        return status::StatusAbort;
     }
 
     validate_fec_packet_(pp);
@@ -131,9 +130,9 @@ status::StatusCode BlockWriter::write(const packet::PacketPtr& pp) {
     }
 
     if (cur_packet_ == 0) {
-        if (!begin_block_(pp)) {
-            // TODO(gh-183): return status
-            return status::StatusOK;
+        const status::StatusCode status_code = begin_block_(pp);
+        if (status_code != status::StatusOK) {
+            return status_code;
         }
     }
 
@@ -156,11 +155,13 @@ status::StatusCode BlockWriter::write(const packet::PacketPtr& pp) {
     return status::StatusOK;
 }
 
-bool BlockWriter::begin_block_(const packet::PacketPtr& pp) {
+status::StatusCode BlockWriter::begin_block_(const packet::PacketPtr& pp) {
     update_block_duration_(pp);
 
     if (!apply_sizes_(next_sblen_, next_rblen_, pp->fec()->payload.size())) {
-        return false;
+        roc_log(LogError,
+                "fec block writer: apply_sizes in begin_block_ failed with StatusNoMem");
+        return status::StatusNoMem;
     }
 
     roc_log(LogTrace,
@@ -168,15 +169,18 @@ bool BlockWriter::begin_block_(const packet::PacketPtr& pp) {
             (unsigned long)cur_sbn_, (unsigned long)cur_sblen_, (unsigned long)cur_rblen_,
             (unsigned long)cur_payload_size_);
 
-    if (!block_encoder_.begin_block(cur_sblen_, cur_rblen_, cur_payload_size_)) {
+    const status::StatusCode status_code =
+        block_encoder_.begin_block(cur_sblen_, cur_rblen_, cur_payload_size_);
+
+    if (status_code != status::StatusOK) {
         roc_log(LogError,
                 "fec block writer: can't begin encoder block, shutting down:"
                 " sblen=%lu rblen=%lu",
                 (unsigned long)cur_sblen_, (unsigned long)cur_rblen_);
-        return (alive_ = false);
+        alive_ = false;
     }
 
-    return true;
+    return status_code;
 }
 
 void BlockWriter::end_block_() {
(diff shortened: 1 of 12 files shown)


gavv:
BTW you can use NoopArena or MockArena to emulate allocation error.

gavv:
Another note: please don't use CHECK_EQUAL, it usually introduces bugs into your tests because it computes its arguments more than once. This may cause problems when its arguments have side effects.

```
#define CHECK_EQUAL_LOCATION(expected, actual, text, file, line)\
  do { if ((expected) != (actual)) { \
      if ((actual) != (actual)) \
      	  UtestShell::getCurrent()->print("WARNING:\n\tThe \"Actual Parameter\" parameter is evaluated multiple times resulting in different values.\n\tThus the value in the error message is probably incorrect.", file, line); \
      if ((expected) != (expected)) \
      	  UtestShell::getCurrent()->print("WARNING:\n\tThe \"Expected Parameter\" parameter is evaluated multiple times resulting in different values.\n\tThus the value in the error message is probably incorrect.", file, line); \
      UtestShell::getCurrent()->assertEquals(true, StringFrom(expected).asCharString(), StringFrom(actual).asCharString(), text, file, line); \
  } \
  else \
  { \
    UtestShell::getCurrent()->assertLongsEqual((long)0, (long)0, NULLPTR, file, line); \
  } } while(0)
```

src/internal_modules/roc_fec/block_writer.cpp
```diff
@@ -121,7 +121,7 @@ status::StatusCode BlockWriter::write(const packet::PacketPtr& pp) {
 
     if (!alive_) {
         // TODO(gh-183): return StatusDead
-        return status::StatusOK;
+        return status::StatusNoMem;
```
gavv, line 124: Please change it to StatusAbort, similar to BlockReader. Setting alive_ to false means that we need to abort the session. StatusDead is the outdated name for StatusAbort, so please also remove TODO comment.

We'll also need to modify tests. E.g. currently you have this:

```cpp
LONGS_EQUAL(status::StatusNoMem, writer.write(generate_packet(sn++)));
```

and we would need something like this:

```
LONGS_EQUAL(i == 0 ? status::StatusNoMem : status::StatusAbort,
                    writer.write(generate_packet(sn++)));
```

i.e. we check that the first call returned error, and all subsequent calls return StatusAbort (because alive became false).

src/internal_modules/roc_fec/block_writer.cpp
```diff
@@ -156,27 +156,32 @@ status::StatusCode BlockWriter::write(const packet::PacketPtr& pp) {
...
-bool BlockWriter::begin_block_(const packet::PacketPtr& pp) {
+status::StatusCode BlockWriter::handle_status_(status::StatusCode code,
+                                               const char* context) {
+    if (code != status::StatusOK) {
+        roc_log(LogError, "fec block writer: %s failed with status code %d", context,
+                code);
+        alive_ = false;
+    }
+    return code;
+}
```
gavv, line 167: I think this helper actually harms readability than improves it.

Yes, it reduces number of lines of code, but extracted code is trivial and it was not really increasing cognitive load. When you read it, you immediately see the idiom (if + logging) and easily skip it.

On the other hand, the helper introduces an indirection level and you need to check out what it's doing. It would be worth it if the help was hiding some complicated logic, but that's not the case.

I usually try to avoid helpers for trivial pieces, less lines does not always mean better code.

src/tests/roc_fec/test_block_writer_reader_errors.cpp
```diff
@@ -384,7 +384,7 @@ TEST(block_writer_reader_errors, reader_cant_decode_packet) {
     // reader should get an error from arena when trying
     // to repair lost packet and shut down
     packet::PacketPtr pp;
-    LONGS_EQUAL(status::StatusAbort, reader.read(pp, packet::ModeFetch));
+    LONGS_EQUAL(status::StatusNoMem, reader.read(pp, packet::ModeFetch));
     CHECK(!pp);
     CHECK(!reader.is_alive());
```
gavv, line 389: Could you also add a check that if we call read() one more time, now it will return StatusAbort?

gavv (commented):
Thanks for the patch! Here are a few comments.

gavv:
Thank for update, LGTM!

Could you **rebase** (not merge) your branch on fresh develop? Currently it's not rebasable due to conflicts.

PS. Please don't resolve discussions by yourself, usually reviewer does it when they re-check code. Instead you can use a thumbs up or leave a comment that the issue is addressed (if you need to track discussions for yourself).

runei:
> Thank for update, LGTM!
> 
> Could you **rebase** (not merge) your branch on fresh develop? Currently it's not rebasable due to conflicts.
> 
> PS. Please don't resolve discussions by yourself, usually reviewer does it when they re-check code. Instead you can use a thumbs up or leave a comment that the issue is addressed (if you need to track discussions for yourself).

Hi @gavv , the branch is rebased.

gavv:
Thank you!

PR #760, opened 2024-07-15
Implemented Moving Histogram algorithm

### LINKED ISSUE #752: Implement moving histogram algorithm
Add a class that implements moving histogram.

The idea is simple:

* a few parameters are given: value range, number of bins to split range into (N), and value window length (K)
* we allocate N bins, each one corresponds to sub-range of value range
* we allocate a ring buffer for last K values
* each bin holds a counter
* when a new value is added, we:
  * add it to ring buffer tail
  * remove oldest value from ring buffer head
  * increment counter of the bin corresponding to the added value
  * decrement counter of the bin corresponding to the removed value

The new class can be named MovHistogram. Interface should be similar to [core::MovStats](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/mov_stats.h) class, which implements moving minimum, maximum, and variance.

We need a method to add a value, and to get counter value for every bin.

Please use [core::RingQueue](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/ring_queue.h) for ring buffer and [core::Array](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_core/array.h) for bins.

This task is needed for #712.

spran180:
Fixes #752

Summary  
This PR implemented Moving Histogram Algorithm as stated in #752. This change includes the addition of a new class 'MovHistogram' and its associated unit tests.

### DIFF
--- a/src/internal_modules/roc_core/mov_histogram.h
+++ b/src/internal_modules/roc_core/mov_histogram.h
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2023 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file roc_core/mov_histogram.h
+//! @brief Rolling window moving histogram.
+
+#ifndef ROC_CORE_MOV_HISTOGRAM_H_
+#define ROC_CORE_MOV_HISTOGRAM_H_
+
+#include "roc_core/array.h"
+#include "roc_core/iarena.h"
+#include "roc_core/panic.h"
+#include "roc_core/ring_queue.h"
+
+namespace roc {
+namespace core {
+
+//! @brief A class that implements a rolling window moving histogram.
+//! The MovHistogram class maintains a histogram of values within a specified window
+//! length. It divides the range of values into a specified number of bins and updates the
+//! histogram as new values are added and old values are removed from the window.
+//! @tparam T The type of values to be histogrammed.
+
+template <typename T> class MovHistogram {
+public:
+    //! @brief Constructs a moving histogram.
+    //! @param arena Memory arena for dynamic allocations.
+    //! @param value_range_min The minimum value of the range to be histogrammed.
+    //! @param value_range_max The maximum value of the range to be histogrammed.
+    //! @param num_bins The number of bins in the histogram. Each bin represents a
+    //! subrange of the value range.
+    //! @param window_length The length of the moving window. Only values within this
+    //! window are considered in the histogram.
+
+    MovHistogram(IArena& arena,
+                 T value_range_min,
+                 T value_range_max,
+                 size_t num_bins,
+                 size_t window_length)
+        : value_range_min_(value_range_min)
+        , value_range_max_(value_range_max)
+        , num_bins_(num_bins)
+        , window_length_(window_length)
+        , ring_buffer_(arena, window_length)
+        , bins_(arena)
+        , valid_(false) {
+        if (num_bins == 0 || window_length == 0 || value_range_min >= value_range_max) {
+            roc_panic("mov histogram: number of bins and window length must be greater "
+                      "than 0 and value_range_min must be less than value_range_max");
+        }
+
+        bin_width_ = (value_range_max - value_range_min) / static_cast<T>(num_bins);
+
+        if (!ring_buffer_.is_valid() || !bins_.resize(num_bins)) {
+            return;
+        }
+
+        valid_ = true;
+    }
+
+    //! Check if the histogram is valid.
+    bool is_valid() const {
+        return valid_;
+    }
+
+    //! Add a value to the histogram.
+    void add_value(const T& value) {
+        T clamped_value = value;
+
+        if (clamped_value < value_range_min_) {
+            clamped_value = value_range_min_;
+        } else if (clamped_value > value_range_max_) {
+            clamped_value = value_range_max_;
+        }
+
+        if (ring_buffer_.size() == window_length_) {
+            T oldest_value = ring_buffer_.front();
+            ring_buffer_.pop_front();
+            size_t oldest_bin_index = get_bin_index_(oldest_value);
+            bins_[oldest_bin_index]--;
+        }
+
+        ring_buffer_.push_back(clamped_value);
+        size_t new_bin_index = get_bin_index_(clamped_value);
+        if (new_bin_index < num_bins_) {
+            bins_[new_bin_index]++;
+        }
+    }
+
+    //! Get the number of values in the given bin.
+    size_t get_bin_counter(size_t bin_index) const {
+        return bins_[bin_index];
+    }
+
+private:
+    //! Get the bin index for the given value.
+    size_t get_bin_index_(const T& value) const {
+        if (value == value_range_max_) {
+            return num_bins_ - 1;
+        }
+
+        return static_cast<size_t>((value - value_range_min_) / bin_width_);
+    }
+
+    T value_range_min_;
+    T value_range_max_;
+    size_t num_bins_;
+    size_t window_length_;
+    T bin_width_;
+    RingQueue<T> ring_buffer_;
+    Array<size_t> bins_;
+    bool valid_;
+};
+
+} // namespace core
+} // namespace roc
+
+#endif // ROC_CORE_MOV_HISTOGRAM_H_
--- a/src/tests/roc_core/test_mov_histogram.cpp
+++ b/src/tests/roc_core/test_mov_histogram.cpp
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2023 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <CppUTest/TestHarness.h>
+
+#include "roc_core/heap_arena.h"
+#include "roc_core/mov_histogram.h"
+
+namespace roc {
+namespace core {
+
+namespace {
+
+enum { NumObjects = 10, EmbeddedCap = 5 };
+
+struct Object {
+    static long n_objects;
+
+    size_t value;
+
+    Object(size_t v = 0)
+        : value(v) {
+        n_objects++;
+    }
+
+    Object(const Object& other)
+        : value(other.value) {
+        n_objects++;
+    }
+
+    ~Object() {
+        n_objects--;
+    }
+};
+
+long Object::n_objects = 0;
+
+} // namespace
+
+TEST_GROUP(movhistogram) {
+    HeapArena arena;
+};
+
+TEST(movhistogram, single_pass) {
+    const size_t value_range_min = 0;
+    const size_t value_range_max = 100;
+    const size_t num_bins = 10;
+    const size_t win_length = 10;
+
+    MovHistogram<size_t> hist(arena, value_range_min, value_range_max, num_bins,
+                              win_length);
+    CHECK(hist.is_valid());
+
+    for (size_t i = 0; i < win_length; i++) {
+        hist.add_value(i * num_bins);
+    }
+
+    for (size_t i = 0; i < num_bins; ++i) {
+        LONGS_EQUAL(1, hist.get_bin_counter(i));
+    }
+}
+
+TEST(movhistogram, rolling_window) {
+    const size_t value_range_min = 0;
+    const size_t value_range_max = 100;
+    const size_t num_bins = 10;
+    const size_t win_length = 5;
+
+    MovHistogram<size_t> hist(arena, value_range_min, value_range_max, num_bins,
+                              win_length);
+    CHECK(hist.is_valid());
+
+    for (size_t i = 0; i < win_length * 2; i++) {
+        hist.add_value(i * (value_range_max / num_bins));
+    }
+
+    for (size_t i = 0; i < num_bins; ++i) {
+        LONGS_EQUAL(i < win_length ? 0 : 1, hist.get_bin_counter(i));
+    }
+}
+
+TEST(movhistogram, value_equal_to_value_range_max) {
+    const size_t value_range_min = 0;
+    const size_t value_range_max = 100;
+    const size_t num_bins = 10;
+    const size_t win_length = 10;
+
+    MovHistogram<size_t> hist(arena, value_range_min, value_range_max, num_bins,
+                              win_length);
+    CHECK(hist.is_valid());
+
+    size_t test_value = value_range_max;
+    hist.add_value(test_value);
+
+    size_t expected_bin_index = num_bins - 1;
+    LONGS_EQUAL(1, hist.get_bin_counter(expected_bin_index));
+}
+
+} // namespace core
+} // namespace roc


gavv:
Hi, I see it's draft and haven't look at the code closely, but here a couple of preliminary comments:

* PRs should be targeted to develop branch, not master (see [here](https://roc-streaming.org/toolkit/docs/development/contribution_guidelines.html))
* Please follow code style (see [here](https://roc-streaming.org/toolkit/docs/development/coding_guidelines.html#coding-style))
* Template classes should be defined in header, not in .cpp

spran180:
Hi @gavv,

Thanks for reviewing my code. I will correct it shortly as per your suggestions.

spran180:
Hello @gavv,

I have addressed the issues you highlighted. The branch has been rebased to develop, and the code has been revised to align with the coding style. Everything should be in order now. Kindly inform me if any further changes or checks are needed.

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,100 @@
...
+//! Efficiently implements moving histogram based on approach
+/// @tparam T
+template <typename T> class MovHistogram {
+public:
+    //! Initialize.
+    MovHistogram(IArena& arena, T value_range, size_t num_bins, size_t window_length)
+        : value_range_(value_range)
+        , num_bins_(num_bins)
+        , window_length_(window_length)
+        , bin_width(static_cast<T>(value_range) / static_cast<T>(num_bins))
```
adrianopol, line 33: I suppose you have  wrong understanding of what `value_range` is: it is supposed to be a pair of values -- $`(min, max)`$, not just a single value. So each $`bin_i`$ $`\left(i \in \left[0,N\right)\right)`$ will be used for counting values in range
$$\left[ min + \dfrac{i(max-min)}{N}, min + \dfrac{(i+1)(max-min)}{N} \right).$$

In your case a range is always a $`\left[0, max\right)`$ which means a loss of generality.

@gavv will correct me if I am wrong.

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,100 @@
...
+        , ring_buffer_(arena, window_length)
+        , bins_(arena)
+        , valid_(false) {
+        if (num_bins == 0 || window_length_ == 0) {
```
adrianopol, line 37: it would be great not to mix parameters and member values in a single expression

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,100 @@
...
+
+        ring_buffer_.push_back(value);
+        int new_bin_index = getBinIndex(value);
+        if (new_bin_index >= 0 && static_cast<size_t>(new_bin_index) < num_bins_) {
+            bins_[static_cast<size_t>(new_bin_index)]++;
+        }
+    }
+
+    //! returns the number of values in the bin
+    int getBinCounter(size_t bin_index) const {
```
adrianopol, line 74: style: we do not use CamelCase for function naming for C language anywhere in this project; please use underscores instead

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,100 @@
...
+
+        return static_cast<int>(value / bin_width);
+    }
+
+    T value_range_;
+    size_t num_bins_;
+    size_t window_length_;
+    T bin_width;
+    RingQueue<T> ring_buffer_;
+    Array<int> bins_;
```
adrianopol, line 93: IIUC, here we should use `T` instead of `int`

src/internal_modules/roc_core/mov_histogram.h
adrianopol, line 1: I didn't look into the `add_value()` thoroughly because it is going to be fixed according to other comments. But it has some problems too. Please, read the task definition, especially these 2 points:

> - increment counter of the bin corresponding to the added value
> - decrement counter of the bin corresponding to the removed value

adrianopol (changes requested):

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,100 @@
...
+    //! returns the number of values in the bin
+    int getBinCounter(size_t bin_index) const {
+        return bins_[bin_index];
+    }
+
+private:
+    //! returns the bin index for a given value
+    int getBinIndex(const T& value) const {
+        if (value > value_range_) {
+            return -1;
```
adrianopol, line 82: Consider validating `value` in `add_value()` instead of using a placeholder here. Therefore the returning type will be size_t.

I asked Victor, he told me you can use saturation there:
- if value > max => value = max
- if value < min => value = min

adrianopol (changes requested):

src/internal_modules/roc_core/mov_histogram.h
gavv: I agree, I also had this in mind.

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,100 @@
...
+    MovHistogram(IArena& arena, T value_range, size_t num_bins, size_t window_length)
+        : value_range_(value_range)
+        , num_bins_(num_bins)
+        , window_length_(window_length)
+        , bin_width(static_cast<T>(value_range) / static_cast<T>(num_bins))
+        , ring_buffer_(arena, window_length)
+        , bins_(arena)
+        , valid_(false) {
+        if (num_bins == 0 || window_length_ == 0) {
+            roc_panic(
```
gavv, line 38: We should do this check before dividing by num_bins.

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,100 @@
...
+
+    //! Check if the histogram is valid.
+    bool is_valid() const {
+        return valid_;
+    }
+
+    //! Add a value to the histogram.
+    void add_value(const T& value) {
+        if (ring_buffer_.size() == window_length_) {
+            T oldest_bin_ = ring_buffer_.front();
```
gavv, line 57: I think oldest_value would be a better name? Also the var can be const.

src/internal_modules/roc_core/mov_histogram.h
gavv: And also please add trailing underscore to private methods and fields.

gavv (commented):
Thanks for PR! A couple of comments from me as well.

spran180:
Thank you for the review, @adrianopol and @gavv. I will look into it and update the code accordingly.

spran180:
Hey @adrianopol, I'm using `pair<T, T>` to represent the value range `{min_val, max_val}` instead of creating separate variables. Is this approach correct?

gavv:
> Hey @adrianopol, I'm using `pair<T, T>` to represent the `value_range` `{min_val, max_val}` instead of creating separate variables. Is this approach correct?

We don't use STL containers, so please use separate arguments. See here: https://roc-streaming.org/toolkit/docs/development/coding_guidelines.html

spran180:
Hey @gavv / @adrianopol, 

I've addressed the changes that you've requested pls go through it once and let me know if any further changes are required.

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,111 @@
...
+            clamped_value = value_range_min_;
+        } else if (clamped_value > value_range_max_) {
+            clamped_value = value_range_max_;
+        }
+
+        if (ring_buffer_.size() == window_length_) {
+            T oldest_value = ring_buffer_.front();
+            ring_buffer_.pop_front();
+            size_t oldest_bin_index = get_bin_index_(oldest_value);
+            if (oldest_bin_index < num_bins_) {
```
adrianopol, line 75: Is there any case when this `if` returns false? If not, you may drop it.

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,111 @@
...
+
+    //! returns the number of values in the bin
+    size_t get_bin_counter(size_t bin_index) const {
+        return bins_[bin_index];
+    }
+
+private:
+    //! returns the bin index for a given value
+    size_t get_bin_index_(const T& value) const {
+        return static_cast<size_t>((value - value_range_min_) / bin_width_);
```
adrianopol, line 95: There is a corner case (see the right parenthesis my previous comment which mean a non-inclusive range: https://github.com/roc-streaming/roc-toolkit/pull/760#discussion_r1684971776): if `value == value_range_max`, bin index will be `num_bins_` (which means N+1), which is incorrect.

There are multiple solutions how to deal with it. One is to add a branch:
```
if (value == value_range_max_) {
    return num_bins_ - 1;
}
return ...
```

This should fix the problem in line 75. Also it would be nice to add a test for this case.

adrianopol (changes requested):

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        ring_buffer_.push_back(clamped_value);
+        size_t new_bin_index = get_bin_index_(clamped_value);
+        if (new_bin_index < num_bins_) {
+            bins_[new_bin_index]++;
+        }
+    }
+
+    //! returns the number of values in the bin
```
adrianopol, line 87:
```suggestion
    //! Get the number of values in the given bin.
```

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,111 @@
...
+    size_t get_bin_counter(size_t bin_index) const {
+        return bins_[bin_index];
+    }
+
+private:
+    //! returns the bin index for a given value
```
adrianopol, line 93:
```suggestion
    //! Get the bin index for the given value.
```

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,123 @@
...
+        , window_length_(window_length)
+        , ring_buffer_(arena, window_length)
+        , bins_(arena)
+        , valid_(false) {
+        if (num_bins == 0 || window_length == 0 || value_range_min >= value_range_max) {
+            roc_panic("mov histogram: number of bins and window length must be greater "
+                      "than 0 and value_range_min must be less than value_range_max");
+        }
+
+        bin_width_ = (value_range_max - value_range_min) / static_cast<T>(num_bins);
```
spran180, line 57: Hey @adrianopol,

Since `num_bins` is initialized as `size_t`, it needs to be cast to type `T`. Please let me know if this is incorrect.

adrianopol: This is correct, I just mean that C/C++ already do this implicit promotion (see https://en.cppreference.com/w/cpp/language/implicit_conversion): if at least on of operands in an arithmetic expression have type float (or double) then the whole expression will have type float (or double).

If T = float, then without casting we have `(float - float) / size_t` == `float / size_t` == `float`.

If T = `short`, we have `(short - short) / size_t` == `short / size_t` == `size_t`. Value (max - min) is always positive, so we have no problem with loosing sign here.

You may leave your variant as is, it is correct too.

adrianopol (approved):

src/tests/roc_core/test_mov_histogram.cpp
```diff
@@ -0,0 +1,109 @@
...
+    const size_t value_range_min = 0;
+    const size_t value_range_max = 100;
+    const size_t num_bins = 10;
+    const size_t win_length = 10;
+
+    MovHistogram<size_t> hist(arena, value_range_min, value_range_max, num_bins,
+                              win_length);
+    CHECK(hist.is_valid());
+
+    size_t values[win_length];
```
adrianopol, line 59: What is the purpose of this value? I think, you can eliminate it here and in the second test.

src/internal_modules/roc_core/mov_histogram.h
```diff
@@ -0,0 +1,123 @@
...
+        ring_buffer_.push_back(clamped_value);
+        size_t new_bin_index = get_bin_index_(clamped_value);
+        if (new_bin_index < num_bins_) {
+            bins_[new_bin_index]++;
+        }
+    }
+
+    //! Get the number of values in the given bin.
+    size_t get_bin_counter(size_t bin_index) const {
+        return bins_[bin_index];
```
adrianopol, line 97: if `bin_index` > `num_bins_` we should panic here

adrianopol (changes requested):
@spran180 , thank you for implementation.

We also need more test cases here, for example:
- T = float
- min and max < 0
- win_length == 1
They may be added in a different PR if you wish.

src/internal_modules/roc_core/mov_histogram.h
gavv: If T is `short`, then `(value_range_max - value_range_min) / num_bins` is size_t, and assigning size_t to bin_width_ would trigger a warning.

If T is `ssize_t` (signed version of size_t), then `(value_range_max - value_range_min) / num_bins` would trigger a warning.

So I think we need to keep the cast here.

BTW we usually use c-style casts or constructor-style casts instead of static_cast, but that's not a big deal.

gavv:
> @spran180 , thank you for implementation.
> 
> We also need more test cases here, for example:
> 
>     * T = float
> 
>     * min and max < 0
> 
>     * win_length == 1
>       They may be added in a different PR if you wish.

I agree.

Also, current tests don't cover a few important aspects:

* when there are multiple values in every bin (currently every bin is either 0 or 1)
* how bin counters increment and decrement when rolling window is moved (e.g. you call add_value, some bin changes from N to N-1, and another bin changes from K to K+1) 
* out of bounds values (clamping)

@spran180 Let me know if you would like to add these tests or it's better to create a new issue for that.

src/internal_modules/roc_core/mov_histogram.h
gavv: Actually `Array` already has bound checking.

adrianopol: ah, OK

src/tests/roc_core/test_mov_histogram.cpp
spran180: The `values` array used to store the added values that we add to the Histogram, but I think it's better to remove it.

spran180:
> > @spran180 , thank you for implementation.
> > We also need more test cases here, for example:
> > ```
> > * T = float
> > 
> > * min and max < 0
> > 
> > * win_length == 1
> >   They may be added in a different PR if you wish.
> > ```
> 
> I agree.
> 
> Also, current tests don't cover a few important aspects:
> 
> * when there are multiple values in every bin (currently every bin is either 0 or 1)
> * how bin counters increment and decrement when rolling window is moved (e.g. you call add_value, some bin changes from N to N-1, and another bin changes from K to K+1)
> * out of bounds values (clamping)
> 
> @spran180 Let me know if you would like to add these tests or it's better to create a new issue for that.

Yes, I believe I should create a separate PR to add tests.

spran180:
Hey @gavv / @adrianopol,

Could you please guide me on how to pass the `build / linux-x86_64/ubuntu-24.04 (pull_request)` test that is failing after 6 seconds?

gavv:
The build failure was docker hub timeout, not related to the PR. I've restarted the build and it's gone.

> Yes, I believe I should create a separate PR to add tests.

@spran180 Great, do you plan to work on it, or do I create an issue for the future?

spran180:
> The build failure was docker hub timeout, not related to the PR. I've restarted the build and it's gone.
> 
> > Yes, I believe I should create a separate PR to add tests.
> 
> @spran180 Great, do you plan to work on it, or do I create an issue for the future?

Indeed, I am interested in working on it.

gavv:
Thanks for PR!

> Indeed, I am interested in working on it.

Perfect. I've copied Andrew's and mine comments here: https://github.com/roc-streaming/roc-toolkit/issues/752#issuecomment-2254591934

PR #763, opened 2024-07-21
Implemented explicit flushing of buffered samples from sink to file

### LINKED ISSUE #703: Add ISink::flush()
Implement explicit flushing of buffered samples from sink to file.

* add `virtual void flush() = 0` to sndio::ISink
* add this method to all ISink implementations
* for most implementations, flush() should be no-op, as they don't use buffering
* for SoxSink, flush() should send buffered samples to disk
* sndio::Pump should invoke flush() when it exits from run()

jeshwanthreddy13:
Fixes #703 
- Added `virtual ROC_ATTR_NODISCARD status::StatusCode flush()=0` pure virtual function in `sndio::ISink`.
- Added this method to all ISink implementations.
- Implementations that don't use buffereing are no-op.
- For SoxSink, flush() sends the buffered samples to disk.
- sndio::Pump invokes `flush()` when it exits from `run()`.

### DIFF
--- a/src/internal_modules/roc_pipeline/sender_sink.h
+++ b/src/internal_modules/roc_pipeline/sender_sink.h
@@ -115,6 +115,9 @@ public:
     //! Write frame.
     virtual ROC_ATTR_NODISCARD status::StatusCode write(audio::Frame& frame);
 
+    //! Flush buffered data, if any.
+    virtual ROC_ATTR_NODISCARD status::StatusCode flush();
+
 private:
     SenderSinkConfig sink_config_;
 
--- a/src/internal_modules/roc_sndio/pump.cpp
+++ b/src/internal_modules/roc_sndio/pump.cpp
@@ -71,6 +71,13 @@ status::StatusCode Pump::run() {
     if (code == status::StatusFinish) {
         code = status::StatusOK; // EOF is fine
     }
+    if (code == status::StatusOK) {
+        code = sink_.flush();
+        if (code != status::StatusOK) {
+            roc_log(LogError, "pump: got error when flushing sink: status=%s",
+                    status::code_to_str(code));
+        }
+    }
 
     status::StatusCode close_code = close_all_devices_();
 
--- a/src/internal_modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
+++ b/src/internal_modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
@@ -219,9 +219,21 @@ status::StatusCode SoxSink::write(audio::Frame& frame) {
         }
     }
 
-    const status::StatusCode code = write_(buffer_data, buffer_pos);
-    if (code != status::StatusOK) {
-        return code;
+    if (buffer_pos > 0) {
+        const status::StatusCode code = write_(buffer_data, buffer_pos);
+        if (code != status::StatusOK) {
+            return code;
+        }
+    }
+
+    return status::StatusOK;
+}
+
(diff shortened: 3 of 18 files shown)


src/internal_modules/roc_pipeline/sender_sink.h
```diff
@@ -109,6 +109,9 @@ class SenderSink : public sndio::ISink, public core::NonCopyable<> {
     //! Write frame.
     virtual ROC_ATTR_NODISCARD status::StatusCode write(audio::Frame& frame);
 
+    //! Flush buffered samples to disk
+    virtual ROC_ATTR_NODISCARD status::StatusCode flush();
```
gavv, line 113: Here and in other headers: it's not necessarily flushing *to disk* because e.g. there are pipeline sinks that will later implement flushing to network queue.

We can reword it like:

> Flush buffered data, if any.

src/internal_modules/roc_sndio/pump.cpp
```diff
@@ -74,7 +74,7 @@ status::StatusCode Pump::run() {
 
     roc_log(LogDebug, "pump: exiting main loop");
 
-    return code;
+    return sink_.flush();
```
gavv, line 77: If code was not StatusOK before flushing, we should return original code. Only if there were no previous errors before flush and code was StatusOK, we should return code from flush.

(It is a good practice to return earliest error happened, because later errors can be caused by the very first one).

Currently, we just drop previous code, which is wrong.

gavv: Also:

* please move flush call before "exiting main loop" message
* if flush fails, please log error and status (see status::code_to_str)

src/internal_modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
```diff
@@ -217,6 +217,14 @@ status::StatusCode SoxSink::write(audio::Frame& frame) {
     return status::StatusOK;
 }
 
+status::StatusCode SoxSink::flush() {
+    if (fflush((FILE*)output_->fp) == 0) {
+        return status::StatusOK;
+    }
+
+    return status::StatusErrFile;
+}
```
gavv, line 226: 1. `output_` may be NULL, when sink is paused. I think flush should be no-op in this case. (BTW we need to address this in write() too..)
2. `output_->fp` can't be used if SoxSink is device. We should use it only if `driver_type_ == DriverType_File`, otherwise flush should be no-op.
3. Let's also check that `output_->fp` is non-NULL. If it's null, let's make flush a no-op too.
4. nit: Usually early returns are used for errors and the very last return for success.

gavv (commented):
Thanks for PR! A few comments.

src/internal_modules/roc_pipeline/sender_sink.h
jeshwanthreddy13: Done

src/internal_modules/roc_sndio/pump.cpp
jeshwanthreddy13: Done

src/internal_modules/roc_sndio/target_sox/roc_sndio/sox_sink.cpp
jeshwanthreddy13: Done, also implemented a fix in write() where buffer will write to sox only if `buffer_pos > 0`

gavv:
Thank you!

Issue #758, opened 2024-07-13
Implement IFrameEncoder / IFrameDecoder using libvorbis


gavv:
We have two interfaces [IFrameEncoder](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_audio/iframe_encoder.h) and [IFrameDecoder](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_audio/iframe_decoder.h) that are used to encode audio samples into packet payload and decode it back. Currently they're implemented using PcmEncoder and PcmDecoder (for uncompressed PCM).

Now we need to add two more implementations: VorbisEncoder and VorbisDecoder, that will use [libvorbis](https://xiph.org/vorbis/) library to do the job.

We also need to add new encoder & decoder to the list of tested codecs in [test_frame_encoder_decoder.cpp](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/tests/roc_audio/test_frame_encoder_decoder.cpp).

runei:
Hi, I am interested in solving this issue.

gavv:
Welcome, thank you!

gavv:
Let me know if you'll need help/hints with the build system. 

Also, if libogg dependency can be disabled, I suppose we should do it, because all we need is vorbis encoder/decoder, we don't need to work with ogg files.

runei:
Hi @gavv ,

I tried without the libogg, but it fails when installing vorbis. According to the README, "You'll need libogg (distributed separately) to compile this library.", which is why I tried to install it before libvorbis.

I managed to install vorbis with the build script after installing libogg. But I'm having trouble in installing the libogg with the script, since I used "make install" on the terminal.
Using "execute(ctx, 'make install')" on the script need permission on CI, as it creates the files at " /usr/local/lib". 
So how can I install libogg within the same folder and utilize the generated files for the libvorbis installation?

gavv:
Oh, I see. I assumed that libogg should support using libvorbis as codec, but instead, libvorbis supports using libogg for muxing.

For [rfc5215](https://datatracker.ietf.org/doc/html/rfc5215), we'll need to disable ogg muxing, which seems to be feasible. libvorbis encodes packets into a struct from libogg (`ogg_packet`), however this struct does not yet have OGG headers, only opaque codec-specific data. OGG headers are added later when you use libogg API, and we can just don't use it and use data from ogg_packet directly.

-----

[...]

PR #768, opened 2024-07-28
Implement IFrameEncoder / IFrameDecoder using libvorbis

runei:
Implements #758


runei:
Hi @gavv  ,

The libvorbis build is successful for all targets except macos14-arm64/build-3rdparty and macos14-arm64/universal-binaries. But it works fine with macos14-arm64/standard-build and other macos builds.

The error message is:

```
*** Could not run Ogg test program, checking why...
*** The test program failed to compile or link. See the file config.log for the
*** exact error that occurred. This usually means Ogg was incorrectly installed
*** or that you have moved Ogg since it was installed.
configure: error: must have Ogg installed!
```

But according to the logs, libogg was installed before this error.

What is the difference between the macos 3rdparty, universal-binaries, to the other macos standard-build?
Do you have any suggestion on how to solve this issue?

gavv:
> Hi @gavv ,
> 
> The libvorbis build is successful for all targets except macos14-arm64/build-3rdparty and macos14-arm64/universal-binaries. But it works fine with macos14-arm64/standard-build and other macos builds.
> 
> The error message is:
> 
> ...

Most likely, `configure` from libvorbis is trying to find libogg using pkg-config instead of using paths that we gave to it. In this case we can prevent it by providing `OGG_CFLAGS` and `OGG_LIBS` variables to `configure`, the same way as we do it with `LIBSNDFILE_` when we're building pulseaudio (also in build-3rdparty.py).

PR #773, opened 2024-07-31
Added more Tests for Moving Histogram Algorithm

### LINKED ISSUE #752: Implement moving histogram algorithm

spran180:
Fixes #752

+ Added unit test for `T = float`.
+ Added unit test for cases when range min and max are negative.
+ Added unit test for case when window length is equal to one.
+ Added unit test for case when there are multiple values in every bin.
+ Added unit test to see bins change when window is moved.
+ Added unit test for out of bounds values.

### DIFF
--- a/src/tests/roc_core/test_mov_histogram.cpp
+++ b/src/tests/roc_core/test_mov_histogram.cpp
@@ -101,5 +101,145 @@ TEST(movhistogram, value_equal_to_value_range_max) {
     LONGS_EQUAL(1, hist.get_bin_counter(expected_bin_index));
 }
 
+TEST(movhistogram, value_is_float) {
+    const float value_range_min = 0.0f;
+    const float value_range_max = 100.0f;
+    const size_t num_bins = 10;
+    const size_t win_length = 10;
+
+    MovHistogram<float> hist(arena, value_range_min, value_range_max, num_bins,
+                             win_length);
+    CHECK(hist.is_valid());
+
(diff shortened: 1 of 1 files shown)


gavv:
Thank you!

A few cosmetic fixes: 2314fa4498b1cec19014cddc80ead534ceaf20a6

Issue #712, opened 2024-04-02
Dynamic latency improvements


gavv:
Follow-up for #688, tasks for future releases:

- [ ] Don't start session from fixed latency. Instead, start from min latency, wait until accumulated amount becomes larger than jitter or max latency, and then start session.

- [ ] Don't take into account short jitter spikes. Maybe use running percentile instead of running maximum.

gavv:
@baranovmv I did a few benchmarks and have an interesting observation.

Please take a look at the `Jitter + Niq + Losses` plot in the end of reports in this benchmark: https://github.com/roc-streaming/qa/tree/main/reports/20250210_gh712_peak_jitter

(Most reports have this section, but not all of them)

For example, this one: https://github.com/roc-streaming/qa/blob/main/reports/20250210_gh712_peak_jitter/05--2ghz--no_load--low_res/README.md

![Image](https://github.com/user-attachments/assets/3fa07252-ca42-4062-8959-f3769d71704b)

This plot shows three metrics from the other sections combined together:

- network incoming queue (niq) spikes (sudden decreasements of NIQ len, in milliseconds)
- current jitter (measured from RTS)
- peak jitter (calculated from jitter)
- missing samples (unrecovered losses, how many samples were missing and replaced with zeros)

On the plot we can see that a few times the following happened:

- we got some unrecovered losses (red bar)
- we got high NIQ spike (yellow dot) that is above both jitter (blue line) and peak jitter (orange line)
- but we got no spike in jitter; it was still lower than current peak jitter, so there was no reaction to this

In other words, in this cases, we *had* information that current latency is not enough and should be increased, we we didn't use it. I guess on burst losses RTS alone is not enough to see spikes. NIQ have this information, on the other hand.

So the question is, do we want to take NIQ into account (somehow) when calculating jitter or target latency?

gavv:
**WIP** benchmark for new jitter meter algorithm: https://github.com/roc-streaming/qa/tree/main/reports/20250214_gh712_jitter_sim

It includes current jitter meter implementation ported from C++ to python, and a few tests that get jitter and peak jitter from csv dumps (from [previous benchmark](https://github.com/roc-streaming/qa/tree/main/reports/20250210_gh712_peak_jitter)) and check that python port of jitter meter calculates identical peak jitter from jitter.

[...]

PR #776, opened 2024-08-11
Improve peak jitter calculations

gavv:
gh-712

- Extract JitterMeter class.
- Remove `min_jitter` metric (it is almost always zero or very close to it).
- Replace `max_jitter` with `peak_jitter`, which is similar to maximum, but tries to exclude harmless spikes to reduce latency.

See comments in JitterMeter for details on the algorithm.

QA: https://github.com/roc-streaming/qa/tree/main/manual/20241011_gh688_adaptive_latency

### DIFF
--- a/src/internal_modules/roc_audio/jitter_meter.cpp
+++ b/src/internal_modules/roc_audio/jitter_meter.cpp
@@ -1,0 +1,152 @@
+/*
+ * Copyright (c) 2024 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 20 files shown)


gavv:
As discussed with Mike, this algorithm will be only temporary solution not to be released, as we're unable to conclude if it's generic enough, especially magic numbers that it uses.

Before releasing 0.5, we should do 2 changes:

1. rewrite peak jitter calculation, most likely use histogram of scaled samples (e.g. using logarithm) instead of peak detector + quantile
2. re-do QA
3. replace MovAggregate with two classes: MovMeanStd and MovMinMax

Mike, could you create a task for (1)+(2)? No need to make it super specific, but would be nice if you can share what data you would like to see in QA reports (given that we still use current tooling) and some ideas for the new algorithm.

Issue #737, opened 2024-06-23
Return status code instead of bool in IParser


gavv:
## Summary

[packet::IParser](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_packet/iparser.h) is interface for protocol-specific deserializators. See [documentation](https://roc-streaming.org/toolkit/docs/internals/packets_frames.html).

Currently its methods return bool (true on success or false on error). We need to replace bool with [status::StatusCode](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_status/status_code.h) and return code that described why the operation failed.

## Implementation

* Update IParser interface and implementations:
   * rtp::Parser
   * fec::Parser
   * rtcp::Parser
* Report appropriate statuses:
   * when allocation failed, return StatusNoMem
   * when buffer is too small, return StatusBadBuffer
   * when packet is malformed, return StatusBadPacket
   * on success, return StatusOK
* Update users of parsers and composers. They should forward status to upper level:
   * fec::BlockReader
   * rtcp::Communicator

## Testing

* Add unit tests for rtp::Parser and fec::Parser that check returned statuses in case of errors.
* Add unit tests for components which use parsers, and check that they forward statuses from parser to the upper level (e.g. if Parser fails with StatusNoMem, BlockReader::read() returns StatusNoMem). One test per component should be enough.

sprinkuls:
hey, is it alright if i take this up?

gavv:
@sprinkuls Sure, thanks!

gavv:
Implementation merged, for tests I've created a separate issue: #792

PR #780, opened 2024-10-03
Return status code instead of bool in IParser

sprinkuls:
For #737


gavv:
Thanks for PR! Please ping me when it's ready for review.

gavv:
Thanks for the patch.

Since this PR is old, I've fixed issues by myself. Follow-up commit: f2ef51b418388ccc5a52a13d1945634e4053a734

Follow-up issue to add tests: #792

PR #781, opened 2024-10-04
Updated insert_sleeping_task function to try to improve the complexity time

JimmyL04:
This PR tried improves the time complexity of the ControlTaskQueue::insert_sleeping_task_ function by replacing the sorted linked list used for delayed tasks with a `std::priority_queue`.

Key Improvements:
- The previous implementation used a sorted linked list, leading to **O(N)** insertion time due to the need to traverse the list to find the correct insertion point.
- The new implementation uses a **priority queue** (min-heap), reducing the insertion time complexity to **O(log N)** while maintaining efficient pop operations.

Changes Made:
- Replaced the sorted linked list with `std::priority_queue`, simplifying the logic for task insertion.
- The priority queue is configured to prioritize tasks based on their `effective_deadline_`, ensuring the task with the earliest deadline is always at the top.
- The overall complexity for inserting tasks is now **O(log N)**, providing better scalability as the number of tasks grows.


gavv:
Hi,

Besides that it doesn't compile and doesn't have actual code (?), please check out discussion and requirements in the original issue for why std::priority_queue is not an option here: #379

Issue #734, opened 2024-06-20
Implement freelist class


gavv:
Subtask extracted from #602. See that task for rationale.

Add new class `core::Freelist<T>`, which implements lock-free free list based on this article:
[**Solving the ABA Problem for Lock-Free Free Lists**](https://moodycamel.com/blog/2014/solving-the-aba-problem-for-lock-free-free-lists.htm)

It should be modeled after `core::List<T>`. It should implement lock-free LIFO based on singly linked list, with just two operations: push_back() and pop_back(). It should be intrusive, just like `core::List`, i.e. elements should inherit node class, so that node data is embedded into element instead of being allocated separately.

We also need to cover it with unit tests.

nataliayave:
Hi, I am interesting in solving this issue. This would be my first time contributing to an open source project. Where would I find where this issue is taking place? Which file should I be working on?

gavv:
Hi, thanks. 

It belongs to roc_core module. See here: https://roc-streaming.org/toolkit/docs/internals/code_structure.html

mihir-mihir:
I'd be interested in giving this one and/or #602 a shot, @nataliayave are you still working on this? I can pick up #602 after if you are, otherwise I can take both

gavv:
@mihir-mihir While we're waiting for reply from Natalia, probably you'd be interested in #749, which is also about lock-free programming.

mihir-mihir:
Thanks, I've commented on #749

gavv:
@mihir-mihir I guess we can assume that the issue is free, please ping me if/when you want to be assigned.

mihir-mihir:
@gavv yes I'll take this one thanks

gavv:
Awesome

veronikaro:
Hey @mihir-mihir, are you still working on this? 
@gavv if not, could you please assign me to the issue? I would like to give it a go.

mihir-mihir:
Hi @veronikaro, go ahead

Ahilan001:
Hi, I know I didn't request the issue here, but I have been working on it too. I am going to finish working on the issue today, so would it be okay to make a pull request when I finish it?

gavv:
@Ahilan001 The patch you sent is confusing, looks like some random pieces of code?.. And it doesn't compile. Also please read coding guidelines: https://roc-streaming.org/toolkit/docs/development/coding_guidelines.html

Veronika, it would be great if you could take this issue :) @veronikaro

[...]

PR #782, opened 2024-10-06
Issue #734 Solution

Ahilan001:
Added the Freelist class with functions push_back and pop_back based on Issue #734. The implementation is based on the article provided in the issue and the List<T> class.


gavv:
https://github.com/roc-streaming/roc-toolkit/issues/734#issuecomment-2397669837

Issue #674, opened 2024-01-20
RTCP support improvements


gavv:
Follow-up for #14.

- [ ] Implement rtcp::IntervalComputer (computes packet generation interval according to algorithm from RFC 3550) and use in rtcp::Communicator instead of hard-coded 200ms report interval.

- [ ] Implement jitter buffer for incoming RTCP reports. Adjust rtt tests to ensure that jitter buffer improves RTT & clock offset estimation precision.

- [ ] Implement RTCP BYE exchange. Handle exit by signal and ask pipeline to generate BYE. Processing of BYE is already implemented.

- [ ] Implement RTCP multicast support. If receiver RTCP endpoint is bound to multicast address, it should use Report_ToAddress instead of Report_Back mode.

gavv:
Results of first 3 tests - good wifi, bad wifi, and ethernet.

* https://github.com/roc-streaming/qa/blob/main/manual/20241230_gh674_clock_offset/01_wifi5ghz_rpi_pc.md
* https://github.com/roc-streaming/qa/blob/main/manual/20241230_gh674_clock_offset/02_wifi2ghz_rpi_pc.md
* https://github.com/roc-streaming/qa/blob/main/manual/20241230_gh674_clock_offset/03_ether_rpi_pc.md

I think we need to try more aggressive smoothing of clock offset, which is especially notable on the second tests where RTT jitter was high.

However, even in that test, clock offset jitter was below 1ms, and 95-percentile of clock offset jitter was 60 microseconds, which I think is already cool!

---

Note: all these tests were done *without* SCHED_RR (no root) and when input & output are files. 

I plan to do tests with SCHED_RR separately to see what effect does it have. (BTW we probably want SCHED_RR in both netio and sndio threads, not just netio).

PulseAudio & ALSA are to be handled separately as well (I guess not in this issue).

---

My next step will be to pass clock offset to e2e latency, enable latency tuning based on e2e, and build test bench to measure actual latency using signal-estimator. These experiments are going to be quite time consuming, so before I start:

1. Do we want to try any other modifications of the algorithm?
2. Do we want to include additional statistics in measurements and reports?

@baranovmv Let me know if you have suggestions.

baranovmv:
Thank you for so thorough evaluation

It might be more representative  to compute `RTT` and `Clock offset` statistics on data which wasn't smoothed by median

```
    RttOffsetPair p;
    p.rtt = rtt;
    p.offset = clock_offset;
[...]

Issue #127, opened 2017-12-07
Total latency calculation


gavv:
Currently, the user can configure the latency between the receiving socket and the audio output (e.g. PulseAudio).

Thus, we don't take into account two important components when reporting/adjusting latency:

* network latency (RTT/2)
* output latency (e.g. PulseAudio/ALSA)

mildsunrise:
I understand this covers implementing SenderSink::latency(), right?

gavv:
I'm not sure why are you asking, so here is some context about that latency() method.

Most likely SenderSink::latency() is not affected by this task and is not actually used currently.

ISink::latency() is used in sndio::Pump. Pump copies samples from ISource to ISink. It uses ISink::latency() to compute current playback position and pass it to ISource::relock(), so that source can adjust its clock. This is the only usage of ISink::latency() IIRC.

Currently, pump is used only in two ways: ISource is file/device, and ISink is SenderSink, or ISource is ReceiverSource and ISink is file/device.

File or device can't adjust its clock, and ignores reclock() calls. It's always either SenderSink or ReceiverSource who adjusts the clock. So.. If ISink is SenderSink, currently it means that ISource is file/device, and relock() is not used, and thus value of SenderSink::latency() is irrelevant.

It may become relevant in future if we'll implement the case when ISource is ReceiverSource and ISink is SenderSink (i.e. transcoder/relay).

PR #786, opened 2024-12-04
Preparation for E2E latency tuning

baranovmv:
gh-127

In order to be able to tune receiver's latency
relying on timestamp mapping that we get from
RTCP feedback, and UDP::Receive_timestamp,
adding these features:

* gh-674: Use receive timestamp (RTS) as report time when processing RTCP report;

* RTT dumping for debugging (csvplotter ts_offset branch);

* SCHED_RR for network io thread (run with root privs).

### DIFF
--- a/src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
+++ b/src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
@@ -46,13 +46,14 @@ uint64_t Thread::get_tid() {
 #endif
 }
 
-bool Thread::enable_realtime() {
+bool Thread::enable_realtime(const int sched_prio) {
     sched_param param;
     memset(&param, 0, sizeof(param));
-    param.sched_priority = sched_get_priority_max(SCHED_RR);
+    param.sched_priority = sched_prio;
 
+    roc_log(LogInfo, "thread: set realtime priority");
     if (int err = pthread_setschedparam(pthread_self(), SCHED_RR, &param)) {
-        roc_log(LogDebug,
+        roc_log(LogError,
                 "thread: can't set realtime priority: pthread_setschedparam(): %s",
                 errno_to_str(err).c_str());
         return false;
--- a/src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
+++ b/src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
@@ -51,7 +51,8 @@ bool Thread::enable_realtime(const int sched_prio) {
     memset(&param, 0, sizeof(param));
     param.sched_priority = sched_prio;
 
-    roc_log(LogInfo, "thread: set realtime priority");
+    roc_log(LogDebug, "thread: setting realtime priority to %d", sched_prio);
+
     if (int err = pthread_setschedparam(pthread_self(), SCHED_RR, &param)) {
         roc_log(LogError,
                 "thread: can't set realtime priority: pthread_setschedparam(): %s",
--- a/src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
+++ b/src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
@@ -156,9 +156,11 @@ NetworkLoop::NetworkLoop(core::IPool& packet_pool,
         return;
     }
 
-    while (init_status_ == status::NoStatus) {
+    {
         core::Mutex::Lock lock(thr_init_mutex_);
-        thr_init_cond_.wait();
+        while (init_status_ == status::NoStatus) {
+            thr_init_cond_.wait();
+        }
     }
 }
 
--- a/src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
+++ b/src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
@@ -110,12 +110,12 @@ NetworkLoop::Tasks::ResolveEndpointAddress::get_address() const {
     return resolve_req_.resolved_address;
 }
 
-NetworkLoop::NetworkLoop(core::IPool& packet_pool,
+NetworkLoop::NetworkLoop(const NetworkConfig& net_config,
+                         core::IPool& packet_pool,
                          core::IPool& buffer_pool,
-                         const int realtime_prio,
                          core::IArena& arena)
-    : packet_factory_(packet_pool, buffer_pool)
-    , realtime_prio_(realtime_prio)
+    : net_config_(net_config)
+    , packet_factory_(packet_pool, buffer_pool)
     , arena_(arena)
     , started_(false)
     , thr_init_cond_(thr_init_mutex_)
@@ -299,7 +299,8 @@ void NetworkLoop::handle_resolved(ResolverRequest& req) {
 void NetworkLoop::run() {
     roc_log(LogDebug, "network loop: starting event loop");
 
-    if (realtime_prio_ > 0 && !Thread::enable_realtime(realtime_prio_)) {
+    if (net_config_.realtime_prio > 0
+        && !Thread::enable_realtime(net_config_.realtime_prio)) {
         core::Mutex::Lock lock(thr_init_mutex_);
 
         roc_log(LogError,
--- a/src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
+++ b/src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
@@ -151,6 +151,8 @@ NetworkLoop::NetworkLoop(core::IPool& packet_pool,
     task_sem_.data = this;
     task_sem_initialized_ = true;
 
+    roc_panic_if(init_status_ != status::NoStatus);
+
     if (!(started_ = Thread::start())) {
         init_status_ = status::StatusErrThread;
         return;
@@ -296,18 +298,18 @@ void NetworkLoop::handle_resolved(ResolverRequest& req) {
 
 void NetworkLoop::run() {
     roc_log(LogDebug, "network loop: starting event loop");
-    if (realtime_prio_ > 0 && !enable_realtime(realtime_prio_)) {
+
+    if (realtime_prio_ > 0 && !Thread::enable_realtime(realtime_prio_)) {
         core::Mutex::Lock lock(thr_init_mutex_);
 
         roc_log(LogError,
-                "network loop: can't set realtime priority of network thread. May need "
-                "to be root");
-        init_status_ = status::StatusFailedRealtime;
+                "network loop: can't elevate realtime priority, may need to be root");
+        init_status_ = status::StatusErrThread;
         thr_init_cond_.signal();
     } else {
         core::Mutex::Lock lock(thr_init_mutex_);
 
-        roc_log(LogDebug, "network loop: elevated realtime priority");
+        roc_log(LogNote, "network loop: elevated realtime priority");
         init_status_ = status::StatusOK;
         thr_init_cond_.signal();
     }
--- a/src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
+++ b/src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
@@ -112,10 +112,13 @@ NetworkLoop::Tasks::ResolveEndpointAddress::get_address() const {
 
 NetworkLoop::NetworkLoop(core::IPool& packet_pool,
                          core::IPool& buffer_pool,
+                         const int realtime_prio,
                          core::IArena& arena)
     : packet_factory_(packet_pool, buffer_pool)
+    , realtime_prio_(realtime_prio)
     , arena_(arena)
     , started_(false)
+    , thr_init_cond_(thr_init_mutex_)
     , loop_initialized_(false)
     , stop_sem_initialized_(false)
     , task_sem_initialized_(false)
@@ -153,7 +156,10 @@ NetworkLoop::NetworkLoop(core::IPool& packet_pool,
         return;
     }
 
-    init_status_ = status::StatusOK;
+    while (init_status_ == status::NoStatus) {
+        core::Mutex::Lock lock(thr_init_mutex_);
+        thr_init_cond_.wait();
+    }
 }
 
 NetworkLoop::~NetworkLoop() {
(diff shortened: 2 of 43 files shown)


src/internal_modules/roc_core/target_posix/roc_core/thread.cpp
```diff
@@ -57,6 +57,7 @@ bool Thread::enable_realtime() {
                 errno_to_str(err).c_str());
         return false;
     }
+    roc_log(LogDebug, "thread: set realtime priority");
```
gavv, line 60: Let's move this log before actually calling pthread_setschedparam, so that error will go after, like this:

```
    roc_log(LogDebug, "thread: setting realtime priority");

    if (int err = pthread_setschedparam(pthread_self(), SCHED_RR, &param)) {
        roc_log(LogError,
                "thread: can't set realtime priority: pthread_setschedparam(): %s",
                errno_to_str(err).c_str());
        return false;
    }
```

Also let's use LogError instead of LogDebug in the error branch.

src/tests/roc_rtcp/test_communicator.cpp
```diff
@@ -438,13 +438,16 @@ void expect_recv_report(const RecvReport& report,
     }
 }
 
-packet::PacketPtr read_packet(packet::FifoQueue& source) {
+packet::PacketPtr read_packet(packet::FifoQueue& source, core::nanoseconds_t ts = -1) {
```
gavv, line 441: nit: Let's rename `ts` to `recv_ts` (too many timestamps).

src/internal_modules/roc_netio/target_libuv/roc_netio/network_loop.cpp
```diff
@@ -140,6 +140,11 @@ NetworkLoop::NetworkLoop(core::IPool& packet_pool,
     task_sem_.data = this;
     task_sem_initialized_ = true;
 
+    if (!enable_realtime()) {
+        roc_log(LogInfo,
+                "network loop: can't set realtime priority of network thread. May need "
+                "to be root");
+    }
```
gavv, line 147: This should be called from `NetworkLoop::run()` because `enable_realtime()` enables SCHED_RR for current thread. Ctor is running on main thread.

Also, I think a few more changes are needed.

1. Add a CLI option (e.g., `--rt`) that enables real-time priority and set `SCHED_RR` only if it's provided. SCHED_RR is not desired or expected by all users (especially if you have many things running on your computer), so automatically enabling it when running as root (which is common e.g. when using ALSA) would be surprising.

2. Ideally CLI option should also accept integer priority which we should use instead of `sched_get_priority_max(SCHED_RR)`. E.g. `--rt=99`. Because you may want to give different priorities to different tools on system.

3. If the option was given, but `enable_realtime()` failed, we should report error and exit. If user asked us to run with SCHED_RR, it would be surprising to silently ignore errors.

   Since `enable_realtime()` should be called from `run()`, in ctor we can wait until `run()` notifies us that it's up and running or it's failed and exited. On error, we'll return error from `init_status()`.

4. Another thread sensitive to lags is pipeline thread that processes samples and writes them to e.g. ALSA. I think if the option is given, we should enable `SCHED_RR` there too. We can do it in `IoPump` ctor.

src/internal_modules/roc_rtcp/rtt_estimator.h
```diff
@@ -51,7 +53,7 @@ struct RttMetrics {
 class RttEstimator {
 public:
     //! Initialize.
-    RttEstimator(const RttConfig& config);
+    RttEstimator(core::IArena& arena, const RttConfig& config, dbgio::CsvDumper* dumper);
```
gavv, line 56: nit: Usual argument order pattern is "...., arena, dumper".

src/internal_modules/roc_pipeline/sender_slot.cpp
```diff
@@ -142,13 +142,13 @@ status::StatusCode SenderSlot::refresh(core::nanoseconds_t current_time,
...
             return code;
         }
     }
 
     if (repair_endpoint_) {
-        if ((code = repair_endpoint_->pull_packets(current_time)) != status::StatusOK) {
+        if ((code = repair_endpoint_->pull_packets(0)) != status::StatusOK) {
             return code;
         }
     }
```
gavv, line 154: Same comment as for receiver slot.

src/internal_modules/roc_pipeline/receiver_slot.cpp
```diff
@@ -84,13 +84,13 @@ status::StatusCode ReceiverSlot::refresh(core::nanoseconds_t current_time,
...
             return code;
         }
     }
 
     if (repair_endpoint_) {
-        if ((code = repair_endpoint_->pull_packets(current_time)) != status::StatusOK) {
+        if ((code = repair_endpoint_->pull_packets(0)) != status::StatusOK) {
             return code;
         }
     }
```
gavv, line 96: Since `current_time` is used only as a fallback when RTS isn't set, we can safely pass it to all calls I guess? Because why not to guarantee RTS for source and repair packets as well.

src/internal_modules/roc_rtcp/rtt_estimator.cpp
```diff
@@ -12,12 +12,17 @@
...
+                           dbgio::CsvDumper* dumper)
     : config_(config)
     , metrics_()
     , has_metrics_(false)
     , first_report_ts_(0)
-    , last_report_ts_(0) {
+    , last_report_ts_(0)
+    , dumper_(dumper)
+    , rtt_stats_(arena, 15, 0.5)
+    , clock_offset_stats_(arena, 100, 0.5) {
```
gavv, line 25: Please make this constants to be RttConfig fields with default values.

src/tests/roc_rtcp/test_communicator.cpp
```diff
@@ -3818,97 +3832,6 @@
     }
 }
 
-// Same, but there is a persistent clock drift between sender and receiver
-TEST(communicator, rtt_clock_drift) {
```
gavv, line 3822: If this test doesn't work due to smoothing of RTT, then we can disable smoothing specifically in this test by setting RttConfig fields (in some comment above I suggested to move smoothing settings to config).

gavv (changes requested):
Wow, this is big. 

I like the approach you use for RTS (set early when possible, set later as a fallback).

gavv (commented):
Wow, this is big. 

I like the approach you use for RTS (set early when possible, set later as a fallback).

gavv:
In `test_communicator.cpp`, there are a several calls to `read_packet()` which provide timestamp argument, but that makes little sense.

Since it's *receive* timestamp, it has meaning only when we're "delivering" packet to a communicator (and it should correspond to it's "local" time). But in cases when we're reading packet just to inspect it and throw away, timestamp is not actually used and specifying it looks confusing IMHO.

Diff below removes timestamps that are not needed.

```diff
diff --git a/src/tests/roc_rtcp/test_communicator.cpp b/src/tests/roc_rtcp/test_communicator.cpp
index fbe1e0d5..227dd902 100644
--- a/src/tests/roc_rtcp/test_communicator.cpp
+++ b/src/tests/roc_rtcp/test_communicator.cpp
@@ -3107,7 +3107,7 @@ TEST(communicator, report_to_address_sender) {
     CHECK_EQUAL(1, send_comm.total_destinations());
     CHECK_EQUAL(1, send_queue.size());
 
-    pp = read_packet(send_queue, send_time);
+    pp = read_packet(send_queue);
     expect_has_dest_address(pp, send_dest_addr);
     expect_has_orig_ssrc(pp, SendSsrc, true);
     expect_has_dest_ssrc(pp, Recv1Ssrc, true);
@@ -3279,7 +3279,7 @@ TEST(communicator, report_back_sender) {
     CHECK_EQUAL(1, send_comm.total_destinations());
     CHECK_EQUAL(1, send_queue.size());
 
-    pp = read_packet(send_queue, send_time);
+    pp = read_packet(send_queue);
     expect_has_dest_address(pp, recv1_addr);
     expect_has_orig_ssrc(pp, SendSsrc, true);
     expect_has_dest_ssrc(pp, Recv1Ssrc, true);
@@ -3320,13 +3320,13 @@ TEST(communicator, report_back_sender) {
     CHECK_EQUAL(2, send_comm.total_destinations());
     CHECK_EQUAL(2, send_queue.size());
 
-    pp = read_packet(send_queue, send_time);
+    pp = read_packet(send_queue);
     expect_has_dest_address(pp, recv1_addr);
     expect_has_orig_ssrc(pp, SendSsrc, true);
     expect_has_dest_ssrc(pp, Recv1Ssrc, true);
     expect_has_dest_ssrc(pp, Recv2Ssrc, false);
 
-    pp = read_packet(send_queue, send_time);
+    pp = read_packet(send_queue);
     expect_has_dest_address(pp, recv2_addr);
     expect_has_orig_ssrc(pp, SendSsrc, true);
     expect_has_dest_ssrc(pp, Recv1Ssrc, false);
@@ -3423,7 +3423,7 @@ TEST(communicator, report_back_receiver) {
     CHECK_EQUAL(1, recv_comm.total_destinations());
     CHECK_EQUAL(1, recv_queue.size());
 
-    pp = read_packet(recv_queue, send1_time);
+    pp = read_packet(recv_queue);
     expect_has_dest_address(pp, send1_addr);
     expect_has_orig_ssrc(pp, RecvSsrc, true);
     expect_has_dest_ssrc(pp, Send1Ssrc, true);
@@ -3469,13 +3469,13 @@ TEST(communicator, report_back_receiver) {
     CHECK_EQUAL(2, recv_comm.total_destinations());
     CHECK_EQUAL(2, recv_queue.size());
 
-    pp = read_packet(recv_queue, recv_time);
+    pp = read_packet(recv_queue);
     expect_has_dest_address(pp, send1_addr);
     expect_has_orig_ssrc(pp, RecvSsrc, true);
     expect_has_dest_ssrc(pp, Send1Ssrc, true);
     expect_has_dest_ssrc(pp, Send2Ssrc, false);
 
-    pp = read_packet(recv_queue, recv_time);
+    pp = read_packet(recv_queue);
     expect_has_dest_address(pp, send2_addr);
     expect_has_orig_ssrc(pp, RecvSsrc, true);
     expect_has_dest_ssrc(pp, Send1Ssrc, false);
@@ -3606,14 +3606,14 @@ TEST(communicator, report_back_combine_reports) {
     CHECK_EQUAL(2, recv_comm.total_destinations());
     CHECK_EQUAL(2, recv_queue.size());
 
-    pp = read_packet(recv_queue, send3_time);
+    pp = read_packet(recv_queue);
     expect_has_dest_address(pp, send1_addr);
     expect_has_orig_ssrc(pp, RecvSsrc, true);
     expect_has_dest_ssrc(pp, Send1Ssrc, true);
     expect_has_dest_ssrc(pp, Send2Ssrc, true);
     expect_has_dest_ssrc(pp, Send3Ssrc, false);
 
-    pp = read_packet(recv_queue, send3_time);
+    pp = read_packet(recv_queue);
     expect_has_dest_address(pp, send3_addr);
     expect_has_orig_ssrc(pp, RecvSsrc, true);
     expect_has_dest_ssrc(pp, Send1Ssrc, false);
@@ -4269,7 +4269,7 @@ TEST(communicator, rtt_network_reordering) {
                 recv_reorder_countdown = ReorderBurst;
             }
             recv_reorder_countdown--;
-            packet::PacketPtr pp = read_packet(recv_queue, recv_time);
+            packet::PacketPtr pp = read_packet(recv_queue);
             recv_packet_stash.push_back(*pp);
         }
```

src/internal_modules/roc_pipeline/sender_slot.cpp
MishaBaranov: I support your thought, but it will make pipeline tests dependent on real-time speed.

So far, `LatencyMonitor` calls `core::timestamp(core::ClockUnix)` when it computes `latency_metrics_.niq_stalling`. If we will set receive timestamp on all passing packets (which is your proposition here), we could not tell the difference between test and real use scenario. In order to pass these tests, I've set `base_cts` to current timestamp:

```diff
diff --git a/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp b/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp
index 4c47930e..2d7e188c 100644
--- a/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp
+++ b/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp
@@ -569,7 +569,7 @@ void send_receive(int flags,
     core::nanoseconds_t virtual_e2e_latency = 0;
 
     if (flags & FlagCTS) {
-        send_base_cts = 1000000000000000;
+        send_base_cts = core::timestamp(core::ClockUnix);
         virtual_e2e_latency = core::Millisecond * 100;
     }
 ```

In `receiver_source` test `latency_lower_bound` I had to do the same.

Please, take a look on 7837ebed578bf37b756de1fa9a43582baf0c2351

baranovmv: I support your thought, but it will make pipeline tests dependent on real-time speed.

So far, `LatencyMonitor` calls `core::timestamp(core::ClockUnix)` when it computes `latency_metrics_.niq_stalling`. If we will set receive timestamp on all passing packets (which is your proposition here), we could not tell the difference between test and real use scenario. In order to pass these tests, I've set `base_cts` to current timestamp:

```diff
diff --git a/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp b/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp
index 4c47930e..2d7e188c 100644
--- a/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp
+++ b/src/tests/roc_pipeline/test_loopback_sink_2_source.cpp
@@ -569,7 +569,7 @@ void send_receive(int flags,
     core::nanoseconds_t virtual_e2e_latency = 0;
 
     if (flags & FlagCTS) {
-        send_base_cts = 1000000000000000;
+        send_base_cts = core::timestamp(core::ClockUnix);
         virtual_e2e_latency = core::Millisecond * 100;
     }
 ```

In `receiver_source` test `latency_lower_bound` I had to do the same.

Please, take a look on 7837ebed578bf37b756de1fa9a43582baf0c2351

baranovmv: So, I propose to hold this change, until (if) we decide to introduce mock for `core::timestamp(...)` in pipeline tests.

gavv: Oh, I see. That's not good but I agree that it's outside of scope of this PR. I'll try to take a look at it later, will add it to my todo. Let's follow your proposal then.

baranovmv: I've reverted these changes for the time being

baranovmv:
@gavv "No conflicts with base branch" -- no excuses to not merge it?

I removed the commit where I deleted public-api tests, don't remember reason for that to exist.

src/internal_modules/roc_pipeline/sender_slot.cpp
gavv: It was actually very easy fix: 05ab131c025f5e848b752a1adfdabd643a0c49ea

gavv:
LGTM, rebased on fresh develop & pushed a few refactorings & fixes:

- 8d275bfa1cc450d1c874b28e7087afe849967db6, 031db3d73b6122d71f5b631296fca0dafc08207c - cosmetics
- 2e8f8a469b54f9dfea86642cd78b31c5117c9bf1, 089a7f01420759bc58cf90459fe861c341029b26 - move rt priority to config structs
- 05ab131c025f5e848b752a1adfdabd643a0c49ea - virtual time in tests

PR #787, opened 2024-12-09
Issue 772 proxy send reverse direction

kiranlahiri:
Introduce changes to the Proxy class to enable bidirectional handling of control packets, which was not previously supported.

Added a control endpoint (receiver_control_endp_) to manage control packets.
Update to handle and route control packets appropriately using a new control queue (control_queue_).
Enhanced configuration to bind control endpoints during Proxy initialization

Add test case ProxyWithControlEndpoint that validates the following
- Initialization of Proxy with source, repair, and control endpoints.
- Routing and handling of control packets via the Proxy class.
- Integration of control packets into the real-time streaming workflow.

### DIFF
--- a/src/tests/public_api/test_helpers/proxy.h
+++ b/src/tests/public_api/test_helpers/proxy.h
@@ -132,11 +132,23 @@ public:
         CHECK(roc_endpoint_set_port(input_repair_endp_,
                                     recv_repair_config_.bind_address.port())
               == 0);
+
+        if (receiver_control_endp) {
+            CHECK(roc_endpoint_allocate(&input_control_endp_) == 0);
+            CHECK(roc_endpoint_set_protocol(input_control_endp_, control_proto) == 0);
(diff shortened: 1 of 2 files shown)


gavv:
Hi, thanks for the patch.

1. Please rebase and retarget your PR on `develop` branch, see [contribution guidelines](https://roc-streaming.org/toolkit/docs/development/contribution_guidelines.html).

2. There is no need to create a new test file, we already have `test_loopback_sender_2_receiver.cpp`.

3. Can you tell me how you created this patch, did you ask an LLM to write it entirely or partially? I'm just curious.

   The code in `ProxyWithControlEndpoint` uses nonexistent API like `roc_proxy` and shouldn't compile. The test needs to be rewritten in line with the other tests and needs to use `test::Context`, `test::Receiver`, etc, instead of using `roc_context`, `roc_receiver`, etc.

I haven't looked into the code very closely yet, as it obviously needs major edits.

Issue #766, opened 2024-07-26
 Return status code instead of bool in IComposer


gavv:
## Summary

[packet::IComposer](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_packet/icomposer.h) is interface for protocol-specific serializators. See [documentation](https://roc-streaming.org/toolkit/docs/internals/packets_frames.html).

Currently its methods return bool (true on success or false on error). We need to replace bool with [status::StatusCode](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/internal_modules/roc_status/status_code.h) and return code that described why the operation failed.

## Implementation

* Update IComposer interface and implementations:
   * rtp::Composer
   * fec::Composer
   * rtcp::Composer
* Report appropriate statuses:
   * when allocation failed, return StatusNoMem
   * when buffer is too small, return StatusBadBuffer
   * on success, return StatusOK
* Update users of composers. They should forward status to upper level:
   * fec::BlockWriter
   * packet::Shipper
   * audio::Packetizer
   * rtcp::Communicator

## Testing

* ~Add unit tests for rtp::Composer and fec::Parser that check returned statuses in case of errors.~
* ~Add unit tests for components which use composers, and check that they forward statuses from composer to the upper level (e.g. if Composer fails with StatusNoMem, Shipper::write() returns StatusNoMem). One test per component should be enough.~

   See #792

ronazf:
Hi @gavv, I'd be happy to help with this issue.

gavv:
@ronazf Thanks, welcome!

gavv:
Landed

PR #789, opened 2024-12-22
gh-766: Refactored Boolean Returns to StatusCode in IComposer

sayedMurtadha:
Closes https://github.com/roc-streaming/roc-toolkit/issues/766

### DIFF
--- a/src/internal_modules/roc_audio/packetizer.cpp
+++ b/src/internal_modules/roc_audio/packetizer.cpp
@@ -161,12 +161,17 @@ status::StatusCode Packetizer::end_packet_() {
     // Fill protocol-specific fields.
     sequencer_.next(*packet_, packet_cts_, (packet::stream_timestamp_t)packet_pos_);
 
+    status::StatusCode code = status::StatusOK;
     // Apply padding if needed.
     if (packet_pos_ < samples_per_packet_) {
-        pad_packet_(written_payload_size);
+        code = pad_packet_(written_payload_size);
     }
 
-    const status::StatusCode code = writer_.write(packet_);
+    if (code != status::StatusOK) {
+        return code;
+    }
+
+    code = writer_.write(packet_);
     if (code != status::StatusOK) {
         return code;
     }
@@ -196,9 +201,10 @@ status::StatusCode Packetizer::create_packet_() {
         return status::StatusNoMem;
     }
 
-    if (!composer_.prepare(*pp, buffer, payload_size_)) {
+    status::StatusCode status = composer_.prepare(*pp, buffer, payload_size_);
+    if (status != status::StatusOK) {
         roc_log(LogError, "packetizer: can't prepare packet");
-        return status::StatusNoMem;
+        return status;
     }
     pp->add_flags(packet::Packet::FlagPrepared);
 
@@ -208,15 +214,18 @@ status::StatusCode Packetizer::create_packet_() {
     return status::StatusOK;
 }
 
-void Packetizer::pad_packet_(size_t written_payload_size) {
+status::StatusCode Packetizer::pad_packet_(size_t written_payload_size) {
     if (written_payload_size == payload_size_) {
-        return;
+        return status::StatusOK;
     }
 
-    if (!composer_.pad(*packet_, payload_size_ - written_payload_size)) {
-        roc_panic("packetizer: can't pad packet: orig_size=%lu actual_size=%lu",
-                  (unsigned long)payload_size_, (unsigned long)written_payload_size);
+    status::StatusCode status =
+        composer_.pad(*packet_, payload_size_ - written_payload_size);
+    if (status != status::StatusOK) {
+        roc_log(LogError, "packetizer: can't pad packet: orig_size=%lu actual_size=%lu",
+                (unsigned long)payload_size_, (unsigned long)written_payload_size);
     }
+    return status;
 }
 
 } // namespace audio
--- a/src/internal_modules/roc_fec/block_writer.cpp
+++ b/src/internal_modules/roc_fec/block_writer.cpp
@@ -223,9 +223,10 @@ status::StatusCode BlockWriter::write_source_packet_(const packet::PacketPtr& pp
 
     fill_packet_fec_fields_(pp, (packet::seqnum_t)cur_packet_);
 
-    if (!source_composer_.compose(*pp)) {
-        // TODO(gh-183): forward status from composer
-        return status::StatusBadBuffer;
+    status::StatusCode status = source_composer_.compose(*pp);
+    if (status != status::StatusOK) {
+        roc_log(LogError, "fec block writer: can't compose packet");
+        return status;
     }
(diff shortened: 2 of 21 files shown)


sayedMurtadha:
@gavv Can you please review the PR? Thanks

src/tests/roc_audio/test_depacketizer.cpp
```diff
@@ -62,7 +62,8 @@ packet::PacketPtr new_packet(IFrameEncoder& encoder,
     core::Slice<uint8_t> bp = packet_factory.new_packet_buffer();
     CHECK(bp);
 
-    CHECK(rtp_composer.prepare(*pp, bp, encoder.encoded_byte_count(SamplesPerPacket)));
+    CHECK(rtp_composer.prepare(*pp, bp, encoder.encoded_byte_count(SamplesPerPacket))
+          == status::StatusOK);
```
gavv, line 66: Here and in other tests: instead of:

```
CHECK(foo() == status::StatusOK);
```

please use:

```
LONGS_EQUAL(status::StatusOK, foo());
```

This way, on failure cpputest will print value returned by `foo()`, which makes debugging easier.

src/internal_modules/roc_packet/icomposer.h
```diff
@@ -42,7 +42,7 @@ class IComposer : public core::ArenaAllocation {
     //! @returns
     //!  true if the buffer was successfully adjusted or false if the @p buffer
     //!  capacity is not enough.
-    virtual bool
+    virtual status::StatusCode
     align(core::Slice<uint8_t>& buffer, size_t header_size, size_t payload_alignment) = 0;
```
gavv, line 46: Please add `ROC_ATTR_NODISCARD` to all methods that return StatusCode. It
instructs compiler to issue a warning if the caller forgets to check function
return value.

Also please add this attribute to all derived classes (rtp::Composer, fec::Composer, etc), because attributes are not inherited.

src/internal_modules/roc_packet/shipper.cpp
```diff
@@ -46,7 +46,7 @@ status::StatusCode Shipper::write(const PacketPtr& packet) {
     }
 
     if (!packet->has_flags(Packet::FlagComposed)) {
-        if (!composer_.compose(*packet)) {
+        if (composer_.compose(*packet) != status::StatusOK) {
             roc_log(LogError, "shipper: can't compose packet");
             return status::StatusNoMem;
         }
```
gavv, line 52: We need to forward status from composer here.

src/internal_modules/roc_fec/block_writer.cpp
```diff
@@ -223,9 +223,9 @@ status::StatusCode BlockWriter::write_source_packet_(const packet::PacketPtr& pp
 
     fill_packet_fec_fields_(pp, (packet::seqnum_t)cur_packet_);
 
-    if (!source_composer_.compose(*pp)) {
-        // TODO(gh-183): forward status from composer
-        return status::StatusBadBuffer;
+    status::StatusCode status = source_composer_.compose(*pp);
+    if (status != status::StatusOK) {
+        return status;
     }
```
gavv, line 229: nitpick: Please add `roc_log` call similar to how we do it in `make_repair_packet_()`.

src/internal_modules/roc_audio/packetizer.cpp
```diff
@@ -213,7 +214,8 @@ void Packetizer::pad_packet_(size_t written_payload_size) {
         return;
     }
 
-    if (!composer_.pad(*packet_, payload_size_ - written_payload_size)) {
+    if (composer_.pad(*packet_, payload_size_ - written_payload_size)
+        != status::StatusOK) {
         roc_panic("packetizer: can't pad packet: orig_size=%lu actual_size=%lu",
                   (unsigned long)payload_size_, (unsigned long)written_payload_size);
     }
```
gavv, line 221: Since we're now able to handle errors properly, could you please replace
`roc_panic` with `roc_log` here and return status from `pad_packet_()` and
forward it from `end_packet_()`?

src/internal_modules/roc_fec/block_writer.cpp
```diff
@@ -304,9 +305,9 @@
             continue;
         }
 
-        if (!repair_composer_.compose(*rp)) {
-            // TODO(gh-183): forward status from composer
-            return status::StatusBadBuffer;
+        status::StatusCode status = repair_composer_.compose(*rp);
+        if (status != status::StatusOK) {
+            return status;
         }
```
gavv, line 311: nit: Please add `roc_log` call similar to how we do it in `make_repair_packet_()`.

src/internal_modules/roc_rtcp/composer.cpp
```diff
@@ -35,51 +35,51 @@ bool Composer::align(core::Slice<uint8_t>& buffer,
...
         roc_log(LogError, "rtcp composer: unexpected null data");
-        return false;
+        return status::StatusBadOperation;
     }
 
     if (packet.rtcp()->payload.size() == 0) {
         roc_log(LogError, "rtcp composer: unexpected zero data");
-        return false;
+        return status::StatusBadOperation;
     }
```
gavv, line 80: A more correct status here would be `StatusBadArg`.

However, seems that all other composers treat invalid arguments as a violation and panic instead. Could you please replace `roc_log` + `StatusBadOperation` here with a panic as well?

gavv (changes requested):
Thanks a lot for the patch! Changes look fine, just a few minor comments.

src/internal_modules/roc_rtcp/composer.cpp
sayedMurtadha: Done

src/internal_modules/roc_fec/block_writer.cpp
sayedMurtadha: Done

src/internal_modules/roc_audio/packetizer.cpp
sayedMurtadha: Done

src/internal_modules/roc_fec/block_writer.cpp
sayedMurtadha: Done

src/internal_modules/roc_packet/shipper.cpp
sayedMurtadha: Done

src/internal_modules/roc_packet/icomposer.h
sayedMurtadha: Done

src/tests/roc_audio/test_depacketizer.cpp
sayedMurtadha: Done

gavv (approved):

gavv:
Thanks!

Issue #795, opened 2025-01-21
Fix bug in Semaphore::timed_wait


gavv:
`deadline` parameter of `Semaphore::timed_wait` is supposed to be absolute time in `CLOCK_MONOTONIC` domain. The code that uses it, as well as your example, pass monotonic timestamps to it.

However in `roc_core/target_posix_ext/roc_core/semaphore.cpp` we pass `deadline` to `sem_timedwait`, which expects time in `CLOCK_REALTIME` domain (i.e. since epoch).

Note that `roc_core/target_darwin/roc_core/semaphore.cpp` doesn't have this problem.

Suggested fix:

1. When [`sem_clockwait`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_clockwait.html) system call is available (we need a compile-time check), then use it instead of `sem_timedwait` with `CLOCK_MONOTONIC` clock id.

   This is the most efficient approach.

2. Otherwise, modify wait loop: each iteration, before calling `sem_timedwait`, compute wallclock deadline based on current time and monotonic deadline.

-----

Another concern with `sem_timedwait` is what happens if system time changes.

Seems that on recent linux / glibc it's not affected by time changes, i.e. if we asked it to wait until now + 100 seconds, it will sleep for 100 seconds even if time changed to past or future.

Just verified it using this small program:

<details>
  <summary>Click to expand</summary>
  
```cpp
#include <errno.h>
#include <semaphore.h>
#include <stdint.h>
#include <stdio.h>
#include <time.h>

int64_t timestamp(clockid_t clock) {
    timespec ts;
    clock_gettime(clock, &ts);
    return (int64_t)ts.tv_sec * 1000000000 + (int64_t)ts.tv_nsec;
}

int main() {
    sem_t sem;
    sem_init(&sem, 0, 0);

    int64_t deadline = timestamp(CLOCK_REALTIME) + 100000000000; // 100 seconds

    timespec ts;
    ts.tv_sec = long(deadline / 1000000000);
    ts.tv_nsec = long(deadline % 1000000000);

    int ret = sem_timedwait(&sem, &ts);

    printf("ret = %d\n", ret);
    perror("errno");
}
```

</details>

It will sleep 100 seconds even if I run `sudo date --set="2000-01-01 00:00:00"` or `sudo date --set="2027-01-01 00:00:00"` from another terminal.

[...]

PR #797, opened 2025-01-26
Update Semaphore::timed_wait with sem_clockwait and converting domain

Flw5469:
Fix according to #795
1. Use sem_clockwait if available
2. Else use sem_timedwait and convert time parameter's domain from ClockMonotonic into ClockUnix

### DIFF
--- a/src/internal_modules/roc_core/target_posix_ext/roc_core/semaphore.cpp
+++ b/src/internal_modules/roc_core/target_posix_ext/roc_core/semaphore.cpp
@@ -10,9 +10,16 @@
 #include "roc_core/cpu_instructions.h"
 #include "roc_core/errno_to_str.h"
 #include "roc_core/panic.h"
+#include "roc_core/log.h"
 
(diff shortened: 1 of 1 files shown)


gavv:
I haven't looked at the PR yet (let me know when it's ready), but I'vr checked how we could detect if sem_clockwait is available.

Seems that POSIX doesn't define any useful macro. The function was introduced in POSIX-2017, but they didn't bump `_POSIX_VERSION` and it's the same for POSIX-2008 and POSIX-2017.

So seems we should check it in the build system. You can add this to `SConstruct` before `conf.Finish()` call:

```python
# feature availability
if conf.CheckFunc('sem_clockwait', header='#include <unistd.h>'):
    conf.env.Append(CPPDEFINES=['ROC_HAVE_SEM_CLOCKWAIT'])

env = conf.Finish()
```

..and then you can use `#ifdef ROC_HAVE_SEM_CLOCKWAIT`

PR #800, opened 2025-02-08
#683 secure random implementation

### LINKED ISSUE #683: Implement secure_random() function using OpenSSL

adrianopol:
resolves #683
continuation of #727

### DIFF
--- a/src/internal_modules/roc_core/target_nouuid/roc_core/uuid.cpp
+++ b/src/internal_modules/roc_core/target_nouuid/roc_core/uuid.cpp
@@ -7,9 +7,9 @@
  */
 
 #include "roc_core/uuid.h"
-#include "roc_core/fast_random.h"
 #include "roc_core/macro_helpers.h"
 #include "roc_core/panic.h"
+#include "roc_core/secure_random.h"
 #include "roc_core/stddefs.h"
 
 namespace roc {
@@ -35,8 +35,9 @@ bool uuid_generate(char* buf, size_t buf_sz) {
     }
 
     uint8_t bytes[16];
-    for (size_t i = 0; i < ROC_ARRAY_SIZE(bytes); i++) {
-        bytes[i] = (uint8_t)fast_random_range(0, 255);
+    bool ok = core::secure_random(bytes, ROC_ARRAY_SIZE(bytes));
+    if (!ok) {
+        return false;
     }
 
     // Set variant to OSF DCE UUID.
--- a/src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
+++ b/src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2025 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "roc_core/secure_random.h"
+#include "roc_core/log.h"
+#include "roc_core/macro_helpers.h"
+
+#include <openssl/err.h>
+#include <openssl/rand.h>
+
+namespace roc {
+namespace core {
+
+bool secure_random(void* buf, size_t bufsz) {
+    // RAND_priv_bytes() is not needed right now as we do not use these random numbers
+    // privately. See also https://docs.openssl.org/3.0/man7/RAND/
+
+    int ok = RAND_bytes((unsigned char*)buf, (int)bufsz);
+
+    if (ok != 1) {
+        unsigned long err;
+        char err_str[256]; // minimum buf length is 256
+        memset(err_str, 0, ROC_ARRAY_SIZE(err_str));
+        while ((err = ERR_get_error()) != 0) {
+            ERR_error_string_n(err, err_str, ROC_ARRAY_SIZE(err_str));
+            roc_log(LogError, "secure random: OpenSSL RAND_bytes() failed: %s", err_str);
+        }
+        return false;
+    }
+    return true;
+}
+
+bool secure_random_range_32(uint32_t from, uint32_t to, uint32_t& dest) {
+    // validation `to >= from` exists in the 64-bit version called below
+
+    if (from == 0 && to == UINT32_MAX) {
+        return secure_random(&dest, sizeof(dest));
+    }
+
+    uint64_t rand64;
+    bool ok = secure_random_range_64(from, to, rand64);
+    if (!ok) {
+        return false;
+    }
+    dest = (uint32_t)rand64;
+    return true;
+}
+
+bool secure_random_range_64(uint64_t from, uint64_t to, uint64_t& dest) {
+    // same as fast_random_range() except it calls CSPRNG to get a uint64 value
+
+    roc_panic_if_msg(from > to, "secure random: invalid range: from=%llu to=%llu",
+                     (unsigned long long)from, (unsigned long long)to);
+
+    // corner case check; needed to avoid a possible uint64_t overflow
+    if (from == 0 && to == UINT64_MAX) {
+        return secure_random(&dest, sizeof(dest));
+    }
+
+    const uint64_t range = to - from + 1;
+
+    // Generate a mask with 1's from bit 0 to the most significant bit in `range`.
+    // At each step, we double the count of leading 1's:
+    //  0001.......
+    //  00011......
+    //  0001111....
+    // Thanks to @rnovatorov for the hint.
+    uint64_t mask = range;
+    mask |= mask >> 1;
+    mask |= mask >> 2;
+    mask |= mask >> 4;
+    mask |= mask >> 8;
+    mask |= mask >> 16;
+    mask |= mask >> 32;
+
+    do {
+        bool ok = secure_random(&dest, sizeof(dest));
+        if (!ok) {
+            return false;
+        }
+        dest &= mask;
+    } while (dest >= range);
+
+    dest += from;
+
+    roc_panic_if_not(dest >= from);
+    roc_panic_if_not(dest <= to);
+
+    return true;
+}
+
+} // namespace core
+} // namespace roc
--- a/src/internal_modules/roc_fec/block_writer.cpp
+++ b/src/internal_modules/roc_fec/block_writer.cpp
@@ -7,11 +7,11 @@
  */
 
 #include "roc_fec/block_writer.h"
-#include "roc_core/fast_random.h"
 #include "roc_core/log.h"
 #include "roc_core/panic.h"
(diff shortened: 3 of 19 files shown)


src/internal_modules/roc_core/fast_random.cpp
```diff
@@ -7,6 +7,7 @@
  */
 
 #include "roc_core/fast_random.h"
+
 #include "roc_core/atomic_ops.h"
```
gavv, line 11: Nit: we use single block for all roc includes.

src/internal_modules/roc_core/target_nouuid/roc_core/uuid.cpp
```diff
@@ -35,8 +36,9 @@ bool uuid_generare(char* buf, size_t buf_sz) {
     }
 
     uint8_t bytes[16];
-    for (size_t i = 0; i < ROC_ARRAY_SIZE(bytes); i++) {
-        bytes[i] = (uint8_t)fast_random_range(0, 255);
+    bool ok = core::secure_random(bytes, ROC_ARRAY_SIZE(bytes));
+    if (!ok) {
+        roc_panic("secure_random() failed");
```
gavv, line 41: This error is caused by external events, so we shouldn't panic, we should return error instead.

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (c) 2025 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "secure_random.h"
```
gavv, line 9: Nit: we always specify module name in includes, like: "roc_foo/bar.h"

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,95 @@
...
+namespace core {
+
+bool secure_random(void* buf, size_t bufsz) {
+    // RAND_priv_bytes() is not needed right now as we do not use these random numbers
+    // privately. See also https://docs.openssl.org/3.0/man7/RAND/
+
+    int ok = RAND_bytes((unsigned char*)buf, (int)bufsz);
+
+    if (ok != 1) {
+        unsigned long err = ERR_get_error();
```
gavv, line 26: Openssl has a queue of errors, and we need to fetch errors in a loop (and print each one).

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,95 @@
...
+        const size_t err_str_len = 256; // minimum buf length is 256
+        char err_str[err_str_len];
+        ERR_error_string_n(err, err_str, err_str_len);
+        roc_log(LogError, "OpenSSL: failed to generate random number: %s", err_str);
```
gavv, line 30: Nit: OpenSSL -> secure random (name of component which produces log message)

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,95 @@
...
+bool secure_random(void* buf, size_t bufsz) {
+    // RAND_priv_bytes() is not needed right now as we do not use these random numbers
+    // privately. See also https://docs.openssl.org/3.0/man7/RAND/
+
+    int ok = RAND_bytes((unsigned char*)buf, (int)bufsz);
+
+    if (ok != 1) {
+        unsigned long err = ERR_get_error();
+        const size_t err_str_len = 256; // minimum buf length is 256
+        char err_str[err_str_len];
```
gavv, line 28: Nit: I'd zero initialize buffer to be sure it's not garbage if for some reason ERR_error_string_n won't fill it.

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+    return true;
+}
+
+bool secure_random_range_32(uint32_t from, uint32_t to, uint32_t& dest) {
+    // validation `to >= from` exists in the 64-bit version called below
+
+    if (from == 0 && to == UINT32_MAX) {
+        return secure_random(&dest, sizeof(dest));
+    }
```
gavv, line 41: I think this check is not needed in 32 bit version

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,95 @@
...
+    bool ok = secure_random_range_64(from, to, rand64);
+    if (!ok) {
+        return false;
+    }
+    dest = (uint32_t)rand64;
+    return true;
+}
+
+bool secure_random_range_64(uint64_t from, uint64_t to, uint64_t& dest) {
+    // same as fast_random_range() except it calls PRNG to get a uint64 value
```
gavv, line 53: PRNG -> CSPRNG?

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    roc_panic_if_msg(from > to, "secure random: invalid range: from=%llu to=%llu",
+                     (unsigned long long)from, (unsigned long long)to);
+
+    if (from == 0 && to == UINT64_MAX) {
+        return secure_random(&dest, sizeof(dest));
+    }
```
gavv, line 60: I'd add a comment that this check is needed to avoid overflow.

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.h
```diff
@@ -0,0 +1,45 @@
+/*
```
gavv, line 1: We can have only one secure_random.h (outside of target dirs) and two secure_random.cpp (inside target dirs)

src/internal_modules/roc_fec/block_writer.cpp
```diff
@@ -46,9 +46,14 @@ BlockWriter::BlockWriter(const BlockWriterConfig& config,
...
-    cur_block_repair_sn_ =
-        (packet::seqnum_t)core::fast_random_range(0, packet::seqnum_t(-1));
+    bool ok = core::secure_random(&cur_sbn_, sizeof(cur_sbn_));
+    if (!ok) {
+        return;
+    }
+    ok = core::secure_random(&cur_block_repair_sn_, sizeof(cur_block_repair_sn_));
+    if (!ok) {
+        return;
+    }
```
gavv, line 56: In case of error we should set init status (we have a code for prng failure)

Same in other ctors.

src/tests/roc_fec/test_block_encoder_decoder.cpp
```diff
@@ -92,9 +94,10 @@ class Codec {
     core::Slice<uint8_t> make_buffer_(size_t p_size) {
         core::Slice<uint8_t> buf = packet_factory_.new_packet_buffer();
         buf.reslice(0, p_size);
-        for (size_t j = 0; j < buf.size(); ++j) {
-            buf.data()[j] = (uint8_t)core::fast_random_range(0, 0xff);
-        }
+
+        bool ok = core::secure_random(buf.data(), buf.size());
+        CHECK_TRUE(ok);
```
gavv, line 99: CSPRNG is generally not needed in tests (it's slower).

gavv (changes requested):
Thanks!!

src/internal_modules/roc_core/fast_random.cpp
gavv: Yep, we don't use this style in this project. Just one block for all roc includes.

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
adrianopol: module name is always written by `roc_log` itself, or I don't get something?
```
#define roc_log(level, ...)                                                              \
    do {                                                                                 \
        ::roc::core::Logger& logger = ::roc::core::Logger::instance();                   \
        if ((level) <= logger.get_level()) {                                             \
            logger.writef((level), ROC_STRINGIZE(ROC_MODULE), __FILE__, __LINE__,        \
                          __VA_ARGS__);                                                  \
        }                                                                                \
    } while (0)
```

src/internal_modules/roc_core/target_openssl/roc_core/secure_random.cpp
adrianopol: it eliminates mask operations and the loop in line 81

src/internal_modules/roc_core/fast_random.cpp
adrianopol: ok, updated

Issue #772, opened 2024-07-30
Teach test::Proxy to send packets in reverse direction


gavv:
[test::Proxy](https://github.com/roc-streaming/roc-toolkit/blob/develop/src/tests/public_api/test_helpers/proxy.h) is a class that we use in integration tests for [C API](https://roc-streaming.org/toolkit/docs/api.html).

Proxy is inserted between sender and receiver. Instead of connecting sender to receiver, we connect sender to proxy and proxy to receiver. Mostly, proxy just forwards packets as is and calculates some metrics, but it can also simulate network losses by dropping some packets.

Currently Proxy supports two uni-directional endpoints: one for source packets (audio packets sent from sender to receiver) and another for repair packets (redundancy packets sent from sender to receiver). See [docs](https://roc-streaming.org/toolkit/docs/internals/packets_frames.html).

We want to **teach Proxy to support optional endpoint for control packets**. This endpoint should be bidirectional, i.e. sender sends control packets to receiver (via proxy), and receiver sends control packets to sender (via proxy).

This would allow us to improve tests for metrics.

kiranlahiri:
No worries. Yes, I am still interested!

gavv:
You're welcome then, let me know if you have any questions.

gavv:
@kiranlahiri Feel free to ping me if you'll decide to work on this.

GabrielPerezCSDev:
Hi, I’d like to work on this issue! I’ve looked into the `Proxy` class and see that it currently handles source and repair packets. I’d like to add support for control packets and ensure they can be forwarded bidirectionally.  

Can I be assigned to this? Let me know if there are any specific guidelines for this issue in particular I should follow. Thanks!

gavv:
@GabrielPerezCSDev You're welcome, thanks. There are no specific guidelines besides issue text and contributions guidelines in our docs.

GabrielPerezCSDev:
@gavv Hey had a question about the code and writing the new control packet functionality.
I added to the constructor to enable control packets as well as the repair and source packets. By "We want to teach Proxy to support optional endpoint for control packets" does this mean you would like something like an overloaded constructor that is only called for specific cases?

Such that we can now have an overload constructor for the proxy class for the optional use of control packets like:

```cpp
// New constructor for control packet support
Proxy(const roc_endpoint* receiver_source_endp,
      const roc_endpoint* receiver_repair_endp,
      const roc_endpoint* receiver_control_endp,
[...]

PR #802, opened 2025-02-12
control packet support in test::Proxy

GabrielPerezCSDev:
### Summary
This PR adds **support for control packet handling** in `Proxy` while maintaining backward compatibility with existing tests.

### Changes Implemented
- **Extended `Proxy` class** to handle control packets in addition to source and repair packets.
- **Added an overloaded constructor** to allow optional control packet support while preserving the existing test proxy API.
- **Updated `write()`** to correctly queue and forward control packets.
- **Ensured existing tests remain unaffected**

### Development Workflow Compliance
- Targeted **`develop`** branch, per project guidelines.
- Applied **code formatting (`scons -Q fmt`)** before submission.
- Preserved backward compatibility with existing tests.

### Notes
- Control packet support is now integrated into `Proxy`, following the same queuing mechanism as source and repair packets.

### DIFF
--- a/src/tests/public_api/test_helpers/proxy.h
+++ b/src/tests/public_api/test_helpers/proxy.h
@@ -43,8 +43,10 @@ public:
         , net_loop_(packet_pool_, buffer_pool_, arena_)
         , n_source_packets_(n_source_packets)
         , n_repair_packets_(n_repair_packets)
+        , n_control_packets_(0)
         , flags_(flags)
         , pos_(0) {
+        input_control_endp_ = NULL;
         LONGS_EQUAL(status::StatusOK, net_loop_.init_status());
 
         roc_protocol source_proto;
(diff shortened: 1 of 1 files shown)


src/tests/public_api/test_helpers/proxy.h
```diff
@@ -119,9 +121,129 @@ class Proxy : public core::Thread, private packet::IWriter {
...
+          const roc_endpoint* receiver_repair_endp,
+          const roc_endpoint* receiver_control_endp,
+          size_t n_source_packets,
+          size_t n_repair_packets,
+          size_t n_control_packets,
+          unsigned flags)
+        : packet_pool_("proxy_packet_pool", arena_)
+        , buffer_pool_("proxy_buffer_pool", arena_, 2000)
+        , queue_(packet::ConcurrentQueue::Blocking)
+        , net_loop_(packet_pool_, buffer_pool_, arena_)
```
gavv, line 134: It's not a good idea to copy-paste this huge constructor. If we need two constructors, we'd need to extract common code into a method and reuse it.

However, in this case I think it's better to have just one constructor and pass NULL for control endpoint in tests that don't use it. It's better to make it explicit.

I saw your message about backwards compatibility but didn't have a chance to answer before you opened PR, sorry. We don't need to preserve compatibility here and want to update tests instead.

src/tests/public_api/test_helpers/proxy.h
```diff
@@ -132,6 +254,10 @@ class Proxy : public core::Thread, private packet::IWriter {
         return input_repair_endp_;
     }
 
+    const roc_endpoint* control_endpoint() const {
+        return input_control_endp_;
+    }
```
gavv, line 259:
```suggestion
    const roc_endpoint* control_endpoint() const {
        CHECK(input_control_endp_);
        return input_control_endp_;
    }
```

Let's make it fail early if trying to use control endpoint when it's not configured.

src/tests/public_api/test_helpers/proxy.h
```diff
@@ -119,9 +121,129 @@ class Proxy : public core::Thread, private packet::IWriter {
               == 0);
     }
 
+    Proxy(const roc_endpoint* receiver_source_endp,
+          const roc_endpoint* receiver_repair_endp,
+          const roc_endpoint* receiver_control_endp,
+          size_t n_source_packets,
+          size_t n_repair_packets,
+          size_t n_control_packets,
```
gavv, line 129: `n_control_packets` is not a thing. source and repair packets are grouped into a FEC block, but control packets are not part of any block. We can just remove this parameter.

src/tests/public_api/test_helpers/proxy.h
```diff
@@ -180,26 +306,33 @@ class Proxy : public core::Thread, private packet::IWriter {
...
+            LONGS_EQUAL(status::StatusOK, control_queue_.write(pp));
         } else {
             pp->udp()->dst_addr = receiver_repair_endp_;
             LONGS_EQUAL(status::StatusOK, repair_queue_.write(pp));
         }
 
         for (;;) {
-            const size_t block_pos = pos_ % (n_source_packets_ + n_repair_packets_);
+            const size_t block_pos =
+                pos_ % (n_source_packets_ + n_repair_packets_ + n_control_packets_);
```
gavv, line 319: As mentioned above, `n_control_packets_` doesn't make sense, let's remove it. We can just never drop control packets for now, if we'll ever need it, we can add separate flag for that.

src/tests/public_api/test_helpers/proxy.h
```diff
@@ -180,26 +306,33 @@ class Proxy : public core::Thread, private packet::IWriter {
         if (pp->udp()->dst_addr == recv_source_config_.bind_address) {
             pp->udp()->dst_addr = receiver_source_endp_;
             LONGS_EQUAL(status::StatusOK, source_queue_.write(pp));
+        } else if (pp->udp()->dst_addr == recv_control_config_.bind_address) {
+            pp->udp()->dst_addr = receiver_control_endp_;
+            LONGS_EQUAL(status::StatusOK, control_queue_.write(pp));
         } else {
             pp->udp()->dst_addr = receiver_repair_endp_;
             LONGS_EQUAL(status::StatusOK, repair_queue_.write(pp));
         }
```
gavv, line 186: nitpick: let's reorder branches in a familiar way: source, repair, control.

gavv (changes requested):
Thanks for the patch! Some comments.

gavv:
The big part missing from the PR is bi-directional proxying of the control packets - control packets are sent in two directions, unlike source and repair packets. Actually that's the most important part of this feature (the rest is more straightforward), please read issue text in #772.

GabrielPerezCSDev:
Sorry for the late reply thank you for the updates. I will get right on it!

GabrielPerezCSDev:
https://github.com/roc-streaming/roc-toolkit/issues/772
## Overview
Updated and modified the PR to adhere to the requested changes. Please let me know if there are any issues and thank you in advance!

## Implementation Details

### Constructor Changes
- Added optional control endpoint parameter (NULL when not needed)
- Removed duplicate overloaded constructor
- Updated the constructor to handle setting up the control packets when not NULL

### Control Packet Handling
- Implemented bidirectional flow for control packets
  - Sender → Receiver: Similar to source/repair packets
  - Receiver → Sender: New handling with appropriate address adjustments
- Control packets processed separately from FEC block structure
- Control packets are never dropped unlike source packets

### Accessor Method
- Added CHECK for NULL in control_endpoint() accessor for early failure detection
- Safe handling in destructor for NULL control endpoint

## Testing
All tests have been updated to use the new constructor signature and pass NULL for the control endpoint parameter when not needed.

gavv (changes requested):

Issue #684, opened 2024-01-30
Support reading/write WAV file from stdin


gavv:
This is follow-up issue for #576.

In roc-recv and roc-send, we have special syntax `-o file:-` and `-i file:-` which means read or write from/to stdin.

This feature is supported by sox backend, but [wav sink]() and [wav source]() are missing support.

Testing:

```
roc-send -vv -s rtp+rs8m://localhost:10001 -r rs8m://localhost:10002 -c rtcp://localhost:10003 \
    -i file:- --input-format wav < input.wav
```

```
roc-recv -vv -s rtp+rs8m://localhost:10001 -r rs8m://localhost:10002 -c rtcp://localhost:10003 \
    -o file:- --output-format wav > output.wav
```

joelkub:
Hey, I'd like to help out on this issue.

gavv:
Thank you, you're welcome

joelkub:
Sorry, but I am having a little trouble figuring out how wav_sink and wav_source are even supposed to take in any input, it seems to me that roc-recv and roc-send take care of handling all the reading and writing to/from stdin. If you could help me understand how any input actually makes its way down to wav_sink and wav_source that would be greatly helpful.

joelkub:
![Image](https://github.com/user-attachments/assets/37b52a6f-86e3-4577-9561-225291a1ce67)

Or is this what the issue is? Some sort of input handling?

joelkub:
I believe I better understand the issue now, what I am not sure still about is how to get the file from the input/output command. When I look at what commands the program receives, input/output.wav do not appear to be one of them. If you could offer some advice on how to proceed that would be greatly appreciated. Thank you.

joelkub:
I believe that I have figured out how to write to stdin. It appears to me that wav_source already takes in input from stdin as well. I will be opening a pull request with my changes soon.

gavv:
> Sorry, but I am having a little trouble figuring out how wav_sink and wav_source are even supposed to take in any input, it seems to me that roc-recv and roc-send take care of handling all the reading and writing to/from stdin. If you could help me understand how any input actually makes its way down to wav_sink and wav_source that would be greatly helpful.

> I believe I better understand the issue now, what I am not sure still about is how to get the file from the input/output command. When I look at what commands the program receives, input/output.wav do not appear to be one of them. If you could offer some advice on how to proceed that would be greatly appreciated. Thank you.

[...]

PR #803, opened 2025-02-17
Added stdin write/read to WAV sink

joelkub:
Changes made for issue #684 

Added an if-statement that checks the output file for the stdin input case which sets the output file to stdout.
Lines 164-167 in wav_sink.cpp

Also  added the is_raw() check to wav_source to reflect code in wav_sink and other sources.
Lines 30-34 in wav_source.cpp

### DIFF
--- a/docs/sphinx/about_project/publications.rst
+++ b/docs/sphinx/about_project/publications.rst
@@ -3,6 +3,8 @@ Publications
 
 This page lists some useful or interesting articles and blog posts, ordered by the publication date.
 
+* `Roc Toolkit 0.4 + updated tutorial for live audio streaming <https://gavv.net/articles/roc-0.4/>`_
+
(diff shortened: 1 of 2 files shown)


gavv:
Thanks for PR,

Could you briefly describe how did you test it?

Also please rebase it on fresh develop, it's currently unmergeble. Note that we use rebase, not regular merge, see https://roc-streaming.org/toolkit/docs/development/workflow.html

src/internal_modules/roc_sndio/wav_sink.cpp
```diff
@@ -161,6 +161,11 @@ bool WavSink::open_(const char* path) {
         return false;
     }
 
+    if (*path == '-') {
```
gavv, line 164: It's not a correct way to compare c strings, use strcmp.

gavv:
https://github.com/roc-streaming/roc-toolkit/issues/684#issuecomment-2968765477

PR #804, opened 2025-02-17
683: libatomic1 dep: use only when needed + update docs

adrianopol:
related to #683

behaviour changes:
- only link with `libatomic1` with enabled OpenSSL

documentation changes regarding `libatomic1` dependency


gavv:
Thx! Follow-up: cada5f52fe6eef2ebcfad8c0d7155f917a475609

PR #812, opened 2025-04-04
Change theme, and other minor fixes

baranovmv:
* Propose to change theme to Read the Docs, and make ToC clickable
* fix docs/SConsript

Also, with RtD becoming possible to use google analytics with [analytics_id](https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-analytics_id) option

### DIFF
--- a/docs/SConscript
+++ b/docs/SConscript
@@ -38,7 +38,7 @@ if GetOption('enable_doxygen'):
         doc_env.Alias('doxygen', doxygen_targets))
 
 if GetOption('enable_sphinx'):
-    if not GetOption('enable_sphinx'):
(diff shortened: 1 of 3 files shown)


gavv:
As discussed in chat, closing in favor to:

* 7270977921ede77289ed4b891da4b2acd020b76a (cherry-picked from this PR)
* eec1be3e2cd3d403a246308f6a15389c11ff79ff (uncollapses TOC without changing theme to RtD)

Google analytics is not needed because we already have cloudflare web analytics, which is more lightweight and a bit more privacy respecting. It is inserted automatically by cloudflare, it's also zero configuration.

Issue #292, opened 2019-12-05
Minimal MinGW build for Windows


gavv:
We have #132 to track efforts related to Windows porting. We've already fixed a few minor issues with SCons on Windows.

This sub-task is the first significant step towards Windows support: make minimal Roc configuration building with MinGW:

```
scons --host=x86_64-w64-mingw32 --compiler=gcc --disable-lib --disable-tools --disable-tests --disable-examples --disable-doc --disable-openfec --disable-sox --disable-libunwind --disable-pulseaudio
```

(The exact --host may differ.)

See also [Dependencies](https://roc-project.github.io/roc/docs/building/dependencies.html) and [SCons options](https://roc-project.github.io/roc/docs/building/scons_options.html).

To achieve this, we should add a `windows` platform to [SConstruct](https://github.com/roc-project/roc/blob/master/SConstruct). As with other platforms, we should auto-detect it from `--host`. Currently supported platforms are `linux`, `darwin`, and `android`.

The selected platform defines what [target directories](https://roc-project.github.io/roc/docs/internals/code_structure.html#targets) are enabled.

We can enable the following target directories on Windows / MinGW:

* target_gcc
* target_stdio
* target_libuv
* target_nobacktrace
* target_nodemangle

If some code from these directories will fail to build on MinGW, we will have to fix it.

The following target directories, most likely, can't be enabled:

* target_posix
* target_posixtime

We will have to create target_windows, enable it on Windows, and fill it with Windows implementations of functions and classes from target_posix and target_posixtime.

We don't have to implement all of them in this task. Some of them are not needed because they are used only on POSIX systems. Some of them are not needed on minimal build. And some of them are needed, but may be no-op on Windows.

A reasonable approach would be to run compilation, see what's missing, and add corresponding implementations.

We will also have to fix any issues with SCons running on Windows, if they will appear.

BTW, MinGW toolchain is available on Linux. If it makes sense, we can start from fixing MinGW build on Linux first, and then try to run it on Windows.

After finishing this task, we should also add minimal MinGW build to Travis.

gavv:
```
scons --disable-shared --disable-tools --disable-soversion --disable-openfec \
  --disable-speexdsp --disable-sox --disable-sndfile --disable-openssl --disable-libunwind \
  --host=x86_64-w64-mingw32 --build-3rdparty=libuv roc_core
```

gavv:
Hi, thanks a lot for looking into this!

[...]

PR #817, opened 2025-06-11
Early fixes for MinGW build - draft PR

JMD-tech:
Hello,
Worked some more on porting. Got rid of all but one compile warning for MinGW.
update: fixed.

Here's the remaining warning:
```
src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp: In member function ‘bool roc::address::SocketAddr::is_multicast() const’:
src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp:150:16: warning: conversion to ‘long unsigned int’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
  150 |         return IN_MULTICAST(core::ntoh32u(saddr_.addr4.sin_addr.s_addr));
      |                ^~~~~~~~~~~~

```
Tried about any kind of casts to no avail, and was surprised to see it was the other way around, as ntohs uint32_t return that would be cast to signed in the macro, also probably the macro expansion made the compiler point to "function" macro instead of parameter, but same attempts with casts of parameter in in function param were also no avail.
Same as some attempts as replacing with !=0 and similar.
_____________________________________________________________________
Now i suppose i i'd have only this one to be solved to consider those modifications to wothy of merge?
Marking it as ready for review, and will start working on (supposed agree on priorities for MinGW32/Windows):
- (2) implement WIN32 native functions
- enable OpenFEC (well, not that a priority but as it's a single line of code away...)
- build roc-recv/roc-send tools

_____________________________________________________________________
gh-292

Ok, i think it's ready for review, removing the 'draft mark'...

With all of those (plus copy of some files from posix) it successfully generates libroc_*.a:
- (generic, all builds) new URL for libatomic_ops releases
- socket_addr & socket_ops modifications
- include errno_to_str.h from wav_sink/source.cpp
- minor fixes in windows cond.cpp and errno_to_str (local var name mismatch, and const attribute conflicting)
- getting MinGW build to compile without -fpermissive:
https://github.com/roc-streaming/roc-toolkit/issues/292#issuecomment-2985951346
- doing the socket_init/deinit() in Context c/dtor

Meanwhile i'll try to static-link this libroc_core.a in Windows roc-send, and try to cross-compile roc-recv/send to see if it works.
Then i think as next priorities:
- clear the last few remaining warnings when compiling for MinGW
- working on item (2) porting posix code to native Windows API

### DIFF
--- a/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
+++ b/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
@@ -6,7 +6,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+#ifndef __WIN32__
 #include <arpa/inet.h>
+#endif
 
 #include "roc_address/socket_addr.h"
 #include "roc_core/endian.h"
--- a/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
+++ b/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
@@ -148,10 +148,10 @@ int SocketAddr::port() const {
 }
 
 #ifndef ROC_TARGET_WINDOWS
-#define IN_MULTICAST_U(i) IN_MULTICAST(i) 
+#define IN_MULTICAST_U(i) IN_MULTICAST(i)
 #else
 // Defined as (((long)(i) & 0xf0000000) == 0xe0000000) in Windows, causes Wsign-conversion
-#define IN_MULTICAST_U(i) (((unsigned long)(i) & 0xf0000000u) == 0xe0000000u)
+#define IN_MULTICAST_U(i) (((unsigned long)(i)&0xf0000000u) == 0xe0000000u)
 #endif
 
 bool SocketAddr::is_multicast() const {
--- a/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
+++ b/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-#ifndef __WIN32__
+#ifdef ROC_TARGET_POSIX
 #include <arpa/inet.h>
 #endif
 
--- a/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
+++ b/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
@@ -31,13 +31,8 @@ socklen_t saddr_size(sa_family_t family) {
     }
 }
 
-/* friend declaration would still require the return type into the header...
-sa_family_t saddr_family(const roc::address::SocketAddr& sa) {
-    return sa.saddr_.addr4.sin_family;
-}*/
-
-sa_family_t saddr_family(const roc::address::saddr_t& sa) {
-    return sa.addr4.sin_family;
+sa_family_t saddr_family(const sockaddr_in& sa4) {
+    return sa4.sin_family;
 }
 
 } // namespace
@@ -54,7 +49,8 @@ void SocketAddr::clear() {
 }
 
 bool SocketAddr::has_host_port() const {
-    return saddr_family(saddr_) == AF_INET || saddr_family(saddr_) == AF_INET6;
+    return saddr_family(saddr_.addr4) == AF_INET
+        || saddr_family(saddr_.addr4) == AF_INET6;
 }
 
 bool SocketAddr::set_host_port_saddr(const sockaddr* sa) {
@@ -119,7 +115,7 @@ const sockaddr* SocketAddr::saddr() const {
 }
 
 socklen_t SocketAddr::slen() const {
-    return saddr_size(saddr_family(saddr_));
+    return saddr_size(saddr_family(saddr_.addr4));
 }
 
 socklen_t SocketAddr::max_slen() const {
@@ -127,7 +123,7 @@ socklen_t SocketAddr::max_slen() const {
 }
 
 AddrFamily SocketAddr::family() const {
-    switch (saddr_family(saddr_)) {
+    switch (saddr_family(saddr_.addr4)) {
     case AF_INET:
         return Family_IPv4;
     case AF_INET6:
@@ -138,7 +134,7 @@ AddrFamily SocketAddr::family() const {
 }
 
 int SocketAddr::port() const {
-    switch (saddr_family(saddr_)) {
+    switch (saddr_family(saddr_.addr4)) {
     case AF_INET:
         return core::ntoh16u((uint16_t)saddr_.addr4.sin_port);
     case AF_INET6:
@@ -149,7 +145,7 @@ int SocketAddr::port() const {
 }
 
 bool SocketAddr::is_multicast() const {
-    switch (saddr_family(saddr_)) {
+    switch (saddr_family(saddr_.addr4)) {
     case AF_INET:
         return IN_MULTICAST(core::ntoh32u(saddr_.addr4.sin_addr.s_addr));
     case AF_INET6:
@@ -160,7 +156,7 @@ bool SocketAddr::is_multicast() const {
 }
 
 bool SocketAddr::get_host(char* buf, size_t bufsz) const {
-    switch (saddr_family(saddr_)) {
+    switch (saddr_family(saddr_.addr4)) {
     case AF_INET:
         if (!inet_ntop(AF_INET, &saddr_.addr4.sin_addr, buf, (socklen_t)bufsz)) {
             return false;
@@ -185,12 +181,11 @@ SocketAddr::operator const struct unspecified_bool *() const {
 }
 
 bool SocketAddr::operator==(const SocketAddr& other) const {
-    //    if (saddr_family(saddr_) != other.saddr_family(saddr_)) {
-    if (saddr_family(saddr_) != saddr_family(other.saddr_)) {
+    if (saddr_family(saddr_.addr4) != saddr_family(other.saddr_.addr4)) {
         return false;
     }
 
-    switch (saddr_family(saddr_)) {
+    switch (saddr_family(saddr_.addr4)) {
     case AF_INET:
         if (saddr_.addr4.sin_addr.s_addr != other.saddr_.addr4.sin_addr.s_addr) {
             return false;
@@ -221,12 +216,6 @@ bool SocketAddr::operator==(const SocketAddr& other) const {
 bool SocketAddr::operator!=(const SocketAddr& other) const {
     return !(*this == other);
 }
-
-/*
-sa_family_t SocketAddr::saddr_family_(saddr_) const {
-    return saddr_.addr4.sin_family;
-}
-*/
 
 } // namespace address
 } // namespace roc
--- a/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
+++ b/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
@@ -18,6 +18,9 @@ namespace {
 #ifdef ROC_TARGET_WINDOWS
 typedef ADDRESS_FAMILY sa_family_t;
 typedef USHORT in_port_t;
+typedef size_t ntop_bufsz;
+#else
+typedef socklen_t ntop_bufsz;
 #endif
 
 socklen_t saddr_size(sa_family_t family) {
@@ -32,7 +35,7 @@ socklen_t saddr_size(sa_family_t family) {
 }
 
 sa_family_t saddr_family(const sockaddr_in& sa4) {
-    return sa4.sin_family;
+    return (sa_family_t)sa4.sin_family;
 }
 
 } // namespace
@@ -158,13 +161,13 @@ bool SocketAddr::is_multicast() const {
 bool SocketAddr::get_host(char* buf, size_t bufsz) const {
     switch (saddr_family(saddr_.addr4)) {
     case AF_INET:
-        if (!inet_ntop(AF_INET, &saddr_.addr4.sin_addr, buf, (socklen_t)bufsz)) {
+        if (!inet_ntop(AF_INET, &saddr_.addr4.sin_addr, buf, (ntop_bufsz)bufsz)) {
             return false;
         }
         break;
 
     case AF_INET6:
-        if (!inet_ntop(AF_INET6, &saddr_.addr6.sin6_addr, buf, (socklen_t)bufsz)) {
+        if (!inet_ntop(AF_INET6, &saddr_.addr6.sin6_addr, buf, (ntop_bufsz)bufsz)) {
             return false;
         }
         break;
--- a/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
+++ b/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
@@ -151,7 +151,9 @@ int SocketAddr::port() const {
 #define IN_MULTICAST_U(i) IN_MULTICAST(i)
 #else
 // Defined as (((long)(i) & 0xf0000000) == 0xe0000000) in Windows, causes Wsign-conversion
-#define IN_MULTICAST_U(i) (((unsigned long)(i)&0xf0000000u) == 0xe0000000u)
+// clang-format off
+#define IN_MULTICAST_U(i) (((unsigned long)(i) & 0xf0000000u) == 0xe0000000u)
+// clang-format on
 #endif
 
 bool SocketAddr::is_multicast() const {
--- a/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
+++ b/src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
@@ -13,6 +13,35 @@
 #include "roc_address/socket_addr.h"
 #include "roc_core/endian.h"
 
+namespace {
+
+#ifdef ROC_TARGET_WINDOWS
+typedef ADDRESS_FAMILY sa_family_t;
+typedef USHORT in_port_t;
+#endif
+
+socklen_t saddr_size(sa_family_t family) {
+    switch (family) {
+    case AF_INET:
+        return sizeof(sockaddr_in);
+    case AF_INET6:
+        return sizeof(sockaddr_in6);
+    default:
+        return 0;
+    }
+}
+
+/* friend declaration would still require the return type into the header...
+sa_family_t saddr_family(const roc::address::SocketAddr& sa) {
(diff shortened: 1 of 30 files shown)


JMD-tech:
Ok, learning to use PR's, i believed it would contain only the commit for the URL, but looks like the PR includes all commits including ones added later, so it contains also the socket_addr and _ops which i intended to put on a separate PR...

Now i get all CI tests pass, i'll continue working on local machine, as i suppose for incremental approach you prefer that i don't push anymore stuff into this PR yet, and wait for this one to be reviewed and maybe fixed some more.

gavv:
Hi, thanks for update!

I'll checkout changes next week, here are just 2 small remarks:

1. Current CI failure is unrelated to PR, you can rebase on fresh develop to fix it.

2. Seems that troubles with formatting are caused by [this poor script](https://github.com/roc-streaming/roc-toolkit/blob/7d7d73cd1cf4a7cc019b6491d45b9eb772dbf4e6/scripts/scons_helpers/format-header.py), not by clang-format ("scons fmt" invokes both things).

    I always wanted to rewrite it, so I did it now: 4d4abbbe67131069d2f5442d779276a6b1fc2be7

    I hope rebasing on develop will fix your issues and you won't need dummy comments.

    If I'm wrong and some problems are caused by clang-format, you can also use clang-format magic comment:

    ```c
    // clang-format off
    <something>
    // clang-format on
    ```

gavv:
> and once i've got it right, i think you'll agree it's better that i do another PR that would be equivalent in files modified but without extra commits like fixing things in previous ones to avoid cluttering branches with those ("playing back" the final modifications, and hopefully have all intermediate commits validate CI checks).

I'm fine with this approach, just FYI:

- When merging PR, github allows to squash all commits into one, so it's not really necessary to create a separate PR just for squashing.

    (Actually I use my own script for merging PRs, but it behaves the same way).

- If you force-push to your PR's branch, github will automatically update PR, so it's also not necessary to create a new PR if you want to edit history by hand.

    (Github PR is tied to fork name + branch name, not specific commit).

src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
```diff
@@ -6,7 +6,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+#ifndef __WIN32__
 #include <arpa/inet.h>
+#endif
```
gavv: Here and in other code, let's use `ROC_TARGET_POSIX` / `ROC_TARGET_WINDOWS` (whatever fits more natural) instead of `__WIN32__`.

E.g.:

```suggestion
#ifdef ROC_TARGET_POSIX
#include <arpa/inet.h>
#endif
```

Just to handle platform-specific code uniformly across the code base.

src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.h
```diff
@@ -12,8 +12,16 @@
 #ifndef ROC_ADDRESS_SOCKET_ADDR_H_
 #define ROC_ADDRESS_SOCKET_ADDR_H_
 
+#ifndef __WIN32__
 #include <netinet/in.h>
 #include <sys/socket.h>
+#else // ! __WIN32__
+#include <cstdint>
+typedef uint16_t sa_family_t;
+typedef uint16_t in_port_t;
```
gavv: Polluting global namespace is quite an unexpected side-effect for a header that just declares a class.

I think in this case we don't really need it - we can move these typedefs to a .cpp file to anonymous namespace. We can hide `saddr_size_()` and `saddr_family_()` from header and make them free functions in .cpp instead.

src/internal_modules/roc_core/target_windows/roc_core/cond.cpp
```diff
@@ -34,7 +34,7 @@ bool Cond::timed_wait(nanoseconds_t timeout) const {
         if (err == ERROR_TIMEOUT || err == WAIT_TIMEOUT) {
             return false;
         }
-        roc_panic("cond: SleepConditionVariableCS(): error %lu", error);
+        roc_panic("cond: SleepConditionVariableCS(): error %lu", err);
```
gavv: Oops, sorry. Actually, could you please use errno_to_str() in panic here and in Cond::wait()?

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
```diff
@@ -6,12 +6,24 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+// dummy comment
```
gavv: I guess it can be removed.

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
```diff
@@ -21,6 +33,15 @@
 #include "roc_core/panic.h"
 #include "roc_netio/socket_ops.h"
 
+// To compile without -fpermissive
+#ifndef __WIN32__
+typedef void os_buf_t;
+typedef int os_val_t;
```
gavv: Maybe something like `sockbuf_t` and `sockopt_t` would be better names? os_buf and os_val are a bit cryptic.

Also let's put them in anonymous namespace, as well as `sa_family_t` and `in_port_t`.

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
gavv, line 1: I suggest to rework & simplify ifdefs in this file a bit, a few small steps against ifdef hell :)

I'm thinking about something like this (pseudo-code):

```cpp
// Make it clear: it's POSIX-specific and it can be used if FD_CLOEXEC is present.
// Define it if it's available, no matter if we'll actually use it.
#if defined(ROC_TARGET_POSIX) && defined(FD_CLOEXEC)

bool set_cloexec(SocketHandle sock) {
   // fcntl
}

// Fallback
#else

bool set_cloexec(SocketHandle sock) {
   // no-op
}

#endif
```

```cpp
// This version should work on any POSIX
#if defined(ROC_TARGET_POSIX)

bool set_nonblock(SocketHandle sock) {
   // fcntl
}

#elif defined(ROC_TARGET_WINDOWS)

bool set_nonblock(SocketHandle sock) {
   // ioctlsocket
}

#endif
```

```cpp
// Optimized version for POSIX + special features.
#if defined(ROC_TARGET_POSIX) && defined(SOCK_CLOEXEC) && defined(SOCK_NONBLOCK)

bool socket_create(address::AddrFamily family, SocketType type, SocketHandle& new_sock) {
    // SOCK_CLOEXEC | SOCK_NONBLOCK
}

// Simple version
#else

bool socket_create(address::AddrFamily family, SocketType type, SocketHandle& new_sock) {
    // set_cloexec(), set_nonblock()
}

#endif

// same for socket_accept

...
```

Also, replace this:

```cpp
#ifdef __WIN32__
#define MSG_DONTWAIT (0) // Eeek! but ok...
#endif
```

with:

```cpp
ssize_t socket_try_recv(SocketHandle sock, void* buf, size_t bufsz) {
    ...

    // Let's handle MSG_DONTWAIT same way as we do it for other
    // optional flags: make it clear for the reader of this function
    // that this flag is actually set only on some platforms
    int flags = 0;
#ifdef MSG_DONTWAIT
    flags |= MSG_DONTWAIT;
#endif

    ...
}

// same for socket_try_send() and socket_try_send_to()

// same for SHUT_RDWR vs. SD_BOTH
```

It seems that `MSG_DONTWAIT` is not really needed here because socket is already in non-blocking mode. It also seems that `MSG_DONTWAIT` is not available even on all POSIX platforms. But I'd keep using it if it's available, just in case.

And also replace this:

```cpp
#if defined(SO_NOSIGPIPE) || defined(MSG_NOSIGNAL)

ssize_t socket_try_send(SocketHandle sock, const void* buf, size_t bufsz) {
    ...
```

with:

```cpp
#if (defined(ROC_TARGET_POSIX) && (defined(SO_NOSIGPIPE) || defined(MSG_NOSIGNAL))) \
  || defined(ROC_TARGET_WINDOWS)

ssize_t socket_try_send(SocketHandle sock, const void* buf, size_t bufsz) {
    ...
    
    // version without pthread_sigmask, covers windows
    
    ...
}

#else

...
```

I'd prefer to test POSIX-specific defines like `SO_NOSIGPIPE` only on "real" POSIX, to be more sure we're not testing something that is defined just for compatibility and not indicating presence of actual feature, like it happens sometimes.

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
```diff
@@ -653,6 +704,10 @@ ssize_t socket_try_send_to(SocketHandle sock,
     return ret;
 }
 
+#ifdef __WIN32__
+#define SHUT_RDWR (SD_BOTH)
+#endif
+
 bool socket_shutdown(SocketHandle sock) {
```
gavv, line 779: Shouldn't we use `::closesocket()` instead of `::close()` in `socket_close()`?

src/internal_modules/roc_netio/target_libuv/roc_netio/udp_port.cpp
```diff
@@ -116,11 +116,16 @@ bool UdpPort::open() {
...
+#ifndef __WIN32__
     const int fd_err = uv_fileno((uv_handle_t*)&handle_, &fd_);
+
     if (fd_err != 0) {
         roc_panic("udp port: %s: uv_fileno(): [%s] %s", descriptor(), uv_err_name(fd_err),
                   uv_strerror(fd_err));
     }
+#else  // __WIN32__
+    fd_ = handle_.socket;
+#endif // __WIN32__
```
gavv: I think we can actually cast fd returned by uv_fileno to a SOCKET?

uv_os_fd_t is HANDLE on windows (PVOID). And SOCKET is an integer of the pointer size that can be, IIRC, safely cast to HANDLE and used with functions that expect HANDLE. I think it's even a part of the contract.

If this is right, we could:

1. Change type of `fd_` field to `SocketHandle` and rename it to `sock_`
2. Call `uv_fileno()` to store int or handle into temporary `uv_os_fd_t fd`
3. Set `sock_` field to `(SocketHandle)fd`

src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp
JMD-tech: Ok, starting with the easiest ones :) , done.

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
JMD-tech: Ah, meanwhile read the "don't resolve" so won't do it for next ones and remember to use only comment and thumbs.

Was following libuv naming at this point, agree there's no reason to follow libuv into berkeley and those are better, done.

About the anonymous namespace, for the `sa_family_t` and `in_port_t` i've done a small portion of `namespace { ... }` inside the ifdefs, rarely used anon namespaces before so don't much know (will read about) the best practices.

Hesitated to put those down below in the already existing anon namespace block but thought for symmetry it could make sense here as in posix those types are provided by the headers included in the ifdef, not sure what is the best way?

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
JMD-tech: Ok, done.

For the `SHUT_RDWR`/`SD_BOTH`, hesitated to do a:

```
#if defined(SHUT_RDWR)
	flags |= SHUT_RDWR;
#elif defined(SD_BOTH)
	flags |= SD_BOTH;
#endif
```

But thought it might be better if someone works on a (very hypothetical?) platform providing neither, to have his attention called by a compile error rather than silently using a "default" value (which would probably be `SHUT_RD` / `SD_READ` which are zero on both platforms), so stayed with simple #else


BTW i noticed there are many `roc_panic_if(sock < 0)` that generates "comparison always false" warnings as `SOCKET` is unsigned in Windows.
Will look into this later, belongs to another commit/later PR, but i think it should be `(sock == SocketInvalid)`, at least for Windows.
(Or should i also keep current check to catch potential negative values other than -1?)

src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.h
JMD-tech: The conversion of `saddr_size_()` was totally straightforward as static member to free func, and i replaced the `uint16_t`'s by the `USHORT` and `ADDRESS_FAMILY` types provided and used by Winsock to get rid of the `<cstdint>` include.

Removed some no more applicable questions, found answer / next comment...

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
JMD-tech: Yes, i'd forgotten this. Done, and like for the others, defined the whole function for Windows.
I wonder if i should keep the `WSAEINTR` check:

```
bool socket_close(SocketHandle sock) {
    roc_panic_if(sock == SocketInvalid);

    if (closesocket(sock) == SOCKET_ERROR) {
        int sock_err = WSAGetLastError();
        if (sock_err == WSAEINTR) {
            // Winsock2: WSAEINTR in closesocket() can only happen using
            // WSACancelBlockingCall
            //  (not anymore supported in Winsock 2.2 we're using so it'd be safe removing
            //  this test, and have it LogError instead?)
            roc_log(
                LogDebug,
                "socket: closesocket(): assuming WSAEINTR does not indicate a failure");
        } else {
            roc_log(LogError, "socket: closesocket(): %s",
                    core::errno_to_str(sock_err).c_str());
            return false;
        }
    }

    return true;
}

```
Without it it would become as simple as the one in set_nonblock.

src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.h
JMD-tech: Ok, seen some builds failing with a "saddr_t not documented" error so i suppose that answers the question :)

Now implementing with passing addr4 argument...

gavv: LGTM

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
gavv: IMHO it would be handy to keep type declarations in one place and not mixed with the includes, it just seems easier to read code this way, WDYT?

Especially if more compatibility typedefs are added in future, e.g. if in_port_t would be missing on some systems.

BTW I guess we could replace:

```
typedef uint16_t sa_family_t;
typedef uint16_t in_port_t;
```

with:

```
typedef ADDRESS_FAMILY sa_family_t;
typedef USHORT in_port_t;
```

same as you did in another file?

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
gavv: > But thought it might be better if someone works on a (very hypothetical?) platform providing neither, to have his attention called by a compile error rather than silently using a "default" value

I agree, let's make it fail if neither is defined.

> BTW i noticed there are many roc_panic_if(sock < 0) that generates "comparison always false" warnings as SOCKET is unsigned in Windows.

I guess we could add a helper like `socket_is_valid(sock)` and use it in panics. It would be nice to catch invalid descriptors on posix.

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
gavv: Agree, I think we don't need to handle WSAEINTR

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
JMD-tech: Yes, agree now it is cleaner this way, done.

src/internal_modules/roc_netio/target_berkley/roc_netio/socket_ops.cpp
JMD-tech: Ok, done it this way.
Even if i first planned doing it later, as it's a minor change that clears most compile warnings i thought it was good having it in this PR.

Doing the symmetry gives a >= 0 comparison for POSIX. While i suppose it doesn't really happen "in the wild", reading some docs said nothing about a zero socket being invalid by the standard so i kept it this way.

About the warnings, as there aren't many remaining in this build (to what i've seen quickly scrolling console output, mostly mismatched signedness assignments and long long unsigned's passed to %lu formats), i'll look into those in detail after i've done the libuv socket handle thing.

src/internal_modules/roc_netio/target_libuv/roc_netio/udp_port.cpp
JMD-tech: Yes, done it this way. 
Was wary of the very special case of MinGW being somewhat "in between" POSIX and win32 API, having both 64b winsock socket handles and 32b signed fds. (No big problem as now Winsock allocates only 32b on the 64b socket handles).
I suppose libuv will have adapted if/when Winsock2 API starts allocating full 64b socket handles.
Yet maybe it'd be a good idea to try to PR libuv to have a similar function like uv_sockno()?
(I've seen there is already a discussion about this in libuv: https://github.com/libuv/libuv/issues/200#issuecomment-74080840)

JMD-tech:
Hello,
Worked some more on porting. Got rid of all but one compile warning for MinGW.

Here's the remaining warning:

```
src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp: In member function ‘bool roc::address::SocketAddr::is_multicast() const’:
src/internal_modules/roc_address/target_berkley/roc_address/socket_addr.cpp:150:16: warning: conversion to ‘long unsigned int’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
  150 |         return IN_MULTICAST(core::ntoh32u(saddr_.addr4.sin_addr.s_addr));
      |                ^~~~~~~~~~~~
```

Tried about any kind of casts to no avail, and was surprised to see it was the other way around, as ntohs uint32_t return that would be cast to signed in the macro, also probably the macro expansion made the compiler point to "function" macro instead of parameter, but same attempts with casts of parameter in in function param were also no avail.
Same as some attempts as replacing with !=0 and similar.

Now i suppose i i'd have only this one to be solved to consider those modifications to wothy of merge?
Marking it as ready for review, and will start working on (supposed agree on priorities for MinGW32/Windows):
- (2) implement WIN32 native functions
- enable OpenFEC (well, not that a priority but as it's a single line of code away...)
- build roc-recv/roc-send tools

Issue #696, opened 2024-02-09
wav backend surround support


gavv:
* WavSink: write WAVEX extension header with channel mask
* WavSource: read channel mask from dr_wav and pass it to ChannelSet

Together with #695 it should allow to properly read & write surround wav files.

Issue #694, opened 2024-02-09
Dynamic channel order


gavv:
Currently channel order can be only one of the predefined orders (`ChannelOrder` enum).

In #693 we're going to support formats which support a lot of channel orders (like AIFF), and instead of explicitly supporting every possible channel order, it would be better to rely on libsndfile and just to query channel map from it in runtime.

To make it work, we can do the following:

* Add new ChanOrder_Custom to ChannelOrder enum.
* Add array of channel positions (ChannelPosition) to ChannelSet (of size ChanPos_Max elements).
* If ChanOrder_Custom is used, allow to get/set this array in ChannelSet.
* In ChannelMapperMatrix, when ChanOrder_Custom is used, use array from ChannelSet instead of ChannelOrderTable.
* Add tests for ChannelSet and ChannelMapper.

Issue #693, opened 2024-02-09
libsndfile surround channel orders


gavv:
Issue to track channel orders of formats supported by sndfile.

https://libsndfile.github.io/libsndfile/api.html

| Format           | Order    | Link |
| ---------------- | -------- | ---- |
| SF_FORMAT_WAV    | SMPTE    | [sample](https://www.jensign.com/bdp95/7dot1voiced/index.html), [msdn](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/dd390971(v=vs.85))
| SF_FORMAT_AIFF
| SF_FORMAT_AU     | -        | not specified (?)
| SF_FORMAT_RAW    | -        | not specified
| SF_FORMAT_PAF
| SF_FORMAT_SVX
| SF_FORMAT_NIST
| SF_FORMAT_VOC
| SF_FORMAT_IRCAM
| SF_FORMAT_W64
| SF_FORMAT_MAT4
| SF_FORMAT_MAT5
| SF_FORMAT_PVF
| SF_FORMAT_XI
| SF_FORMAT_HTK
| SF_FORMAT_SDS
| SF_FORMAT_AVR
| SF_FORMAT_WAVEX
| SF_FORMAT_SD2
| SF_FORMAT_FLAC  | SMPTE    | [spec](https://xiph.org/flac/format.html#frame_header)
| SF_FORMAT_CAF   | SMPTE    | [spec](https://developer.apple.com/library/archive/documentation/MusicAudio/Reference/CAFSpec/CAF_spec/CAF_spec.html)
| SF_FORMAT_WVE
| SF_FORMAT_OGG
| SF_FORMAT_MPC2K
| SF_FORMAT_RF64
| SF_FORMAT_MPEG

gavv:
AIFF channel order:

![image](https://github.com/roc-streaming/roc-toolkit/assets/8436629/794950bd-1384-4e15-a730-3dd6cafde2f6)

gavv:
### FLAC

```
Channel assignment

    0000-0111 : (number of independent channels)-1. Where defined, the channel order follows
    SMPTE/ITU-R recommendations. The assignments are as follows:
        1 channel: mono
        2 channels: left, right
        3 channels: left, right, center
        4 channels: front left, front right, back left, back right
        5 channels: front left, front right, front center, back/surround left, back/surround right
        6 channels: front left, front right, front center, LFE, back/surround left, back/surround right
[...]

PR #830, opened 2026-02-23
cli tools: enable multi channels io- and packet- encodings args

baranovmv:
gengetopt intervents into parsing packet-encoding by splitting comma-separated channels set into array, e.g. 100:pcm@s16/44100/FL,FR
  [0]: 100:pcm@s16/44100/FL
  [1]: FR

Here is a proposition to allow other channels list separator such as ',._ ', e.g.:
--packet-encoding=100:pcm/4800/stereo,"101:pcm/48000/FL FR BL BR"

### DIFF
--- a/src/internal_modules/roc_audio/sample_spec_parse.rl
+++ b/src/internal_modules/roc_audio/sample_spec_parse.rl
@@ -298,9 +298,10 @@ bool parse_sample_spec_imp(const char* str, SampleSpec& sample_spec) {
             sample_spec.set_sample_rate(rate);
         }
 
+        delimiter = [,_\.] | space;
         surround_mask = ([a-z] [a-z0-9.]+) >start_token %set_surround_mask;
         surround_channel = [A-Z]+ >start_token %set_surround_channel;
-        surround_list = surround_channel (',' surround_channel)*;
+        surround_list = surround_channel (delimiter surround_channel)*;
 
         surround = (surround_mask | surround_list) %set_surround;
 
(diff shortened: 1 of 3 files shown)


src/internal_modules/roc_audio/sample_spec_parse.rl
```diff
@@ -312,15 +313,14 @@ bool parse_sample_spec_imp(const char* str, SampleSpec& sample_spec) {
...
         mtr_channel = mtr_number | mtr_range;
-        mtr_list = mtr_channel (',' mtr_channel)*;
+        mtr_list = mtr_channel (delimiter mtr_channel)*;
 
         mtr = (mtr_mask | mtr_list) %set_mtr;
 
         format = [a-z0-9_]+ >start_token %set_format;
         subformat = [a-z0-9_]+ >start_token %set_subformat;
         rate = [0-9]+ >start_token %set_rate;
-        channels_DISABLED = surround | mtr;
```
gavv, line 322: Surround and MTR are disabled in CLI because sndio backends don't have support yet:

- gh-696
- gh-694
- gh-693

(They are available in API, though. That's fine because API does not expose access to sndio).

src/internal_modules/roc_audio/sample_spec_parse.rl
```diff
@@ -298,9 +298,10 @@ bool parse_sample_spec_imp(const char* str, SampleSpec& sample_spec) {
             sample_spec.set_sample_rate(rate);
         }
 
+        delimiter = [,_\.] | space;
```
gavv, line 301: I don't like the idea of multiple separators. The spec already has 4 special symbols: `/` `@` `-` `,`.

Now we want to allow +3 more different special symbols equivalent to one of these 4. That looks pretty WTF and also may create ambiguity if we'll need more special symbols in future (though I hope not).

If we can't prevent gengetopt from doing this, I suggest to choose single separator.

I think `+` instead of `,` would look logical:

```
--packet-encoding 101:pcm@s18/48000/FL+FR+FC
```

gavv (changes requested):

gavv:
As discussed in chat, it's better to keep surround disabled in 0.5.

I suggest:

1. Revert enabling of surround
2. Keep separator fix, but instead of allowing multiple separators, replace `,` with `+`
3. Rebase on fresh develop for green CI

Issue #821, opened 2025-10-29
Multitrack packets too big per default ("packetizer: can't prepare packet")


fallingcats:
While trying to add support for additional channels to the pipewire module, I ran into the following issue while testing with 8 channels:

    [err] roc_audio: packetizer: can't prepare packet

I discovered I could prevent that error from occurring by setting `roc_sender_config.packet_length` to less than 2_891_156ns. I'd appreciate though if the defaults were set (or rather calculated?) such that this didn't happen.

I understand that using MULTITRACK is not exactly default in itself, but it's not documented (to my knowledge) that I manually need to limit packet size through `packet_length` (or possibly other means?). Increasing `roc_context_config.max_packet_size` also works (locally) but I'm not sure how that translates to network packets w.r.t MTU

Currently I'm using `sender_config.packet_length = 20_000_000 / channels`. Please let me know if that's a bad idea, thanks :)

<sub>roc `v0.4.0` (and pipewire master `76a31a47`)</sub>

baranovmv:
Thank you for filing the idea, seems like it could be negotiated automatically indeed!

Ayoub-glitsh:
Hi 👋
I’ve read through the issue and the discussion, and I’d be happy to help if you’re looking for contributors.
Let me know if you’d like me to investigate the packet sizing defaults or try to propose an improvement.
Thanks!

baranovmv:
@Ayoub-glitsh you're very much welcome! It'd be great if you'll be able to come up with your thoughts and general plan in order to get initial feedback earlier.

BTW, I wonder what discussion have you read through?

baranovmv:
@fallingcats have you managed to get the update to pipewire module doing something useful for you?  Have you uploaded anywhere? It'd be cool to see!

If it is under construction please don't hesitate to upload it please! I'm also into roc-sink module these days, so it might be cool to join forces

Ayoub-glitsh:
Hi @baranovmv, thanks for the welcome . Sorry for the confusion by “discussion” I mainly meant the original issue description and your comment about automatic negotiation, not an external thread.

Ayoub-glitsh:
@baranovmv , Before I dive deeper and start implementing changes, I just wanted to ask a quick question:
is this issue planned to be associated with a bounty?

If a bounty is added, could you please share the link to the platform where the bounty is hosted so I can register and follow the process properly?

I’m happy to work on the issue either way, but knowing this would help me better prioritize the time and scope of the work.

[...]

PR #831, opened 2026-02-28
cli tools: fix frame size estimation when frame_length isn't provided

baranovmv:
In multi channel scenario a frame sz become too big and Frame Factory refuses to allocate frames. Here the max frame size is estimated similar way to max packet sz estimation.

gh-821

### DIFF
--- a/src/internal_modules/roc_audio/profiler.h
+++ b/src/internal_modules/roc_audio/profiler.h
@@ -21,6 +21,7 @@
 #include "roc_core/rate_limiter.h"
 #include "roc_core/time.h"
 #include "roc_packet/units.h"
+#include "roc_sndio/io_config.h"
 #include "roc_status/code_to_str.h"
 
 namespace roc {
@@ -32,7 +33,7 @@ struct ProfilerConfig {
     //! Default Initialization.
     ProfilerConfig()
         : profiling_interval(core::Second)
-        , chunk_duration(10 * core::Millisecond) {
+        , chunk_duration(sndio::DefaultFrameLength) {
     }
 
     //! Override Initialization.
--- a/src/internal_modules/roc_sndio/io_config.h
+++ b/src/internal_modules/roc_sndio/io_config.h
@@ -20,6 +20,15 @@
 namespace roc {
 namespace sndio {
 
+namespace {
+
+//! Default duration of a frame. 10ms is rather high,
+//! but works well even on cheap sound cards and CPUs.
+//! Usually you can use much lower values.
+const core::nanoseconds_t DefaultFrameLength = 10 * core::Millisecond;
+
+} // namespace
+
 //! Sink and source config.
 struct IoConfig {
     //! Sample spec
--- a/src/tools/roc_copy/main.cpp
+++ b/src/tools/roc_copy/main.cpp
@@ -124,7 +124,12 @@ size_t compute_max_frame_size(const sndio::IoConfig& io_config) {
                       audio::ChanLayout_Surround, audio::ChanOrder_Smpte,
                       audio::ChanMask_Surround_7_1_4, 48000);
 
-    return spec.ns_2_samples_overall(io_config.frame_length) * sizeof(audio::sample_t);
+    core::nanoseconds_t len = io_config.frame_length;
+    if (len == 0) {
+        len = 10 * core::Millisecond;
+    }
+
+    return spec.ns_2_samples_overall(len) * sizeof(audio::sample_t);
 }
 
 bool parse_input_uri(const gengetopt_args_info& args, address::IoUri& input_uri) {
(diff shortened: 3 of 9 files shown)


src/tools/roc_copy/main.cpp
```diff
@@ -124,7 +124,12 @@ size_t compute_max_frame_size(const sndio::IoConfig& io_config) {
                       audio::ChanLayout_Surround, audio::ChanOrder_Smpte,
                       audio::ChanMask_Surround_7_1_4, 48000);
 
-    return spec.ns_2_samples_overall(io_config.frame_length) * sizeof(audio::sample_t);
+    core::nanoseconds_t len = io_config.frame_length;
+    if (len == 0) {
+        len = 10 * core::Millisecond;
```
gavv: Let's add:

```cpp
DefaultFrameLength = 10 * core::Millisecond;
```

to `roc_sndio/io_config.h` and remove it from:

- `roc_sndio/io_pump.cpp`
- `roc_sndio/target_pulseaudio/roc_sndio/pulseaudio_device.cpp`
- `roc_sndio/target_sox/roc_sndio/sox_sink.cpp`
- `roc_sndio/target_sox/roc_sndio/sox_source.cpp`

(all these definitions have same value of 10ms)

...and reuse it in both `main.cpp`

gavv (commented):
Thanks

gavv:
Also please rebase on fresh develop to fix CI

gavv:
If you remember a command that reproduces the bug, could you create an issue with command and output? I didn't see this issue in my tests. Nice to have, but not critical.

baranovmv:
Fixes #821

baranovmv:
@gavv 

> If you remember a command that reproduces the bug, could you create an issue with command and output? I didn't see this issue in my tests. Nice to have, but not critical.

The issue itself needs arguebly a better solution. This PR proposes a hot-fix for basic and straightforward scenarios in our CLI tools. So at the moment **there is no easy way to verify the fix proposed in this PR with our unit tests**.

IMHO the issue reveals a usability problem in our public API and as a consequence lead to a corresponding issue in command line parameter of CLI tools.

Let me elaborate on a *"better solution"* a bit: 

* I'd rather see `MTU` (size in bytes) parameter instead of `roc_sender_config::packet_length` (duration)
* `roc_context_config::max_frame_size` in it's order could be deduced from `MTU`, but sometimes it'd make more sense to deduce it from the application. E.g. for pipewire modules it'd be great to tie max_frame_size to quants

gavv:
I think you've accidentally committed distfiles from master:

image

src/internal_modules/roc_audio/profiler.h
```diff
@@ -32,7 +33,7 @@ struct ProfilerConfig {
     //! Default Initialization.
     ProfilerConfig()
         : profiling_interval(core::Second)
-        , chunk_duration(10 * core::Millisecond) {
+        , chunk_duration(sndio::DefaultFrameLength) {
     }
```
gavv, line 37: So far roc_audio doesn't depend on roc_sndio (it's the opposite), and I think it's different 10ms. Here 1sec and 1ms are related, and OTOH frame size should be irrelevant.

src/internal_modules/roc_sndio/io_config.h
```diff
@@ -20,6 +20,15 @@
...
+namespace {
+
+//! Default duration of a frame. 10ms is rather high,
+//! but works well even on cheap sound cards and CPUs.
+//! Usually you can use much lower values.
+const core::nanoseconds_t DefaultFrameLength = 10 * core::Millisecond;
+
+} // namespace
+
 //! Sink and source config.
```
gavv, line 32: Not sure what is best practice, but elsewhere in code we don't have anon namespace for .h consts.

```suggestion
//! Default duration of a frame. 10ms is rather high,
//! but works well even on cheap sound cards and CPUs.
//! Usually you can use much lower values.
const core::nanoseconds_t DefaultFrameLength = 10 * core::Millisecond;

//! Sink and source config.
```

gavv:
Agree with the better solution.

To take the idea further:

> * I'd rather see `MTU` (size in bytes) parameter instead of `roc_sender_config::packet_length` (duration)

I think we should have both. User can set packet length (if they care about latency; if they use non-PCM), or MTU (if they care), or both.

If only one is set, the other is derived. If both are set, we may either validate that packet len fits MCU, or use the smaller one.

> * `roc_context_config::max_frame_size` in it's order could be deduced from `MTU`, but sometimes it'd make more sense to deduce it from the application. E.g. for pipewire modules it'd be great to tie max_frame_size to quants

Actually, I think we should get make max_frame_size and max_packet_size optional, and if they're not set, don't place any limits. This would require reworking slab pool to support dynamic chunk size, which should be feasible (e.g. using buddy allocator).

gavv:
Please rebase on fresh develop to make CI green.

PR #833, opened 2026-04-24
fix #821: raise default max_packet_size for multitrack support

### LINKED ISSUE #821: Multitrack packets too big per default ("packetizer: can't prepare packet")

Ayoub-glitsh:
Fixes #821

The default `max_packet_size` of 2048 bytes is too small for multitrack
audio. For example, 8 channels at 48kHz with the default 5ms packet
length requires ~3904 bytes (240 samples × 8 channels × 2 bytes + RTP
header), causing the packetizer to silently fail with "can't prepare packet".

Changes:
- Raise `ContextConfig` default `max_packet_size` from 2048 to 16384 bytes,
  which covers up to ~34 channels at 48kHz/5ms/16-bit PCM without any
  manual configuration
- Improve the error message in `Packetizer::create_packet_()` to include
  `payload_size` and `buffer_capacity`, and hint the user to increase
  `max_packet_size` if they do hit the limit

### DIFF
--- a/src/internal_modules/roc_audio/packetizer.cpp
+++ b/src/internal_modules/roc_audio/packetizer.cpp
@@ -185,7 +185,12 @@ packet::PacketPtr Packetizer::create_packet_() {
     }
 
     if (!composer_.prepare(*packet, buffer, payload_size_)) {
-        roc_log(LogError, "packetizer: can't prepare packet");
+        roc_log(LogError,
+                "packetizer: can't prepare packet:"
+                " packet buffer is too small for the requested payload:"
+                " payload_size=%lu buffer_capacity=%lu"
+                " (increase max_packet_size in context config)",
+                (unsigned long)payload_size_, (unsigned long)buffer.capacity());
         return NULL;
     }
     packet->add_flags(packet::Packet::FlagPrepared);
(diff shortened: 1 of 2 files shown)


baranovmv:
The should definately be treated, but I'd rather pick a solution that handles possible packet framentations more explicit. Closing this PR, and let's discuss it further in the issue

Ayoub-glitsh:
Got it, thanks for the quick review!
What did you have in mind for handling the fragmentation side? I want to make sure I'm going in the right direction before reworking it. Happy to rebase on develop once we're aligned.

PR #834, opened 2026-04-24
fix #807: add --packet-encoding option to roc-send

Ayoub-glitsh:
The --rate flag only sets the input device sample rate, but the packet encoding defaults to L16_Stereo at 44100 Hz. When input rate differs from packet encoding rate, the sender creates a resampler even if the user doesn't want one.

Add --packet-encoding option that lets users register a custom RTP encoding and select it for outgoing packets. This allows sending at 48kHz (or any rate) without resampling:

  roc-send --rate=48000 --packet-encoding=96:s16/48000/stereo ...

The option accepts the format '<id>:<format>/<rate>/<channels>', e.g.:
  96:s16/48000/stereo
  97:s16/48000/mono

The encoding is registered in the context encoding map and the sender is configured to use it as the packet encoding.

PR #835, opened 2026-04-24
test #764: add unit tests for LatencyTuner

### LINKED ISSUE #764: Tests for dynamic latency and related metrics
We need two more tests distinguished to dynamic latency.

I propose to test `LatencyTuner` specifically and the overall receiver pipeline in different tests. This task describes tests for `LatencyTuner` in isolation: check how `LatencyTuner` drives target latency depenging on incoming metrics.

**How automatic latency works**

Naturally we test here **optimal latency estimation** and the behavior of a **state machine** integrated into `LatencyTuner` in conjunction.

* In order to test **latency estimation** and decision making we could put `LatencyTuner` into situations where:
  * Decreasing is enabled **iff** current latency is significantly greater optimal latency estimation. Resulting target latency gets into --min/max-latency limits and it does not equal to optimal latency estimation, but conforms to a formula described in related comments in the source code.
  * Increasing is enabled more or less immediately when it is needed, and resulting target latency gets into --min/max-latency limits.
* For state machine we check that transition happens if aforementioned conditions met and certain timeout passed.
  * From the beginning up to startup timeout period no transition takes place
  *  Decreasing and Increasing could happen only from idle state and then specific timeouts should pass before transition back to idle state happens.

Ayoub-glitsh:
Closes #764

Added unit tests for LatencyTuner in isolation (test_latency_tuner.cpp).

Tests cover the main scenarios from the issue:

valid/invalid initialization
bounds checking in bounds, out of bounds, stalling suppression, no data yet
scaling direction and clamping
E2E backend behavior
No build changes needed, test files are picked up automatically.

### DIFF
--- a/src/tests/roc_audio/test_latency_tuner.cpp
+++ b/src/tests/roc_audio/test_latency_tuner.cpp
@@ -1,0 +1,361 @@
+/*
+ * Copyright (c) 2024 Roc Streaming authors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <CppUTest/TestHarness.h>
+
+#include "roc_audio/latency_tuner.h"
+#include "roc_core/time.h"
+#include "roc_packet/units.h"
+
+namespace roc {
+namespace audio {
+
+namespace {
+
+enum {
+    SampleRate = 44100,
+    NumCh = 2,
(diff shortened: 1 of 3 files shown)


baranovmv:
Hi @Ayoub-glitsh ! I love this PR, thx for your efforts

I haven't been able to look onto the code itself, here are some general comments:

1. remove irrelevant changes which sneaked into from #834 
2. target branch must be `develop`, not `master`
3. please don't use dashes as logic delimiters, bring the test file code style into consistency with the others.
4. before you manage to pass the tests, please make draft, not an open pr

PR #836, opened 2026-04-25
Fix/821 mtu aware packet length

### LINKED ISSUE #821: Multitrack packets too big per default ("packetizer: can't prepare packet")

Ayoub-glitsh:
Fixes #821

Implements the approach discussed with @baranovmv adds a packet_mtu parameter and auto-derives packet length from MTU when neither is set.

neither set → min(5ms, derived from 1200 bytes default MTU)
only packet_mtu set → derive length from it
only packet_length set → nothing changes
both set → warn and pick the smaller one
Existing stereo users are unaffected since 5ms fits fine within 1200 bytes.

### DIFF
--- a/src/internal_modules/roc_pipeline/config.cpp
+++ b/src/internal_modules/roc_pipeline/config.cpp
@@ -7,6 +7,10 @@
  */
 
 #include "roc_pipeline/config.h"
+#include "roc_audio/sample_spec.h"
+#include "roc_core/log.h"
+#include "roc_core/stddefs.h"
+#include "roc_core/time.h"
 #include "roc_rtp/headers.h"
 
 namespace roc {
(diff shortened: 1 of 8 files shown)


baranovmv:
Hi @Ayoub-glitsh 

Thanks for posting.

1. target branch must be `develop`, not `master`
2. make tests pass, before that better make a draft pr
3. remove irrelevant changes

PR #846, opened 2026-06-27
ci: add workflow tests and fix concurrency group

sumitsaini03:
- tests/test_build_workflow.py: matrix/script consistency, dup job names, concurrency group, structural sanity

- requirements-dev.txt: PyYAML for the workflow tests

- build.yml: replace 'pull_request.number || github.ref' with workflow+event_name+ref so non-PR triggers don't share a slot under cancel-in-progress: true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,3 +1,22 @@
+# --------------------------------------------------------------------------
+# Notes on CI invariants guarded by tests/test_build_workflow.py
+# --------------------------------------------------------------------------
+#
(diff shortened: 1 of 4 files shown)


baranovmv:
Thanks for contribution. Unfortunately I'm closing the PR as it lacks
* motivation for the changes -- I mean, some of it might be useful, but it'd be better to bring an explicit explanation what lead you to introducing these changes, an issue would be perfect, but not required;
* and coherency -- keep PR small and related to a single purpose, fix or feature

